PR #47, opened 2026-02-16
Implement entry point discovery for custom digest hooks

pmrv:
This change allows other packages to register custom digest functions for their types using Python entry points. The discovery mechanism is lazy, triggered only when a type cannot be hashed by the built-in logic or existing hooks. It also ensures that manual registration via `add_hook` maintains the highest priority and provides helpful logging when hooks are overridden or multiple entry points conflict.

---
*PR created automatically by Jules for task [6183351316784295145](https://jules.google.com/task/6183351316784295145) started by @pmrv*

### DIFF
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -2,6 +2,7 @@ import hashlib
 import dataclasses
 import struct
 import types
+import importlib.metadata
 from collections.abc import Iterable
 from typing import Any, TypeVar, Callable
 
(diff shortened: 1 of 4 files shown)


src/fleche/digest.py
```diff
@@ -41,7 +43,53 @@ def add_hook(hook: Hook | tuple[str, Callable[[T], str]]):
...
+                if hook.type in seen_types:
+                    source = seen_types[hook.type]
+                    if source == "add_hook":
+                        print(
+                            "INFO",
+                            f"add_hook for {hook.type} overrides entry point {ep.name}",
+                        )
+                    else:
+                        existing_hook = next(h for h in _EP_HOOKS if h.type == hook.type)
+                        if existing_hook.digest != hook.digest:
```
pmrv: Probably want to use `is` here.

tests/unit/test_entry_points.py
pmrv: Add docstrings to all remaining tests that clearly spell out the test intent.

PR #79, opened 2026-02-21
Introduce cache hit/miss statistics on Call objects

pmrv:
This change introduces tracking of cache hits and misses for each cached function call. The statistics are stored in the `Call` object and persisted in the storage backends (including SQL). 

Key changes:
1. `Call` object now has a `stats` attribute containing `hits` and `misses`.
2. `Cache.load` increments `hits` upon a successful cache hit.
3. `Cache.save` increments `misses` (it's called when a result is newly computed or updated).
4. `Sql` storage implementation was updated to handle these new fields efficiently, using row updates instead of full replacements where possible.
5. Invalidation is avoided by ensuring `stats` is not part of the `to_lookup_key()` digest.
6. Existing tests were updated to handle the new `load` calls during `save` operations.
7. New tests verify persistence, key stability, and stack behavior.

---
*PR created automatically by Jules for task [12014194863574883799](https://jules.google.com/task/12014194863574883799) started by @pmrv*

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -7,7 +7,7 @@ import pandas as pd
 
 from .digest import digest, Digest
 from . import storage
-from .call import Call
+from .call import Call, Statistics
 
 
 class Rejected(Exception):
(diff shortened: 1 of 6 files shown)


pmrv:
Actually only hits are required and misses are meaningless to track on the call level (since we only ever miss once).  Need more time to conceptualize.

PR #91, opened 2026-02-24
Add file locking to FileStorages

pmrv:
Implemented process-aware file locking for all `FileStorage` backends (`PickleFile`, `CloudpickleFile`, `BagOfHoldingH5File`).
- On `save`, a `{key}.lock` file is created containing the process ID and timestamp.
- On `load`, if a lock file exists, the process waits with exponential backoff (starting at 1ms) up to a configurable timeout (default 1s).
- If the timeout is reached, it attempts to read anyway, logging a warning. If it fails, it logs an error and raises `KeyError`.
- Lock files are excluded from `list()` and removed during `evict()`.
- Added unit tests in `tests/unit/storage/test_file_locking.py`.

---
*PR created automatically by Jules for task [9718337648981191387](https://jules.google.com/task/9718337648981191387) started by @pmrv*

### DIFF
--- a/src/fleche/storage/file.py
+++ b/src/fleche/storage/file.py
@@ -16,53 +16,61 @@ logger = logging.getLogger("fleche.storage")
 
 
 @contextmanager
-def file_lock(
-    lock_path: Path, timeout: float, wait_start: float, key: str, mode: str
-) -> Generator[None, None, None]:
-    if mode == "write":
-        lock_path.parent.mkdir(parents=True, exist_ok=True)
-        lock_path.write_text(f"{socket.gethostname()}\n{os.getpid()}\n{time.time()}")
-        try:
-            yield
-        finally:
-            lock_path.unlink(missing_ok=True)
-    elif mode == "read":
-        tried_anyway = False
-        if lock_path.exists():
-            start_time = time.perf_counter()
-            wait_time = wait_start
-            while (
-                lock_path.exists()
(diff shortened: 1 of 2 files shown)


src/fleche/storage/file.py
```diff
@@ -1,11 +1,62 @@
...
+logger = logging.getLogger("fleche.storage")
+
+
+@contextmanager
+def file_lock(
+    lock_path: Path, timeout: float, wait_start: float, key: str, mode: str
+) -> Generator[None, None, None]:
+    if mode == "write":
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        lock_path.write_text(f"{os.getpid()}\n{time.time()}")
```
pmrv: Also add the hostname

src/fleche/storage/file.py
```diff
@@ -25,7 +78,30 @@ def _path(self, key: str) -> Path:
...
 
     def _evict(self, key: Digest) -> None:
         self._path(key).unlink(missing_ok=True)
+        (self.root / f"{key}.lock").unlink(missing_ok=True)
+
+    def save(self, value: Any, key: Digest | None = None) -> Digest:
+        if key is None:
+            key = digest(value)
+
+        lock_path = self.root / f"{key}.lock"
```
pmrv: Reuse `_path` method.

PR #144, opened 2026-03-08
benchmarks: Add contains checks to storage and integration benchmarks

pmrv:
Adds benchmarking loops for `storage.contains()` in `benchmark_storage.py` and for `func.contains()` in `benchmark_integration.py` to measure the performance of cache existence checks.

---
*PR created automatically by Jules for task [927007830346278516](https://jules.google.com/task/927007830346278516) started by @pmrv*


pmrv:
For contains there should be two values measures, contains when the key is not present and contains when the key is present.

PR #154, opened 2026-03-08
Clean up benchmark results table

pmrv:
Closes the issue where the benchmark results table output was ad-hoc and contained `None` values for some tests. Replaced the old grouping logic with a simpler flattening logic to generate a single structured table.

---
*PR created automatically by Jules for task [11660033117187782806](https://jules.google.com/task/11660033117187782806) started by @pmrv*


pmrv:
This is not enough.  You can drop the iterations column.  Instead of 

`storage_evict,,Memory/small_strings,100,4.2e-07`

consider something like

`storage, Memory, small_string, evict, 4.2e-7`

You may add additional columns if you find it simplifies the table wrangling in the main orchestration script to output the benchmark tables.

patch.py
pmrv: Remove.

update_table.py
pmrv: Remove

PR #163, opened 2026-03-11
Fix code_digest persistence in SQL storage

### LINKED ISSUE #77: Sql table() can return multiple entries for the same key, but list unique?!

pmrv:
The SQL storage backend was failing to persist the `code_digest` field of `Call` objects. This resulted in `Call` objects having a different `to_lookup_key()` after being loaded from the database, which caused issues with cache consistency and deduplication.

I have updated the `CallModel` to include a `code_digest` column and modified the `_save` and `_load` methods in `Sql` storage to handle this field.

I also verified that `Sql.query()` correctly uses `.distinct()` to avoid duplicate rows from joins and that `BaseCache.table()` deduplicates by lookup key.

Fixes #77

---
*PR created automatically by Jules for task [16942242522553011969](https://jules.google.com/task/16942242522553011969) started by @pmrv*

### DIFF
--- a/tests/integration/test_sql_table_uniqueness.py
+++ b/tests/integration/test_sql_table_uniqueness.py
@@ -1,0 +1,34 @@
+
+import pytest
+from fleche.storage.sql import Sql
+from fleche.storage.memory import Memory
+from fleche.caches import Cache
+from fleche import fleche, cache, tags
+
+def test_sql_table_uniqueness(tmp_path):
+    """Integration test to verify that cache.table() does not contain more items than cache.calls.list()
+    when using a SQL backend, even with joins.
+    """
+    db_path = tmp_path / "test.db"
+    sql_storage = Sql(str(db_path))
+    val_storage = Memory(storage={})
+    c = Cache(values=val_storage, _calls=sql_storage)
+
+    with cache(c):
+        @fleche
+        def my_func(a, b):
+            return a + b
+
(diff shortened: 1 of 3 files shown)


tests/integration/test_sql_table_uniqueness.py
```diff
@@ -0,0 +1,40 @@
...
+
+            # Use multiple arguments to ensure multiple rows in arguments table if not distinct
+            my_func(a=3, b=4)
+
+            # Check counts
+            num_calls = len(list(c.calls.list()))
+            table_df = c.table()
+
+            print(f"Number of calls: {num_calls}")
+            print(f"Table rows: {len(table_df)}")
```
pmrv: Do not print

tests/integration/test_sql_table_uniqueness.py
```diff
@@ -0,0 +1,40 @@
...
+from fleche.storage.sql import Sql
+from fleche.storage.memory import Memory
+from fleche.caches import Cache
+from fleche import fleche, cache, tags
+
+def test_sql_table_uniqueness():
+    """Integration test to verify that cache.table() does not contain more items than cache.calls.list()
+    when using a SQL backend, even with joins.
+    """
+    with tempfile.TemporaryDirectory() as tmpdir:
```
pmrv: Use a pytest fixture for that.  There are builtin options to get temp files.

PR #174, opened 2026-03-16
Verify threadpool support and add sticky state management

### LINKED ISSUE #171: Check that fleche wrapper functions can be called through a threadpool

pmrv:
This change verifies that `fleche` wrapper functions can be called through threadpools. It confirms that `contextvars` state does not propagate automatically to worker threads in some environments and provides tests for manual context propagation. Additionally, it introduces a `sticky=True` parameter to state management functions, allowing permanent state updates within the current context, which is useful for initializing worker threads or configuring global state before thread creation. The implementation also ensures that `sticky=True` calls return a `nullcontext()` for compatibility with `with` statements and maintains late-binding for stacked context managers.

Fixes #171

---
*PR created automatically by Jules for task [16618559736406541909](https://jules.google.com/task/16618559736406541909) started by @pmrv*


notebooks/Executors.ipynb
pmrv: Do not add the complicated examples of writting a new notebook within a notebook.  But do show the same example of preserving cache config across thread as in the tests.

PR #200, opened 2026-03-22
Cache _depth via @fleche with a per-save temporary in-memory cache

pmrv:
Extracts the nesting-depth logic into a module-level `_depth_impl` function decorated lazily with `@fleche`.  `DestructuringStorage._save` now wraps each call in a `with cache(Cache(Memory({}), Memory({})))` context so that repeated `_depth` calls for the same sub-value within one save operation hit the cache instead of re-traversing the sub-tree, reducing the worst-case from O(n²) to O(n) for data with repeated elements.

`DestructuringStorage._depth` becomes a thin wrapper over `_depth_impl` preserving the existing public API; all existing tests remain valid.

### DIFF
--- a/src/fleche/storage/base.py
+++ b/src/fleche/storage/base.py
@@ -11,6 +11,29 @@ from ..call import Call
 
 logger = logging.getLogger("fleche.storage")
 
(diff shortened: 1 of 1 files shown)


pmrv:
I like the idea of using fleche inside fleche, but don't like the lazy imports/wrapping at call time, skip.

PR #204, opened 2026-03-23
Let query return a dedicated iterator and move `table()` there

pmrv:
Implement also some convenience methods on it.

Incidentally fix a bug when expanding arguments in c34497046a3fa4052d5810ef5238d22c6899ab0b

### DIFF
--- a/notebooks/CacheStack.ipynb
+++ b/notebooks/CacheStack.ipynb
@@ -19,10 +19,10 @@
    "execution_count": 1,
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-03-20T00:49:47.020119Z",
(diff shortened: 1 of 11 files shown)


tests/unit/fleche/test_query_iterator.py
```diff
@@ -0,0 +1,251 @@
...
+from fleche.query import QueryIterator
+from fleche.storage import Memory
+
+
+@pytest.fixture
+def test_cache():
+    return Cache(values=Memory({}), _calls=Memory({}))
+
+
+def _make_cache_with_calls(*calls):
```
pmrv: @claude There should be a global fixture for this

PR #205, opened 2026-03-23
Make D() accept arbitrary values, add expand/shrink to Digest

### LINKED ISSUE #105: Loading values directly considered awkward
need to know hash from somewhere and no tags available to interactively find out

Maybe the upshot is that querying calls has to be to extremely simple and ergonomic one can pick values of argument and result as required.

pmrv:
Implements idea 2 from #105:

- `D()` now accepts any value (strings wrapped directly, others digested)
- `D` added to `__all__`
- `Digest.expand(cache=None)` and `Digest.shrink(cache=None)` delegate to the cache, defaulting to the current context cache
- `BaseCache.expand()` abstract method added with implementations in all concrete cache classes

Closes #105

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -328,7 +328,7 @@ class Cache(BaseCache):
         # AmbiguousDigestError propagates directly
 
         if calls_result is not None and values_result is not None:
-            return max(calls_result, values_result, key=len)
+            return max(calls_result, values_result, key=len)  # type: ignore
         elif calls_result is not None:
             return calls_result
         elif values_result is not None:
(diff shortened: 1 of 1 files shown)


src/fleche/__init__.py
```diff
@@ -10,14 +10,28 @@
...
+
+    If given a string, wraps it in a Digest directly after validating that it is a valid
+    hexadecimal string of at most :data:`~fleche.digest.DIGEST_LENGTH` characters. For any
+    other value, computes its digest.
 
     Digests passed as arguments to @fleche decorated functions are automatically expanded
     to their cached values.
+
+    Raises:
+        ValueError: If value is a string that is not a valid hexadecimal digest.
```
pmrv: Don't raise an error, just digest strings that do not look like digests.  Document this in the docstring clearly with a sphinx `warning` admonition.

pmrv:
@claude Venture a guess why the test aren't working.

PR #208, opened 2026-03-23
Add SizeLimitedCache with random eviction

### LINKED ISSUE #162: Size Limited with Random Eviction
1. wraps a plain Cache
2. keeps track of size and takes as init parameter the maximum size
3. whenever size larger maximum, evict a random call
4. check if arguments or result of evicted guy are used in other calls, if not drop them too
5. 3 & 4 should be in an atomic threadsafe step

pmrv:
Implements a size-limited cache wrapper as described in #162.

- Wraps a plain `Cache` and enforces a configurable `max_size`
- Randomly evicts a call when the limit is exceeded
- Removes orphaned argument/result values atomically under a lock
- Single overridable `_pick_eviction_target` method for future policy generalization

Closes #162

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -584,7 +584,7 @@ class CacheStack(BaseCache):
                 yield c
 
 
-class SizeLimitedMixin(BaseCache):
+class SizeLimitedMixin:
     """Mixin that enforces a maximum number of cached calls with random eviction.
 
     Combine this with :class:`Cache` (mixin first in MRO) to get a size-limited


src/fleche/caches.py
```diff
@@ -580,3 +582,135 @@ def _query(self, call: call.QueryCall) -> Iterable[LazyCall]:
...
+        return random.choice(keys)
+
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
+    def _evict_call_and_orphaned_values(self, key: _digest.Digest) -> None:
+        """Evict a call and any of its values not referenced by other calls.
+
+        Must be called while ``self._lock`` is held.
```
pmrv: @claude Comments like this don't help.  Ensure _lock is reentrant and pick it up on method entrance.

src/fleche/caches.py
```diff
@@ -580,3 +582,135 @@ def _query(self, call: call.QueryCall) -> Iterable[LazyCall]:
                     continue
                 seen.add(k)
                 yield c
+
+
+@dataclass
+class SizeLimitedCache(BaseCache):
```
pmrv: @claude Consider a mixin implementation

PR #227, opened 2026-03-25
Add integration test for fleche cache with random kwargs

pmrv:
## Summary
This PR adds a comprehensive integration test that validates the caching and querying behavior of the `@fleche` decorator when called with random keyword arguments. The test uses property-based testing via Hypothesis to ensure robustness across a wide variety of input scenarios.

## Key Changes
- **New integration test** (`test_fleche_cache_random_kwargs.py`): Property-based test using Hypothesis that verifies:
  - `function.query(**kwargs)` correctly retrieves cached calls with exact argument matching
  - `Cache.query()` with wildcard templates finds cached calls regardless of argument specificity
  - Query behavior remains correct even after modifying the original kwargs dictionary
  - Strict subset queries properly fail to match the original call (digest mismatch)
  
- **Test strategy**: Uses custom strategies to generate random valid Python identifiers as keys mapped to nested values (including dataclasses), with 100 test examples per run

- **.gitignore**: Added common Python artifacts and Hypothesis cache directories to version control exclusions

## Notable Implementation Details
- The test reserves the `"metadata"` key since it's reserved by the `.query()` API
- Validates that the cached result persists independently of modifications to the local kwargs dictionary
- Tests both exact matching (full kwargs) and wildcard matching (partial/no arguments) query patterns
- Uses `deadline=None` to accommodate slower property-based test execution

https://claude.ai/code/session_016njk389y4hC6evaUTtaAWn

### DIFF
--- a/tests/integration/test_fleche_cache_random_kwargs.py
+++ b/tests/integration/test_fleche_cache_random_kwargs.py
@@ -1,0 +1,132 @@
+"""Integration test: random kwargs through @fleche, then query via Cache.query and function.query."""
+
+import random
+from dataclasses import dataclass
+
+import hypothesis.strategies as st
+from hypothesis import given, settings
+
+from fleche import fleche, cache
+from fleche.call import QueryCall
+from fleche.caches import Cache
+from fleche.storage import Memory
+
+
+# ---------------------------------------------------------------------------
+# Sample dataclasses for hypothesis
+# ---------------------------------------------------------------------------
+
+@dataclass(frozen=True)
+class Point:
+    x: float
+    y: float
+
+
+@dataclass(frozen=True)
+class Config:
+    name: str
+    value: int
+
+
+# ---------------------------------------------------------------------------
+# Hypothesis strategy for values that fleche can digest
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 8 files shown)


pmrv (commented):
@claude see comments

tests/integration/test_fleche_cache_random_kwargs.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+        assert any(m.result == result for m in matches)
+
+        # --- 2. Cache.query with wildcard arguments ------------------------
+        tpl = QueryCall(name="test", arguments={"kwargs": None})
+        cache_matches = list(test_cache.query(tpl))
+        assert len(cache_matches) >= 1, (
+            "Cache.query with wildcarded kwargs should find the call"
+        )
+        # --- 3. Delete random keys, then query ----------------------------
```
pmrv:
```suggestion
        # --- 2. Delete random keys, then query ----------------------------
```

pmrv (commented):
@claude Using `arguments={'kw': kwargs}` makes no sense when you call the function with `**`.  Rather you will need arguments=kwargs and variations.

Issue #240, opened 2026-03-30
Catalogue and reorganize test suite for tighter, more purposeful grouping


pmrv:
## Objective

Reorganize the test suite to be tighter and more purposeful, without losing coverage. This issue catalogues every test group, evaluates whether the current grouping makes sense, and identifies opportunities for consolidation or restructuring.

**Current stats:** ~270 unit tests across ~35 files, ~20 integration tests across 8 files, plus shared `fixtures.py` and `strategies.py`.

---

### Group 1: Digest — Core hashing logic
**Files:** `tests/unit/digest/test_digest.py` (19 tests), `tests/unit/digest/test_numpy_digest.py` (2 tests), `tests/unit/digest/test_entry_points.py` (6 tests)

**Core purpose:** Verify that `digest()` produces correct, deterministic, collision-resistant SHA256 hashes for all supported Python types; that type-specific paths (numpy, complex, iterables, dataclasses) are used; and that the hook/entry-point extension system works.

**Assessment:** Grouping is sound. `test_numpy_digest.py` has only 2 tests and could be folded into `test_digest.py` since those tests are just type-specific cases of the same `digest()` function. Entry points are a distinct subsystem and deserve their own file.

- [x] **Consolidate `test_numpy_digest.py` into `test_digest.py`** — both test `digest()` on different types. Numpy tests are just two more type-specific cases.
- [x] **Review `test_entry_points.py`** — well-scoped, keep separate. Verify no overlap with hook tests in `test_digest.py::test_custom_digest*`.

---

### Group 2: Digest — Ignored/Required argument handling
**Files:** `tests/unit/digest/test_ignore_digest.py` (9 tests), `tests/unit/digest/test_dataclass_input.py` (2 tests)

**Core purpose:** Verify that `Ignored` arguments are excluded from digest computation and that `Required` arguments enforce keyword-only passing; verify dataclass/dict inputs produce stable digests.

**Assessment:** `test_ignore_digest.py` tests the interaction between `Ignored`/`Required` markers and digest computation — this straddles the `wrapper.py` and `digest.py` boundary. `test_dataclass_input.py` is misplaced: it tests end-to-end caching of dataclass/dict inputs via the `@fleche` decorator (not the digest module directly). It belongs with the fleche decorator tests.

- [x] **Move `test_dataclass_input.py` to `tests/unit/fleche/`** — it tests `@fleche` caching with dataclass inputs, not the digest module.
[...]

PR #290, opened 2026-04-10
Optimize picklability tests: split Hypothesis tests from slow h5/sql …

pmrv:
…backends

HDF5 file-open/write/close per Hypothesis trial (~20-50 ms each) made the 21 h5-involved parametrized variants across the three functional roundtrip tests account for ~80 s (of 105 s total).

Fix: introduce fast_value_storage and fast_call_storage fixtures (memory, cloudpickle, dill, pickle only) for the three Hypothesis-based functional roundtrip tests.  Add six dedicated fixed-example tests to cover HDF5 and SQL functional roundtrips with representative Call shapes (minimal and full).

Every backend retains a functional roundtrip test; property-based exploration continues for all fast backends.  Test count: 237 → 213 Hypothesis-backed variants + 6 fixed h5/sql tests.  Wall time: 105 s → ~14 s (~7.5× speedup).

Part of #240.

### DIFF
--- a/tests/unit/test_pickle.py
+++ b/tests/unit/test_pickle.py
@@ -10,7 +10,9 @@ import pytest
 from hypothesis import given, settings, HealthCheck
 
 from fleche import fleche
-from fleche.storage import ValueMemory, CallMemory, ValueVoid, ValuePickleFile
+from fleche.call import Call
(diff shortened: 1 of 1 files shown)


pmrv:
Denied.  Let's make fast and slow test configs and require all to pass before merging.

PR #299, opened 2026-04-10
Special case namedtuples in DestructuringMixin

### LINKED ISSUE #298: NamedTuples break destructuring because they break LSP
https://docs.python.org/3/library/typing.html#typing.NamedTuple

derived classes define a `__new__` that expect each item separately, at variance with the `tuple` base class.  This is Liskov-Substitution-Principle violation.

We'll need to fix this here by special cases namedtuples in `DestructuringMixin`.

pmrv:
namedtuples are not real tuples, because they change __new__ such that it expects each value as a separate arg, rather than tuple's __init__ that accepts the full iterable.  Because they are by type subclasses of tuple, however, _rec_intern let's them into the DestructuringMixin garden.
DigestedIterable.mend is then hit by the colletaral when it cannot reinstantiate the namedtuple.
Because namedtuples are also have not common ancestor and are just a lose collection of sub types of tuple that happen to do subclass in the same way, there's not real way to check subtype relationship here. I follow an empiric approach here and check for some attributes. While I could apply this check in DigestedIterable.mend and special case the reconstruction, I am hurt now and I don't want to play with namedtuple anymore until it apologizes.

Fixes #298

### DIFF
--- a/src/fleche/storage/base.py
+++ b/src/fleche/storage/base.py
@@ -1,4 +1,5 @@
 import logging
+import collections
 from dataclasses import dataclass
 from numbers import Number
 
@@ -235,6 +236,10 @@ class DestructuringMixin(StorageBackend):
 
     remaining_depth: int = 0
 
+    @staticmethod
+    def _is_trojan_tuple(value):
+        return hasattr(value, "_fields") and hasattr(value, "_field_defaults")
+
     def _intern_rec(self, value: Any, key: Digest | None = None) -> tuple[Any, int | float]:
         """Post-order traversal: recurse to leaves, decide inline-vs-store on the way back up.
 
@@ -256,6 +261,12 @@ class DestructuringMixin(StorageBackend):
             # destructuring the empty container
             case dict() | list() | tuple() if not value:
                 depth = 0
+
+            # because nothing is ever simple, namedtuple break LSP by not accepting a single iterable
+            # this is considered highly annoying, because e.g. a lot of numpy functions we'd like to wrap return
+            # NamedTuple
+            case tuple() if self._is_trojan_tuple(value):
+                pass
 
             case list() | tuple():
                 value, depth = DigestedIterable.sunder(self._intern_rec, value)
--- a/tests/strategies.py
+++ b/tests/strategies.py
@@ -1,5 +1,6 @@
 from hypothesis import strategies as st
 from dataclasses import make_dataclass
+import collections
 import string
 import keyword
 import numpy as np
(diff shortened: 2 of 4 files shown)


pmrv:
@claude Add tests for namedtuples in the destructuring test suite and also add some to the hypothesis strategies.

pmrv:
**Claude finished @pmrv's task in 3m 11s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24265732371)

---
### Adding namedtuple tests and hypothesis strategies

- [x] Gather context: read existing test files and strategies
- [x] Understand the destructuring mechanism and what tests exist
- [x] Add namedtuple strategies to hypothesis strategies
- [x] Add namedtuple tests to the destructuring test suite
- [x] Run tests to verify (39 passed)
- [x] Commit and push

---

**What was added:**

**`tests/strategies.py`**
- New `namedtuples(field_types)` composite strategy — same caching pattern as `dataclasses`, uses `collections.namedtuple` with hypothesis-generated field names/values
- Added to `key_strategies` (namedtuples are hashable, so valid dict keys) so the existing roundtrip and determinism property tests automatically cover them
- Added to the `st_nested_values` recursive strategy so namedtuples appear in deeply-nested structures

**`tests/unit/storage/test_destructuring_storage.py`** — 10 new tests:
| Test | What it checks |
|------|---------------|
| `test_namedtuple_roundtrip` | Basic Point namedtuple round-trips with type preserved |
| `test_namedtuple_not_destructured` | Raw stored value is the namedtuple itself, never `DigestedIterable` |
| `test_namedtuple_in_list_roundtrip` | Namedtuple inside a list round-trips; type of element is preserved |
| `test_namedtuple_in_dict_roundtrip` | Namedtuple as a dict value round-trips correctly |
| `test_namedtuple_nested_in_list_not_destructured` | When a namedtuple is a list element the sibling can still be a `Digest`, but the namedtuple itself loads back as the right type |
| `test_namedtuple_hypothesis_roundtrip` | Hypothesis-generated namedtuples round-trip with type preserved |
| `test_namedtuple_hypothesis_not_destructured` | Hypothesis-generated namedtuples are always stored atomically |
| `test_namedtuple_save_is_deterministic` | Same namedtuple → same key every time |
 | [Branch: namedtuple](https://github.com/pmrv/fleche/tree/namedtuple)

pmrv:
@claude Merkle property tests fail.  I'm intuiting because namedtuples fall into the case of tuple() and then hash the same.  Have a look.

pmrv:
**Claude finished @pmrv's task in 8m 32s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24274705670)

---
### Fix: Merkle property test failures with namedtuples

- [x] Gather context and understand the request
- [x] Read the relevant test files and digest/storage code
- [x] Identify why namedtuples hash the same as tuples in Merkle tests
- [x] Fix the issue
- [x] Run tests to verify (716 passed, 3 skipped)
- [x] Commit and push

---

**Root cause:** In `randomly_digest_subvalues` (test_digest.py:202), namedtuples hit `isinstance(value, tuple)` and were reconstructed as plain `tuple(...)`, losing the namedtuple type. Since `_digest()` prefixes every hash with `type(value).__name__` (digest.py:160), `digest(NT(AAA=None))` ≠ `digest((None,))` — so replacing the namedtuple's field with its digest and then digesting the whole thing produced a different result, breaking the Merkle property.

**Fix** ([`tests/unit/digest/test_digest.py`](https://github.com/pmrv/fleche/blob/namedtuple/tests/unit/digest/test_digest.py)): Added a namedtuple guard _before_ the generic `tuple` branch using the same `_fields`/`_field_defaults` heuristic as `DestructuringMixin._is_trojan_tuple`, and reconstruct via `type(value)(*...)` to preserve the concrete type:

```python
elif isinstance(value, tuple) and hasattr(value, "_fields") and hasattr(value, "_field_defaults"):
    # namedtuple: reconstruct preserving the concrete type, since digest() includes type.__name__
    return type(value)(*[randomly_digest_subvalues(v, data) for v in value])
elif isinstance(value, tuple):
    return tuple(randomly_digest_subvalues(v, data) for v in value)
```
 | [Branch: namedtuple](https://github.com/pmrv/fleche/tree/namedtuple)

Issue #297, opened 2026-04-10
`required` handling still weird


pmrv:
```python
@fl.fleche(require='seed')
def draw(base, T, delta_mu, elements=np.array(['Au','Pt']), seed=None):
    if seed is None:
        return draw(base, T, delta_mu, elements, seed=np.random.randint(2**63))
  ...
```
this works as intended, i.e. generates a new seed if none given and then makes a cachable call to the wrapped function.  Yet, this doesn't!
```python
@fl.fleche(require='seed')
def draw(base, T, delta_mu, elements=np.array(['Au','Pt']), seed=None):
    print(seed)
    if seed is None:
        return draw(base, T, delta_mu, elements, np.random.randint(2**63))
```
And logs the `required` warning twice, indicated that the handling makes a difference whether its passed as a key or straight arg.  There likely was a confusion, because `Required` and `require` make only sense on keyword arguments at definition, but they should still catch one when called with positional arguments.

pmrv:
@claude Take a snoop where the issue is.  do not implement anything yet.

pmrv:
**Claude finished @pmrv's task in 1m 43s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24275832875)

---
### Analysis of `required` handling issue

- [x] Set up todo list
- [x] Read relevant source files (`wrapper.py`, `call.py`)
- [x] Trace the `require`/`Required` handling code path
- [x] Identify where positional vs keyword argument distinction causes the bug
- [x] Report findings

---

### Root Cause

The bug is in `wrapper.py` at line 277:

```python
missing = [r for r in required_args if r not in kwargs]
```

This check only looks at `kwargs` (keyword arguments dict). If a required argument is passed **positionally**, it will not appear in `kwargs` — even though the caller explicitly provided a value. The check then incorrectly treats it as "missing" and logs the warning, bypassing the cache.

### Tracing Both Examples

**Example 1 (works):**
```python
return draw(base, T, delta_mu, elements, seed=np.random.randint(2**63))
```
- `seed` is passed as a keyword argument → it is in `kwargs`
- `r not in kwargs` is `False` → `missing` is empty → caching proceeds normally ✓

**Example 2 (broken):**
```python
return draw(base, T, delta_mu, elements, np.random.randint(2**63))
```
- `seed` is passed as the 5th positional argument → it is in `args`, **not in `kwargs`**
[...]

PR #309, opened 2026-04-14
Fix required check to recognise positionally-supplied arguments (#297)

pmrv:
Previously the required-args guard only scanned `kwargs`, so an argument passed positionally (e.g. `draw(..., np.random.randint(...))`) was treated as missing and the warning was logged a second time, bypassing the cache.

Restructure `wrapper` so the `Call` object is built first; the required check is then done via `sig.bind(*args, **kwargs)` *without* `apply_defaults`.  Only arguments the caller explicitly passed appear in `bound.arguments`, whether they were positional or keyword — this is the same sig-binding machinery used inside `Call.from_call`, avoiding any duplication of argument-traversal logic.

Also remove the now-inaccurate POSITIONAL_ONLY warning from `process_ignore_required_args`; positional args are handled correctly.

Tests updated to assert the fixed behaviour and add a regression test for the exact scenario reported in the issue.

### DIFF
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -168,11 +168,6 @@ def fleche(
 
         ignored_args, required_args = process_ignore_required_args(func, ignore, require)
 
-        try:
-            sig = signature(func)
-        except (TypeError, ValueError):
-            sig = None
-
         @wraps(func)
         def get_call(*args, **kwargs):
             call = Call.from_call(func, *args, **kwargs)
@@ -289,7 +284,7 @@ def fleche(
                 logger.warning("No hash for argument: %s", e.args[0])
                 return func(*args, **kwargs)
 
-            if required_args and sig is not None:
+            if required_args:
                 explicit = set(bind(func, args, kwargs).keys())
                 missing = [r for r in required_args if r not in explicit]
                 if missing:
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -97,6 +97,17 @@ def process_ignore_required_args(
         require = tuple(require)
     required_args = require + tuple(type_required)
 
+    try:
+        sig = signature(func)
+        for r in required_args:
+            if r in sig.parameters and sig.parameters[r].kind == sig.parameters[r].POSITIONAL_ONLY:
+                logger.warning(
+                    "Argument '%s' is marked as Required but is positional-only. Required only works for keyword arguments.",
+                    r
+                )
+    except (TypeError, ValueError):
+        pass
+
     return ignored_args, required_args
 
 
@@ -279,11 +290,14 @@ def fleche(
                 return func(*args, **kwargs)
 
             if required_args and sig is not None:
-                # Use sig.bind without apply_defaults: only explicitly-provided arguments
-                # appear in bound.arguments, regardless of whether they were positional
-                # or keyword.  This fixes the bug where positional required args were
-                # incorrectly treated as missing.
-                explicit = set(sig.bind(*args, **kwargs).arguments.keys())
+                # Determine which required args were explicitly provided, without
+                # re-binding.  Positional args map to their parameter names by index;
+                # keyword args are in kwargs directly.
+                positional_params = [
+                    name for name, p in sig.parameters.items()
+                    if p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD)
+                ]
+                explicit = set(positional_params[:len(args)]) | set(kwargs.keys())
                 missing = [r for r in required_args if r not in explicit]
                 if missing:
                     logger.warning(
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -11,7 +11,7 @@ from collections import defaultdict
 from . import digest
 from . import state
 from . import metadata
-from .call import Call, AnyCall, QueryCall
+from .call import Call, AnyCall, QueryCall, bind
 from .caches import Rejected, BaseCache, RefreshingCache
 
 
@@ -290,14 +290,7 @@ def fleche(
                 return func(*args, **kwargs)
 
             if required_args and sig is not None:
-                # Determine which required args were explicitly provided, without
-                # re-binding.  Positional args map to their parameter names by index;
-                # keyword args are in kwargs directly.
-                positional_params = [
-                    name for name, p in sig.parameters.items()
-                    if p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD)
-                ]
-                explicit = set(positional_params[:len(args)]) | set(kwargs.keys())
+                explicit = set(bind(func, args, kwargs).keys())
                 missing = [r for r in required_args if r not in explicit]
                 if missing:
                     logger.warning(
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -97,14 +97,6 @@ def process_ignore_required_args(
         require = tuple(require)
     required_args = require + tuple(type_required)
 
-    try:
-        sig = signature(func)
-        for r in required_args:
-            if r in sig.parameters and sig.parameters[r].kind == sig.parameters[r].POSITIONAL_ONLY:
-                logger.warning("Argument '%s' is marked as Required but is positional-only. Required only works for keyword arguments.", r)
-    except (TypeError, ValueError):
-        pass
-
     return ignored_args, required_args
 
 
@@ -164,6 +156,11 @@ def fleche(
             func.__version__ = version  # ty: ignore
 
         ignored_args, required_args = process_ignore_required_args(func, ignore, require)
+
+        try:
+            sig = signature(func)
+        except (TypeError, ValueError):
+            sig = None
 
         @wraps(func)
         def get_call(*args, **kwargs):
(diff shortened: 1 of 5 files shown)


pmrv:
This is wrong.  Should use Call.from_call if at all possible.  Should keep the warning around.  It warns against users writing

```python
def foo(a: Required[int], b: float):...
```

as that makes no sense.  Positional arguments always have to passed and so are always present for the cache case.

src/fleche/wrapper.py
```diff
@@ -274,22 +271,30 @@ def wrapper(*args: Any, **kwargs: Any) -> _T:
...
                 logger.warning("No hash for argument: %s", e.args[0])
                 return func(*args, **kwargs)
+
+            if required_args and sig is not None:
+                # Use sig.bind without apply_defaults: only explicitly-provided arguments
+                # appear in bound.arguments, regardless of whether they were positional
+                # or keyword.  This fixes the bug where positional required args were
+                # incorrectly treated as missing.
+                explicit = set(sig.bind(*args, **kwargs).arguments.keys())
+                missing = [r for r in required_args if r not in explicit]
```
pmrv: There is no point in recalculating this.  You already have access to `call`.  Just check its arguments, rather than binding again.

src/fleche/wrapper.py
```diff
@@ -97,14 +97,6 @@ def is_required(hint):
...
     required_args = require + tuple(type_required)
 
-    try:
-        sig = signature(func)
-        for r in required_args:
-            if r in sig.parameters and sig.parameters[r].kind == sig.parameters[r].POSITIONAL_ONLY:
-                logger.warning("Argument '%s' is marked as Required but is positional-only. Required only works for keyword arguments.", r)
-    except (TypeError, ValueError):
-        pass
-
```
pmrv, line 107: This needs to stay.  See my [comment](https://github.com/pmrv/fleche/pull/309#issuecomment-4240529893).

pmrv (commented):
@claude  Fix my complaints.

pmrv:
**Claude finished @pmrv's task in 3s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24398203234)

---
### Fixed review complaints

- [x] Read `wrapper.py` and test files
- [x] Fix line 107: restore positional-only `Required` warning in `process_ignore_required_args`
- [x] Fix line 287: avoid re-binding with `sig.bind()` — compute `explicit` directly from `sig.parameters` + `args`/`kwargs`
- [x] Restore deleted test `test_process_positional_only_required_warning`
- [x] Run tests — 734 passed, 4 skipped
- [x] Commit and push

---

### What was done

**`process_ignore_required_args` (line 107):** Restored the `POSITIONAL_ONLY` warning block that was removed. It warns when a user marks a positional-only argument as `Required`, which is semantically meaningless since positional-only args always have to be provided.

**`wrapper()` required check (line 287):** Replaced the second `sig.bind(*args, **kwargs)` call with a direct computation:
```python
positional_params = [
    name for name, p in sig.parameters.items()
    if p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD)
]
explicit = set(positional_params[:len(args)]) | set(kwargs.keys())
```
This maps positional `args[i]` to their parameter names in signature order, then unions with `kwargs.keys()`. Gives the same result as `sig.bind(...).arguments` (without `apply_defaults`) but without a second bind call.

All 734 tests pass.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-297-20260414-0014)

src/fleche/wrapper.py
```diff
@@ -274,22 +282,33 @@ def wrapper(*args: Any, **kwargs: Any) -> _T:
...
                 call = get_call(*args, **kwargs)
                 key = call.to_lookup_key()
-                result = cache.load(key).result
-                logger.debug("Cache hit for %s with key %s", call.name, key)
-                return result
             except digest.Unhashable as e:
                 logger.warning("No hash for argument: %s", e.args[0])
                 return func(*args, **kwargs)
+
+            if required_args and sig is not None:
```
pmrv: @claude Not sure why you like to be dense here.  In line 286 we already define call (of type Call) with attribute .arguments.  This dictionary is a result of a bind call.  Use it in place of `explicit` in line 301 below.

pmrv: **Claude finished @pmrv's task in 3m 56s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24409176237)

---
### Review: `call.arguments` cannot replace `explicit` here

I tried using `call.arguments` in place of `explicit` — it breaks 4 tests:

- `test_required_invariant[required_via_decorator]`
- `test_required_invariant[required_via_hint]`
- `test_required_default_positional_caches`
- `test_combined_hints_and_args`

**Root cause:** `Call.from_call` (in `call.py:31`) calls `bound.apply_defaults()` before building `call.arguments`. This means **every** parameter ends up in `call.arguments`, including those not explicitly passed by the caller (they receive their default values). So `r not in call.arguments` is always `False` for any `Required` arg that has a default — the warning is never triggered and everything caches regardless.

Example: for `def foo(a, b: Required[Any] = None)`, calling `foo(1)` produces `call.arguments = {'a': 1, 'b': None}` — `b` appears with its default, so the required check passes silently.

The `explicit` set correctly captures only what the caller actually supplied:

```python
positional_params = [name for name, p in sig.parameters.items()
                     if p.kind in (p.POSITIONAL_ONLY, p.POSITIONAL_OR_KEYWORD)]
explicit = set(positional_params[:len(args)]) | set(kwargs.keys())
```

It maps positional `args[i]` → parameter names by index, unions with `kwargs.keys()`, and does not involve a second `sig.bind` call.

**Is `apply_defaults()` intentional?** Yes — it's needed for stable cache keys (so `foo(1)` and `foo(1, b=None)` hash the same). But that means `call.arguments` can't also serve as an "explicitly-passed" sentinel.

If you'd like to remove the `positional_params` intermediate variable, the expression can be inlined — just let me know. Or if the intent is to change `Required` semantics so defaults are treated as "provided", I can update the tests too; please clarify.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-297-20260414-0014)

pmrv (commented):
@claude I see your reasoning with the apply_default messing things up.  OTOH, QueryCall does it with that.  On the third hand, using it here would be wrong.  Do this:
1. factor out a bind function taking func, args, kwargs, and optional parameter to apply defaults.  This is a very thin wrapper around the signature.bind method.
2. use that guy in the two call classes and use it here.

pmrv:
**Claude finished @pmrv's task in 8m 33s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24436055595)

---
### Factored out `bind` helper

- [x] Read `call.py` and `wrapper.py`
- [x] Factor out `bind(func, args, kwargs, apply_defaults=False, partial=False)` helper in `call.py`
- [x] Update `Call.from_call` to use new `bind` helper
- [x] Update `QueryCall.from_call` to use new `bind` helper
- [x] Update `wrapper.py` required check to use new `bind` helper (without `apply_defaults`)
- [x] Run tests — 734 passed, 4 skipped
- [x] Commit and push

---

### What was done

**`call.py` — new `bind` helper (module-level, exported):**
```python
def bind(func, args, kwargs, apply_defaults=False, partial=False):
    sig = signature(func)
    if partial:
        bound = sig.bind_partial(*args, **kwargs)
    else:
        bound = sig.bind(*args, **kwargs)
    if apply_defaults:
        bound.apply_defaults()
    return bound.arguments
```
Added `partial` alongside `apply_defaults` so `QueryCall` (which needs `bind_partial`) can also use it cleanly.

**`Call.from_call`:** now a one-liner `dict(bind(func, args, kwargs, apply_defaults=True))`.

**`QueryCall.from_call`:** uses `bind(func, args, kwargs, partial=True)` then fills `None` for unspecified params (wildcards) by iterating `signature(func).parameters`.

**`wrapper.py` required check:** replaced the manual `positional_params` / `explicit` computation with `set(bind(func, args, kwargs).keys())` — `sig.bind` without `apply_defaults` returns exactly what the caller explicitly supplied (positional args mapped to their names by index, kwargs by name), so the semantics are identical but the logic lives in one place.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-297-20260414-0014)

src/fleche/wrapper.py
```diff
@@ -165,6 +168,11 @@ def decorator(func: Callable[..., _T]) -> Callable[..., _T]:
 
         ignored_args, required_args = process_ignore_required_args(func, ignore, require)
 
+        try:
+            sig = signature(func)
+        except (TypeError, ValueError):
+            sig = None
+
```
pmrv: @claude you dont need that now anymore

pmrv: ### Removed redundant `sig` variable

- [x] Read wrapper.py to identify what at line 175 is no longer needed
- [x] Remove the redundant `sig` try/except block and update the guard
- [x] Run tests — 734 passed, 4 skipped
- [x] Commit and push

---

**What was done:**

Removed the `try: sig = signature(func) / except: sig = None` block (old lines 171–175) from the decorator. This variable was only used for the `sig is not None` guard on the required-args check.

The guard is now unnecessary because:
- `bind(func, args, kwargs)` (line 293) internally calls `signature(func)` — so any failure would have already surfaced when `Call.from_call` ran in `get_call()` earlier in the same `try` block.
- Changed `if required_args and sig is not None:` → `if required_args:`.

The `signature` import is kept since it's still used in `process_ignore_required_args`.

[View job run](https://github.com/pmrv/fleche/actions/runs/24439144971) | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-297-20260414-0014)

tests/unit/fleche/test_type_hints.py
```diff
@@ -96,10 +89,40 @@ def foo(a: Required, b=1):
 
     caplog.clear()
     with caplog.at_level(logging.WARNING):
-        foo(2) # Positional a, should NOT cache
+        foo(2)  # a provided positionally — should cache
+        assert len(calls_storage.storage) == 1
+        assert "Missing required keyword arguments" not in caplog.text
+
+
+def test_required_default_positional_caches(memory_cache, caplog):
```
pmrv: @claude Move to regression tests.

pmrv: **Claude finished @pmrv's task in 2m 1s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24439189363)

---
### Moved test to regression tests

- [x] Read `tests/unit/fleche/test_type_hints.py` to identify the test at line 97
- [x] Find the regression tests directory (`tests/regression/`)
- [x] Create `tests/regression/test_issue_297.py` with `test_required_default_positional_caches`
- [x] Remove the test from `test_type_hints.py`
- [x] Run tests — 9 passed
- [x] Commit and push

Moved `test_required_default_positional_caches` from `tests/unit/fleche/test_type_hints.py` to a new `tests/regression/test_issue_297.py` file. The new file includes its own `memory_cache` fixture (same as in `test_type_hints.py`) since that fixture isn't shared via `conftest.py`.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-297-20260414-0014)

PR #312, opened 2026-04-14
feat(query): auto-convert timestart/timestop to datetime in table()

### LINKED ISSUE #311: table() should automatically convert timestart/stop into human readable format
Example:

```
tf = smc.run.query().table()
tf.timestart = pd.to_datetime(tf.timestart * 1000000000)
```

pmrv:
Automatically converts `timestart` and `timestop` columns (produced by the `Runtime` metadata) from Unix timestamps to `pandas.Timestamp` in `QueryIterator.table()`, so callers no longer need to do the conversion manually.

Closes #311

### DIFF
--- a/src/fleche/query.py
+++ b/src/fleche/query.py
@@ -34,8 +34,8 @@ class QueryIterator(Iterable[call.LazyCall]):
         Only requested arguments are loaded from cache.
 
         ``timestart`` and ``timestop`` columns (produced by the :class:`~fleche.metadata.Runtime`
-        metadata) are automatically converted from Unix timestamps (float seconds) to
-        :class:`pandas.Timestamp` for human-readable display.
+        metadata) are automatically converted from UTC Unix timestamps (float seconds) to
+        timezone-aware :class:`pandas.Timestamp` objects in the local timezone.
 
         Args:
             arguments (iterable of str): add the given arguments (of the queried calls) as columns to the table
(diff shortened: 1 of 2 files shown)


pmrv:
@claude Check also how we dealing with timezone.  Expected behavior is that storage always saves things in UTC (and MetaData should just return that), while table loads into the local timezone.

pmrv:
**Claude finished @pmrv's task in 1m 56s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24410882561)

---
### Timezone Review

- [x] Gather context
- [x] Check Runtime metadata — does it store timestamps in UTC?
- [x] Check table() conversion — does it load into local timezone?
- [x] Implement fix and update tests

---

**Findings:**

**Storage (UTC)** — `Runtime.pre()` uses `time.time()` (`metadata.py:85`), which returns a UTC-based Unix epoch float. ✅ No change needed.

**table() (local timezone)** — The original code used:
```python
df[col] = pd.to_datetime(df[col], unit="s")
```
This produces **timezone-naive** timestamps. While the underlying value is correct (UTC epoch), the `Timestamp` has no `tz` attached and is not in local time. ❌

**Fix applied** (`query.py:74`):
```python
df[col] = pd.to_datetime(df[col], unit="s", utc=True).dt.tz_convert("localtime")
```
- `utc=True` marks the parsed timestamps as UTC-aware
- `.dt.tz_convert("localtime")` converts to the system's local timezone

The test was also updated to assert that the resulting `Timestamp` columns are timezone-aware (`df["timestart"].dt.tz is not None`).

All 22 tests pass. ✅
 · [Branch](https://github.com/pmrv/fleche/tree/claude/issue-311-20260414-1619)

PR #318, opened 2026-04-17
Fix Sql race condition by integrating PerKeyLockMixin

### LINKED ISSUE #218: `Sql._save()` check-evict-insert is not serialized across sessions
**File:** `src/fleche/storage/sql.py:158-165`

`_save()` reads, optionally deletes, then inserts — all across separate SQLAlchemy sessions with no row-level locking. Two threads can both read the same existing row, both delete it in their sessions, then both try to insert the new row — causing `IntegrityError` or silent loss of one result. Additionally, `create_engine` for SQLite doesn't set `connect_args={"check_same_thread": False}`, which may raise `ProgrammingError` when the engine is created on one thread and sessions used on another.

**MWE:**
```python
import concurrent.futures, contextvars, tempfile
from pathlib import Path
import fleche
from fleche.storage.sql import Sql

@fleche.fleche
def compute(x):
    return x * 2

with tempfile.TemporaryDirectory() as tmp:
    sql = Sql(str(Path(tmp) / "cache.db"))
    mem = Memory({})
    cache = Cache(mem, sql)  # sql as CallStorage, Memory for values

    with fleche.cache(cache):
            futures = [pool.submit(ctx.run, compute, i % 3) for i in range(12)]
            results = [f.result() for f in futures]
    # May raise IntegrityError or OperationalError: database is locked
```

pmrv:
Integrates `PerKeyLockMixin` into `Sql` to fix the check-evict-insert race condition described in #218.

Changes:
- `Sql(PerKeyLockMixin, CallStorage)` serializes save operations per-key
- `check_same_thread=False` for SQLite connections
- `__getstate__`/`__setstate__` updated for lock field pickling
- Regression test for concurrent saves

Closes #218

### DIFF
--- a/src/fleche/storage/thread_safe.py
+++ b/src/fleche/storage/thread_safe.py
@@ -47,11 +47,8 @@ class _PicklableLock:
     def __init__(self):
         self._lock = threading.Lock()
 
-    def __getstate__(self):
-        return {}
-
-    def __setstate__(self, state):
-        self.__init__()
+    def __reduce__(self):
+        return (type(self), ())
 
     def __enter__(self):
         return self._lock.__enter__()
@@ -70,11 +67,8 @@ class _PicklableRLock:
     def __init__(self):
         self._lock = threading.RLock()
 
-    def __getstate__(self):
-        return {}
-
-    def __setstate__(self, state):
-        self.__init__()
+    def __reduce__(self):
+        return (type(self), ())
 
     def __enter__(self):
         return self._lock.__enter__()
--- a/src/fleche/storage/thread_safe.py
+++ b/src/fleche/storage/thread_safe.py
@@ -34,6 +34,55 @@ from .base import KeyManagement
 from ..digest import Digest
 
 
+class _PicklableLock:
+    """A ``threading.Lock`` wrapper that survives pickle round-trips.
+
+    The lock is re-initialised fresh on unpickle — its acquired/released state
+    is **not** preserved.  This is intentionally an in-process pickling aid
+    (e.g. for ``multiprocessing`` spawn or ``joblib``), **not** an
+    inter-process synchronisation primitive: each process gets its own
+    independent lock that shares no state with locks in other processes.
+    """
+
+    def __init__(self):
+        self._lock = threading.Lock()
+
+    def __getstate__(self):
+        return {}
+
+    def __setstate__(self, state):
+        self.__init__()
+
+    def __enter__(self):
+        return self._lock.__enter__()
+
+    def __exit__(self, *args):
+        return self._lock.__exit__(*args)
+
+
+class _PicklableRLock:
+    """A ``threading.RLock`` wrapper that survives pickle round-trips.
+
+    Same in-process-only semantics as :class:`_PicklableLock`; reentrant so
+    that nested acquisitions (e.g. ``expand`` inside ``load``) do not deadlock.
+    """
+
+    def __init__(self):
+        self._lock = threading.RLock()
+
+    def __getstate__(self):
+        return {}
+
+    def __setstate__(self, state):
+        self.__init__()
+
+    def __enter__(self):
+        return self._lock.__enter__()
+
+    def __exit__(self, *args):
+        return self._lock.__exit__(*args)
+
+
 @dataclass(frozen=True)
 class SerializingMixin(KeyManagement):
     """Mixin that serializes all storage operations behind a single reentrant lock.
@@ -44,8 +93,8 @@ class SerializingMixin(KeyManagement):
         class SerializingValueMemory(SerializingMixin, ValueMemory): ...
     """
 
-    _lock: threading.RLock = field(
-        default_factory=threading.RLock, init=False, repr=False, compare=False
+    _lock: _PicklableRLock = field(
+        default_factory=_PicklableRLock, init=False, repr=False, compare=False
     )
 
     @contextlib.contextmanager
@@ -74,8 +123,8 @@ class PerKeyLockMixin(KeyManagement):
     _key_locks: weakref.WeakValueDictionary[Digest | str, threading.RLock] = field(
         default_factory=weakref.WeakValueDictionary, init=False, repr=False, compare=False
     )
-    _meta_lock: threading.Lock = field(
-        default_factory=threading.Lock, init=False, repr=False, compare=False
+    _meta_lock: _PicklableLock = field(
+        default_factory=_PicklableLock, init=False, repr=False, compare=False
     )
 
     def _get_key_lock(self, key: Digest | str) -> threading.RLock:
--- a/src/fleche/storage/sql.py
+++ b/src/fleche/storage/sql.py
@@ -1,8 +1,11 @@
 import logging
+import threading
+import weakref
 from typing import Iterable, Any, List
 from pathlib import Path
 from dataclasses import dataclass, field
 from .base import KeyManagement, CallStorage, AmbiguousDigestError
+from .thread_safe import PerKeyLockMixin
 from ..call import Call, QueryCall
 from ..digest import Digest, DIGEST_LENGTH, digest
 
@@ -123,7 +126,7 @@ def _enable_sqlite_foreign_keys(engine) -> None:
 
 
 @dataclass(frozen=True)
-class Sql(CallStorage):
+class Sql(PerKeyLockMixin, CallStorage):
     """SQLAlchemy-backed CallStorage with JSON metadata and DB-backed expand()."""
 
     url: str | None = None
@@ -137,7 +140,8 @@ class Sql(CallStorage):
         coerced_url = _coerce_sqlite_url(self.url)
         assert coerced_url is not None
         object.__setattr__(self, "url", coerced_url)
-        engine = create_engine(coerced_url, echo=self.echo, future=True)
+        connect_args = {"check_same_thread": False} if coerced_url.startswith("sqlite:") else {}
+        engine = create_engine(coerced_url, echo=self.echo, future=True, connect_args=connect_args)
         _enable_sqlite_foreign_keys(engine)
         Base.metadata.create_all(engine)
         object.__setattr__(self, "engine", engine)
@@ -149,12 +153,16 @@ class Sql(CallStorage):
         state = self.__dict__.copy()
         state.pop("engine", None)
         state.pop("session", None)
+        state.pop("_key_locks", None)
+        state.pop("_meta_lock", None)
         return state
 
     def __setstate__(self, state):
         self.__dict__.update(state)
         # Re-initialize unpickleable fields
         self.__post_init__()
+        self.__dict__["_key_locks"] = weakref.WeakValueDictionary()
+        self.__dict__["_meta_lock"] = threading.Lock()
 
     def put(self, call: Any, key: Digest) -> Digest:
         session = self.session()
--- a/src/fleche/storage/sql.py
+++ b/src/fleche/storage/sql.py
@@ -1,6 +1,4 @@
 import logging
-import threading
-import weakref
 from typing import Iterable, Any, List
 from pathlib import Path
 from dataclasses import dataclass, field
@@ -149,20 +147,10 @@ class Sql(PerKeyLockMixin, CallStorage):
             bind=engine, expire_on_commit=False, future=True
         ))
 
-    def __getstate__(self):
-        state = self.__dict__.copy()
-        state.pop("engine", None)
-        state.pop("session", None)
-        state.pop("_key_locks", None)
-        state.pop("_meta_lock", None)
-        return state
-
-    def __setstate__(self, state):
-        self.__dict__.update(state)
-        # Re-initialize unpickleable fields
-        self.__post_init__()
-        self.__dict__["_key_locks"] = weakref.WeakValueDictionary()
-        self.__dict__["_meta_lock"] = threading.Lock()
+    def __reduce__(self):
+        # Constructor args are the complete durable state; __post_init__ rebuilds
+        # engine/session and the mixin's lock fields are self-contained picklable types.
+        return (type(self), (self.url, self.echo))
 
     def put(self, call: Any, key: Digest) -> Digest:
         session = self.session()
--- a/tests/regression/test_sql_concurrent_save.py
+++ b/tests/regression/test_sql_concurrent_save.py
@@ -1,0 +1,33 @@
+"""Regression test for issue #218: Sql._save() race condition under concurrent access."""
+
+import concurrent.futures
+
+from fleche.call import Call
+from fleche.storage.sql import Sql
+
+
+def test_sql_concurrent_save_no_integrity_error(tmp_path):
+    """Multiple threads saving calls with overlapping keys must not raise IntegrityError.
+
+    Threads save calls whose lookup keys collide (only 3 distinct inputs) so the
+    check-evict-insert path in save() is exercised under genuine data races.
+    """
+    sql = Sql(str(tmp_path / "cache.db"))
+
+    errors = []
+
+    def save_call(x):
+        try:
+            call = Call(name="compute", arguments={"x": x}, result=None)
+            sql.save(call)
+        except Exception as exc:
+            errors.append(exc)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=8) as pool:
+        futures = [pool.submit(save_call, i % 3) for i in range(24)]
+        for f in futures:
+            f.result()
+
+    assert not errors, f"Concurrent saves raised errors: {errors}"
+    stored_keys = list(sql.list())
+    assert len(stored_keys) == 3, f"Expected 3 unique keys, got {len(stored_keys)}"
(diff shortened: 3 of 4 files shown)


pmrv:
@claude I don't like that the final class needs to know how to manage the mixin's state.  Workshop some solutions and present them with pro/cons.

pmrv:
**Claude finished @pmrv's task in 2m 48s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24581430699)

---
### Workshopping `PerKeyLockMixin` state management solutions

- [x] Gather context
- [x] Read current implementation
- [x] Workshop solutions with pros/cons

---

### Problem statement

`Sql.__getstate__`/`__setstate__` currently reference `_key_locks` and `_meta_lock` — private internals of `PerKeyLockMixin`. If the mixin renames or restructures those fields, `Sql` silently breaks:

```python
# sql.py – the coupling in question
def __getstate__(self):
    state = self.__dict__.copy()
    state.pop("engine", None)
    state.pop("session", None)
    state.pop("_key_locks", None)   # ← mixin internals leak here
    state.pop("_meta_lock", None)   # ← and here
    return state

def __setstate__(self, state):
    self.__dict__.update(state)
    self.__post_init__()
    self.__dict__["_key_locks"] = weakref.WeakValueDictionary()  # ← mixin internals
    self.__dict__["_meta_lock"] = threading.Lock()                # ← mixin internals
```

Three approaches to push the mixin's state knowledge back into the mixin:

---

### Option A — Cooperative pickle chaining

Add `__getstate__`/`__setstate__` to `PerKeyLockMixin` itself. `Sql` chains via `super()` and only touches its own fields.

```python
# thread_safe.py
@dataclass(frozen=True)
class PerKeyLockMixin(KeyManagement):
    # ... existing fields ...

    def __getstate__(self):
        state = self.__dict__.copy()
        state.pop("_key_locks", None)
        state.pop("_meta_lock", None)
        return state

    def __setstate__(self, state):
        self.__dict__.update(state)
        self.__dict__["_key_locks"] = weakref.WeakValueDictionary()
        self.__dict__["_meta_lock"] = threading.Lock()


# sql.py — no more mixin field names
def __getstate__(self):
    state = super().__getstate__()   # mixin strips its own fields
    state.pop("engine", None)
    state.pop("session", None)
    return state

def __setstate__(self, state):
    super().__setstate__(state)      # mixin restores its own fields
    self.__post_init__()             # Sql restores engine/session
```

**Pros:**
- Minimal change surface — two methods added to the mixin, two lines changed in `Sql`
- Conventional Python pickle protocol; immediately readable to anyone familiar with `__getstate__`/`__setstate__`
- Works automatically for any future class that mixes in `PerKeyLockMixin`
- `SerializingMixin` gets the same treatment for free (its `_lock` field has the same problem today)

**Cons:**
- `PerKeyLockMixin.__getstate__` seeds the state dict from `self.__dict__.copy()`, which captures all fields across the whole object — it just happens to only strip its own. This "starts from the full dict" responsibility placed on the base of the chain is a mild footgun if another mixin tries to do the same
- Requires both sides to consistently remember to call `super()` — a contract that must be documented but can't be statically enforced

---

### Option B — Picklable field wrappers

Make the mixin's fields picklable so `Sql` never needs to mention them. `threading.Lock` isn't picklable, but a thin wrapper can be. `weakref.WeakValueDictionary` already serializes as empty (which is exactly the right post-unpickle state).

```python
# thread_safe.py
class _PicklableLock:
    """threading.Lock that re-initializes on unpickle."""
    def __init__(self):
        self._lock = threading.Lock()
    def __getstate__(self): return {}
    def __setstate__(self, _): self.__init__()
    def __enter__(self): return self._lock.__enter__()
    def __exit__(self, *a): return self._lock.__exit__(*a)

@dataclass(frozen=True)
class PerKeyLockMixin(KeyManagement):
    _key_locks: weakref.WeakValueDictionary = field(
        default_factory=weakref.WeakValueDictionary, init=False, repr=False, compare=False
    )
    _meta_lock: _PicklableLock = field(
        default_factory=_PicklableLock, init=False, repr=False, compare=False
    )
    # No __getstate__/__setstate__ needed


# sql.py — no mixin field names at all
def __getstate__(self):
    state = self.__dict__.copy()
    state.pop("engine", None)
    state.pop("session", None)
    return state

def __setstate__(self, state):
    self.__dict__.update(state)
    self.__post_init__()
```

**Pros:**
- `Sql` has zero knowledge of the mixin's internals — the strongest decoupling of the three
- No `super()` chaining convention to remember or enforce
- The mixin's pickle semantics live entirely in the mixin's field types

**Cons:**
- Introduces a new `_PicklableLock` class (or similar) — more types to understand and test
- `WeakValueDictionary` silently losing all entries on pickle is correct, but surprising to someone reading the code for the first time (worth a comment)
- `_get_key_lock` calls `self._meta_lock` which is now a `_PicklableLock` — need to make sure the call-site signature matches (`with self._meta_lock:` still works if `_PicklableLock` is a context manager)

---

### Option C — Declarative transient-field protocol

Introduce a `_transient_field_names` class variable and a chainable `_reinit_transient()` hook. Each class declares its own unpicklable fields and reinit logic; a shared base handles `__getstate__`/`__setstate__` once.

```python
# base.py (or thread_safe.py)
class TransientMixin:
    _transient_field_names: tuple[str, ...] = ()

    def __getstate__(self):
        state = self.__dict__.copy()
        for cls in type(self).__mro__:
            for name in getattr(cls, "_transient_field_names", ()):
                state.pop(name, None)
        return state

    def __setstate__(self, state):
        self.__dict__.update(state)
        self._reinit_transient()

    def _reinit_transient(self):
        pass  # base no-op; subclasses chain via super()


# thread_safe.py
@dataclass(frozen=True)
class PerKeyLockMixin(TransientMixin, KeyManagement):
    _transient_field_names = ("_key_locks", "_meta_lock")

    def _reinit_transient(self):
        self.__dict__["_key_locks"] = weakref.WeakValueDictionary()
        self.__dict__["_meta_lock"] = threading.Lock()
        super()._reinit_transient()


# sql.py
@dataclass(frozen=True)
class Sql(PerKeyLockMixin, CallStorage):
    _transient_field_names = ("engine", "session")

    def _reinit_transient(self):
        self.__post_init__()
        super()._reinit_transient()

    # No __getstate__/__setstate__ at all
```

**Pros:**
- `__getstate__`/`__setstate__` are written exactly once — in `TransientMixin`
- Adding a new mixin with transient fields is purely declarative: add `_transient_field_names` and override `_reinit_transient`
- The MRO walk in `__getstate__` makes the field collection explicit and correct for any depth of inheritance

**Cons:**
- Most infrastructure of the three — adds a new `TransientMixin` class and a new protocol (`_transient_field_names` + `_reinit_transient`)
- `_reinit_transient` still requires `super()` chaining, which is the same footgun as Option A but now for a non-standard protocol
- Overkill if `PerKeyLockMixin` is the only mixin that ever needs this treatment; pays for generality that may never be used

---

### Recommendation

**Option A** is the pragmatic pick: minimal change, conventional protocol, directly solves the stated problem. **Option B** is the conceptually cleanest if the `_PicklableLock` wrapper is acceptable — zero leakage from mixin into consumer. **Option C** only makes sense if you anticipate several more transient-field mixins in the future.

pmrv:
@claude Let's converge on B, but pontificate a bit first why and whether at all `Sql` and the mixin need to manage their own state.  Is there no way we can leverage from dataclass features here to avoid this?

pmrv:
**Claude finished @pmrv's task in 2m 18s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24581963890)

---
### Dataclass pickle mechanics and the path to Option B

- [x] Gather context
- [x] Read current sql.py and thread_safe.py
- [x] Analyse why state management is needed and whether dataclasses can help
- [x] Evaluate Option B in that light

---

### Why does the state management exist at all?

Python's pickle protocol **bypasses `__init__`**. It calls `type.__new__(cls)` and then `__setstate__`. The `field(default_factory=..., init=False)` declaration in a dataclass is purely an `__init__`-time instruction — it tells the generated `__init__` to call the factory and assign the result. Pickle never sees it.

So `init=False` fields with factories like `threading.Lock` or `weakref.WeakValueDictionary` end up in one of two bad states after a pickle round-trip, depending on whether the field type is picklable:

| Field | Picklable? | Effect without custom state management |
|---|---|---|
| `_meta_lock: threading.Lock` | ❌ | `pickle.dumps()` raises `TypeError` |
| `_key_locks: WeakValueDictionary` | ✓ (serialises empty) | Restored correctly (empty is correct) |
| `engine` (SQLAlchemy) | ❌ | `pickle.dumps()` raises `TypeError` |
| `session` (SQLAlchemy) | ❌ | `pickle.dumps()` raises `TypeError` |

The `_meta_lock` and the SQLAlchemy objects are what force the hand. Everything else would survive pickle untouched.

### Can dataclass features avoid the problem natively?

No, and it's a known gap in the stdlib. The `field()` keyword arguments control exactly four things: `__init__`, `repr`, `compare`/`hash`, and `metadata`. There is no `field(transient=True)` and no proposal for one that was accepted. The pickle path is completely outside the dataclass machinery.

The closest "dataclass feature" that touches pickling is **`__reduce__`** (a standard Python pickle hook, not dataclass-specific). For a frozen dataclass whose only durable state is its constructor arguments, `__reduce__` can express exactly that:

```python
# sql.py
def __reduce__(self):
    return (type(self), (self.url, self.echo))
```

`pickle.loads(...)` will call `Sql(url, echo)`, which goes through the full `__init__` → `__post_init__` path. Every `init=False` field with a `default_factory` is re-created fresh — including the mixin's `_key_locks` and `_meta_lock` — **without `Sql` naming them**. The constructor argument list becomes the serialisation format, which is exactly what a dataclass is meant to encode.

This removes `__getstate__`/`__setstate__` from `Sql` entirely and makes the mixin coupling disappear. But it only works because `Sql`'s init args fully capture its persistent state. It's a coincidence of this class's design, not a general mechanism.

### Why Option B is still the right mixin-side fix

`__reduce__` on `Sql` solves the concrete-class pickling problem, but it doesn't make `PerKeyLockMixin` safe to use in other concrete classes that might forget to define `__reduce__`. The mixin currently exports a `threading.Lock` as a plain dataclass field, meaning **any concrete class that inherits it will blow up on `pickle.dumps()`** unless it handles the mixin's internals itself — the exact coupling problem the issue is about.

`_PicklableLock` (Option B) fixes this at the source: the field's *type* knows how to survive a pickle round-trip, so the mixin is self-contained. The concrete class doesn't need to know the field name or the lock type.

### Proposed combined implementation

`PerKeyLockMixin` gains `_PicklableLock`; `Sql` gains `__reduce__` instead of `__getstate__`/`__setstate__`. The two fixes are orthogonal and handle different scopes of the problem.

```python
# thread_safe.py

class _PicklableLock:
    """threading.Lock that reconstructs itself on unpickle."""
    def __init__(self):
        self._lock = threading.Lock()
    def __getstate__(self): return {}
    def __setstate__(self, _): self.__init__()
    def __enter__(self): return self._lock.__enter__()
    def __exit__(self, *a): return self._lock.__exit__(*a)

@dataclass(frozen=True)
class PerKeyLockMixin(KeyManagement):
    _key_locks: weakref.WeakValueDictionary = field(
        default_factory=weakref.WeakValueDictionary, init=False, repr=False, compare=False
    )
    _meta_lock: _PicklableLock = field(          # ← was threading.Lock
        default_factory=_PicklableLock, init=False, repr=False, compare=False
    )

    def _get_key_lock(self, key):
        with self._meta_lock:                    # _PicklableLock is a context manager
            lock = self._key_locks.get(key)
            if lock is None:
                lock = threading.RLock()
                self._key_locks[key] = lock
            return lock
    ...
```

```python
# sql.py

@dataclass(frozen=True)
class Sql(PerKeyLockMixin, CallStorage):
    url: str | None = None
    echo: bool = field(default=False, compare=False)
    engine: Any = field(init=False, repr=False, compare=False)
    session: Any = field(init=False, repr=False, compare=False)

    def __reduce__(self):                        # ← replaces __getstate__/__setstate__
        return (type(self), (self.url, self.echo))

    # __getstate__ and __setstate__ deleted
    ...
```

`SerializingMixin._lock` has the same `threading.RLock`-not-picklable problem and should receive the same `_PicklableLock` treatment.

### Summary

The state management in the current PR exists because pickle bypasses `__init__`, and `field()` is purely an `__init__` facility. Dataclasses offer no native `transient` concept. The cleanest exit is a two-pronged approach: make the mixin self-contained with a picklable lock wrapper (Option B, mixin-level), and express `Sql`'s serialisation intent with `__reduce__` (concrete-class level) — which is effectively saying "my constructor args are my serialised form", the most dataclass-aligned statement possible.

Shall I implement this?

pmrv:
@claude Go.  Document clearly the PicklableLock is a workaround for us and not interprocess serialization primitive

pmrv:
**Claude encountered an error** —— [View job](https://github.com/pmrv/fleche/actions/runs/24586377909)

---
### Implementing Option B with documented `_PicklableLock`

- [x] Gather context
- [x] Read current `thread_safe.py` and `sql.py`
- [ ] Add `_PicklableLock`/`_PicklableRLock` to `thread_safe.py` with clear "in-process only" docs
- [ ] Update `PerKeyLockMixin._meta_lock` and `SerializingMixin._lock` to use picklable wrappers
- [ ] Replace `Sql.__getstate__`/`__setstate__` with `__reduce__`
- [ ] Run tests
- [ ] Commit and push

pmrv:
@claude  Pick it up again.

pmrv:
**Claude finished @pmrv's task in 3s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24597983791)

---
### Option B implemented

- [x] Gather context
- [x] Read current `thread_safe.py` and `sql.py`
- [x] Add `_PicklableLock`/`_PicklableRLock` to `thread_safe.py` with "in-process only" docs
- [x] Update `PerKeyLockMixin._meta_lock` and `SerializingMixin._lock` to use picklable wrappers
- [x] Replace `Sql.__getstate__`/`__setstate__` with `__reduce__`
- [x] Run tests (743 passed, 4 skipped)
- [x] Commit and push

---

### What changed

**`thread_safe.py`** — two new wrapper classes, clearly documented as in-process-only:

```python
class _PicklableLock:
    """A threading.Lock wrapper that survives pickle round-trips.
    ...
    Warning: in-process only — NOT an inter-process synchronisation primitive.
    """
    def __getstate__(self): return {}
    def __setstate__(self, _): self.__init__()
    # __enter__/__exit__ delegate to self._lock

class _PicklableRLock:
    """Same but wraps threading.RLock (reentrant)."""
    ...
```

`PerKeyLockMixin._meta_lock` now uses `_PicklableLock`; `SerializingMixin._lock` now uses `_PicklableRLock`. Neither mixin needs `__getstate__`/`__setstate__` — the lock fields are self-contained.

**`sql.py`** — `__getstate__`/`__setstate__` replaced by the much simpler:

```python
def __reduce__(self):
    return (type(self), (self.url, self.echo))
```

Unpickling calls `Sql(url, echo)`, which goes through the full `__init__` → `__post_init__` path. Every `init=False` field (including the mixin's locks) is re-created automatically. `Sql` now has zero knowledge of the mixin's internals. `threading` and `weakref` imports removed from `sql.py` as they're no longer referenced there.

src/fleche/storage/thread_safe.py
```diff
@@ -34,6 +34,55 @@ class ThreadSafeValueMemory(SerializingMixin, ValueMemory): ...
...
+    """
+
+    def __init__(self):
+        self._lock = threading.Lock()
+
+    def __getstate__(self):
+        return {}
+
+    def __setstate__(self, state):
+        self.__init__()
```
pmrv: @claude Just overload reduce

src/fleche/storage/thread_safe.py
```diff
@@ -34,6 +34,55 @@ class ThreadSafeValueMemory(SerializingMixin, ValueMemory): ...
...
+    """
+
+    def __init__(self):
+        self._lock = threading.RLock()
+
+    def __getstate__(self):
+        return {}
+
+    def __setstate__(self, state):
+        self.__init__()
```
pmrv: Same

src/fleche/storage/thread_safe.py
pmrv: **Claude finished @pmrv's task in 2s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24598571020)

---
### Replacing `__getstate__`/`__setstate__` with `__reduce__` in lock wrappers

- [x] Gather context
- [x] Read `thread_safe.py`
- [x] Replace pickle methods in `_PicklableLock` and `_PicklableRLock`
- [x] Run tests (743 passed, 4 skipped)
- [x] Commit and push

---

Both `_PicklableLock` and `_PicklableRLock` now use `__reduce__` instead of `__getstate__`/`__setstate__`:

```python
def __reduce__(self):
    return (type(self), ())
```

`(type(self), ())` tells pickle to reconstruct by calling the constructor with no arguments — one method instead of two, and the intent ("reconstruct fresh") is stated directly.

PR #321, opened 2026-04-17
Add regression tests for table().index vs storage key consistency (issue #319)

### LINKED ISSUE #319: query().table() index doesn't match actual call storage keys
Feels like this is a regression

pmrv:
Adds `tests/regression/test_table_index_matches_storage_keys.py` with 14 tests that directly assert `set(cache.table().index) == set(str(k) for k in cache.calls.list())` across all backends and usage patterns.

Closes #319.


pmrv:
This doesn't seem to find the bug yet.

pmrv:
See #351  for the actual fix.

PR #346, opened 2026-04-18
Integrate SQL session lifecycle into _operation_context

### LINKED ISSUE #326: Refactor: wrap `Sql` session lifecycle in a context manager
## Location

`src/fleche/storage/sql.py` — every storage method (`put`, `get`, `_contains`, `list`, `expand`, `_evict`, `load`, `query`) opens and closes a SQLAlchemy session manually. The pattern appears ~8 times:

```python
session = self.session()
try:
    ...
    session.commit()      # in mutating methods
except Exception:
    session.rollback()
    raise
finally:
    session.close()
```

See specifically: `sql.py:159-201`, `203-232`, `234-244`, `246-252`, `254-287` (expand opens its own session inside `_operation_context`), `289-301`, `303-318`, `423-439`.

## Problem

- 200+ lines of boilerplate.
- Easy to forget `rollback()` on a new mutating method, or to leak a session on an early return.
- Inconsistent with the documented `_operation_context` pattern that the rest of the storage hierarchy uses for resource scoping.

## Suggested refactor

Introduce a single context manager:

```python
@contextlib.contextmanager
def _session_scope(self, *, commit: bool = False):
    session = self.session()
    try:
        yield session
        if commit:
            session.commit()
    except Exception:
        session.rollback()
        raise
    finally:
        session.close()
```

and rewrite each method as `with self._session_scope(commit=True) as session: ...`. Alternatively, integrate the session into `_operation_context` so the same helper that guards the locks also scopes the DB session.

## Pros

- Removes ~6 lines of ceremony per method.
- Consistent rollback-on-error semantics everywhere.
- Makes the data flow in each method obvious at a glance.

## Cons / risks

- The integrity of `_save()` / `_evict()` under errors must be re-verified. Easy to cover with existing tests.
- Entangling this with `_operation_context` could make the locking story subtler — stick to a dedicated `_session_scope` first.
- Related to threadsafety issues #213 / #217 / #218 — fixing those may dictate a particular session shape, so this refactor should ideally land alongside or after a decision on the session-per-op vs. session-per-transaction model.

pmrv:
Override _operation_context in Sql to create/close a SQLAlchemy session (stored in threading.local) for every operation, chaining to super() so the per-key lock from PerKeyLockMixin is still applied. Re-entrant calls on the same thread reuse the active session. Rollback on exception and close in finally are now handled in one place instead of repeating try/except/finally boilerplate across every method.

Removes ~6 lines of ceremony from put, get, _contains, _evict, and expand. list() and query() wrap themselves with _operation_context("") to get a session without a meaningful per-key concern.

Closes #326

### DIFF
--- a/src/fleche/storage/sql.py
+++ b/src/fleche/storage/sql.py
@@ -156,23 +156,26 @@ class Sql(PerKeyLockMixin, CallStorage):
         return (type(self), (self.url, self.echo))
 
     @contextlib.contextmanager
-    def _operation_context(self, key):
+    def _session_context(self):
         if getattr(self._local, "session", None) is not None:
-            # Re-entrant call on the same thread: reuse the active session.
-            with super()._operation_context(key):
-                yield
+            yield
             return
         session = self.session()
         self._local.session = session
         try:
-            with super()._operation_context(key):
-                yield
+            yield
         except Exception:
             session.rollback()
             raise
         finally:
             session.close()
             self._local.session = None
+
+    @contextlib.contextmanager
+    def _operation_context(self, key):
+        with self._session_context():
+            with super()._operation_context(key):
+                yield
 
     def put(self, call: Any, key: Digest) -> Digest:
         session = self._local.session
@@ -248,7 +251,7 @@ class Sql(PerKeyLockMixin, CallStorage):
         )
 
     def list(self) -> Iterable[Digest]:
-        with self._operation_context(""):
+        with self._session_context():
             return [Digest(row[0]) for row in self._local.session.execute(select(CallModel.key))]
 
     def expand(self, key: Digest | str) -> Digest:
@@ -407,7 +410,7 @@ class Sql(PerKeyLockMixin, CallStorage):
         Yields:
             Call: Matching calls including their decoded metadata.
         """
-        with self._operation_context(""):
+        with self._session_context():
             stmt = select(CallModel.key).select_from(CallModel)
 
             conditions = self._build_call_conditions(template)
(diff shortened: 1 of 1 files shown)


src/fleche/storage/sql.py
```diff
@@ -413,8 +407,7 @@ def query(self, template: QueryCall) -> Iterable[Call]:
         Yields:
             Call: Matching calls including their decoded metadata.
         """
-        session = self.session()
-        try:
+        with self._operation_context(""):
```
pmrv: @claude This will also trigger the lock from the mixin and serialize all queries to the database.  Let's make a private method instead that does the session handling, use that one here, and also in the operations_context.

pmrv: **Claude finished @pmrv's task in 1m 22s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24603228855)

---
### Session handling refactor complete

- [x] Gather context and understand request
- [x] Read `src/fleche/storage/sql.py` to understand current structure
- [x] Create private `_session_context` method for session-only handling
- [x] Refactor `_operation_context` to use `_session_context`
- [x] Update `list()` and `query()` to use `_session_context` directly (no lock)
- [x] Commit and push changes

Extracted session lifecycle into `_session_context` — a re-entrancy-aware context manager that creates/closes the SQLAlchemy session and handles rollback, with no locking. `_operation_context` now composes both: `_session_context` for the session then `super()._operation_context(key)` for the per-key lock. `list()` and `query()` call `_session_context` directly, so they no longer acquire any lock.

PR #348, opened 2026-04-18
Add PerKeyLockMixin to FileStorage for in-process thread safety

### LINKED ISSUE #214: `FileStorage` advisory write lock is not thread-safe within a process
**File:** `src/fleche/storage/file.py:23`

`file_write_lock` creates the lock file with `lock_path.write_text(...)`. This is not an exclusive atomic operation: two threads in the same process can both call `write_text()` and both proceed past the context manager simultaneously, writing to the same data file concurrently and corrupting it.

**MWE:**
```python
import concurrent.futures, contextvars, tempfile
from pathlib import Path
import fleche
from fleche.storage.pickle_file import PickleFile

@fleche.fleche
def compute(x):
    return x * 2

with tempfile.TemporaryDirectory() as tmp:
    store = PickleFile.with_pickle(root=Path(tmp) / "values")
    calls = Memory({})
    cache = Cache(store, calls)

    with fleche.cache(cache):
            futures = [pool.submit(ctx.run, compute, i) for i in range(20)]
            results = [f.result() for f in futures]
    # May see corrupted pickles or KeyError due to simultaneous file writes
```

---

pmrv:
Fixes the race condition in issue #214 where two threads in the same process could both proceed through file_write_lock simultaneously, corrupting cached data.

- Add PerKeyLockMixin to FileStorage so all file-backed storage classes are thread-safe out of the box
- Add _PicklableWeakValueDictionary so PerKeyLockMixin survives pickle round-trips
- Update storage_to_config to exclude init=False mixin fields from serialised config
- Add direct test for ValuePickleFile concurrent saves

Closes #214

### DIFF
--- a/src/fleche/storage/file.py
+++ b/src/fleche/storage/file.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import Iterable, Any, Generator
 
 from .base import StorageBackend
+from .thread_safe import PerKeyLockMixin
 from ..digest import Digest
 
 logger = logging.getLogger("fleche.storage")
@@ -70,7 +71,7 @@ def file_read_lock(
 
 
 @dataclass(frozen=True)
-class FileStorage(StorageBackend):
+class FileStorage(PerKeyLockMixin, StorageBackend):
     """File-based storage backend using pickle.
 
     Stores objects on the filesystem.
--- a/src/fleche/storage/file.py
+++ b/src/fleche/storage/file.py
@@ -9,7 +9,6 @@ from pathlib import Path
 from typing import Iterable, Any, Generator
 
 from .base import StorageBackend
-from .thread_safe import PerKeyLockMixin
 from ..digest import Digest
 
 logger = logging.getLogger("fleche.storage")
@@ -71,7 +70,7 @@ def file_read_lock(
 
 
 @dataclass(frozen=True)
-class FileStorage(PerKeyLockMixin, StorageBackend):
+class FileStorage(StorageBackend):
     """File-based storage backend using pickle.
 
     Stores objects on the filesystem.
--- a/src/fleche/storage/thread_safe.py
+++ b/src/fleche/storage/thread_safe.py
@@ -77,6 +77,27 @@ class _PicklableRLock:
         return self._lock.__exit__(*args)
 
 
+class _PicklableWeakValueDictionary:
+    """A ``weakref.WeakValueDictionary`` wrapper that survives pickle round-trips.
+
+    Re-initialised empty on unpickle — its contents are not preserved.  This is
+    intentionally an in-process aid: each process gets its own independent lock
+    table with no shared state across processes.
+    """
+
+    def __init__(self):
+        self._dict: weakref.WeakValueDictionary = weakref.WeakValueDictionary()
+
+    def __reduce__(self):
+        return (type(self), ())
+
+    def get(self, key, default=None):
+        return self._dict.get(key, default)
+
+    def __setitem__(self, key, value):
+        self._dict[key] = value
+
+
 @dataclass(frozen=True)
 class SerializingMixin(KeyManagement):
     """Mixin that serializes all storage operations behind a single reentrant lock.
@@ -114,8 +135,8 @@ class PerKeyLockMixin(KeyManagement):
         class PerKeyValueMemory(PerKeyLockMixin, ValueMemory): ...
     """
 
-    _key_locks: weakref.WeakValueDictionary[Digest | str, threading.RLock] = field(
-        default_factory=weakref.WeakValueDictionary, init=False, repr=False, compare=False
+    _key_locks: _PicklableWeakValueDictionary = field(
+        default_factory=_PicklableWeakValueDictionary, init=False, repr=False, compare=False
     )
     _meta_lock: _PicklableLock = field(
         default_factory=_PicklableLock, init=False, repr=False, compare=False
--- a/src/fleche/storage/thread_safe.py
+++ b/src/fleche/storage/thread_safe.py
@@ -77,27 +77,6 @@ class _PicklableRLock:
         return self._lock.__exit__(*args)
 
 
-class _PicklableWeakValueDictionary:
-    """A ``weakref.WeakValueDictionary`` wrapper that survives pickle round-trips.
-
-    Re-initialised empty on unpickle — its contents are not preserved.  This is
-    intentionally an in-process aid: each process gets its own independent lock
-    table with no shared state across processes.
-    """
-
-    def __init__(self):
-        self._dict: weakref.WeakValueDictionary = weakref.WeakValueDictionary()
-
-    def __reduce__(self):
-        return (type(self), ())
-
-    def get(self, key, default=None):
-        return self._dict.get(key, default)
-
-    def __setitem__(self, key, value):
-        self._dict[key] = value
-
-
 @dataclass(frozen=True)
 class SerializingMixin(KeyManagement):
     """Mixin that serializes all storage operations behind a single reentrant lock.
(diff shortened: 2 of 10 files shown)


src/fleche/storage/thread_safe.py
```diff
@@ -114,8 +135,8 @@ class PerKeyLockMixin(KeyManagement):
         class PerKeyValueMemory(PerKeyLockMixin, ValueMemory): ...
     """
 
-    _key_locks: weakref.WeakValueDictionary[Digest | str, threading.RLock] = field(
-        default_factory=weakref.WeakValueDictionary, init=False, repr=False, compare=False
+    _key_locks: _PicklableWeakValueDictionary = field(
+        default_factory=_PicklableWeakValueDictionary, init=False, repr=False, compare=False
```
pmrv: @claude I think this is unnecessary, see our discussion in https://github.com/pmrv/fleche/pull/318

pmrv: **Claude finished @pmrv's task in 2m 21s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24603374033)

---
### Done

- [x] Read `src/fleche/storage/thread_safe.py` around line 139
- [x] Review PR #318 discussion context
- [x] Remove `_PicklableWeakValueDictionary` and revert `_key_locks` to plain `weakref.WeakValueDictionary`
- [x] Push changes

Removed `_PicklableWeakValueDictionary` and reverted `_key_locks` back to `weakref.WeakValueDictionary[Digest | str, threading.RLock]`. As established in #318: concrete classes that need pickle support define `__reduce__` to reconstruct via `__init__`, which re-creates all `init=False` fields fresh — so the lock table doesn't need its own pickling wrapper.

src/fleche/storage/file.py
```diff
@@ -70,7 +71,7 @@ def file_read_lock(
 
 
 @dataclass(frozen=True)
-class FileStorage(StorageBackend):
+class FileStorage(PerKeyLockMixin, StorageBackend):
```
pmrv: @claude  I'd rather see it mixed in in the final value/call storage class

pmrv: **Claude finished @pmrv's task in 2m 25s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24603555886)

---
### Done

- [x] Read current file.py and thread_safe.py to understand structure
- [x] Remove PerKeyLockMixin from FileStorage
- [x] Add PerKeyLockMixin to ValuePickleFile and CallPickleFile
- [x] Add PerKeyLockMixin to ValueBagOfHoldingH5File and CallBagOfHoldingH5File
- [x] Update test comments/docstrings
- [x] Commit and push

`PerKeyLockMixin` is now applied directly to the four concrete classes (`ValuePickleFile`, `CallPickleFile`, `ValueBagOfHoldingH5File`, `CallBagOfHoldingH5File`) instead of at the `FileStorage` base. `FileStorage` is back to a plain `StorageBackend` subclass with no threading concern; intermediate backends (`PickleFileBackend`, `BagOfHoldingH5FileBackend`) are unaffected.

pmrv:
> FAILED tests/unit/test_pickle.py::test_cache_stack_functional_roundtrip[cloudpickle-memory] - AttributeError: Can't pickle local object 'WeakValueDictionary.__init__.<locals>.remove'

@claude Seems there's actually a problem with the weakvalue dict pickling, or rather the values inside?

pmrv:
**Claude finished @pmrv's task in 5m 38s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24607545359)

---
### Done

- [x] Understand root cause
- [x] Add `__getstate__`/`__setstate__` to `PerKeyLockMixin`
- [x] Run failing test to verify fix — all 251 tests pass
- [x] Commit and push

**Root cause:** `weakref.WeakValueDictionary.__init__` stores `self._remove = remove` where `remove` is a local closure. Even an empty `WeakValueDictionary` carries this unpicklable closure, so any instance of `PerKeyLockMixin` failed to pickle.

**Fix** ([`34a2953`](https://github.com/pmrv/fleche/commit/34a2953)): Added `__getstate__`/`__setstate__` to `PerKeyLockMixin` in `thread_safe.py`. `__getstate__` drops `_key_locks` from the pickled state; `__setstate__` restores everything else and creates a fresh `WeakValueDictionary`. This is safe because the lock table is ephemeral — it only guards in-flight operations, and after unpickling there are none.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-214-20260418-1050)

pmrv:
This won't work as already discussed in https://github.com/pmrv/fleche/pull/318 We cannot have overriden state setter/getters without making all downstream classes deal with this explicitly.  try making the mixin not a dataclass, but rather that it just manages its own attribute and then we hope it won't be picked up by dataclasses default state.

pmrv:
@claude check my last comment

pmrv:
**Claude finished @pmrv's task in 10m 42s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24607954607)

---
### Done

- [x] Understand pmrv's last comment
- [x] Read current `thread_safe.py`
- [x] Implement: make `PerKeyLockMixin` a plain (non-dataclass) class with external lock table storage
- [x] Run all 251 tests — all pass
- [x] Commit and push → [`4f8ebfb`](https://github.com/pmrv/fleche/commit/4f8ebfb)

**Approach:** `PerKeyLockMixin` is now a plain class. Lock tables are stored in a module-level `dict[int, tuple]` keyed by `id(instance)`, with `weakref.finalize` to evict entries when instances are GC'd. Nothing is stored in `self.__dict__`, so pickle works for all downstream classes with zero changes needed there.

One subtlety: `WeakKeyDictionary` (keyed by the instance itself) didn't work because `ValuePickleFile` is unhashable — its `secret_key: list[bytes]` field causes `hash(instance)` to fail for frozen dataclasses. Using `id()` sidesteps this entirely.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-214-20260418-1050)

pmrv:
@claude Try weakref dict again. to make the PickleFile classes work, make the backend store the keys in a tuple.  It is not to be modified at runtime anyway.

pmrv:
**Claude finished @pmrv's task in 12m 7s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24609405272)

---
### Done

- [x] Read relevant files
- [x] Change `secret_key: list[bytes]` → `tuple[bytes, ...]` in `PickleFileBackend`
- [x] Revert `PerKeyLockMixin` to `WeakKeyDictionary`-based outer table
- [x] Update test classes to be hashable
- [x] Run tests — all 767 pass
- [x] Commit and push → [`b1f0f86`](https://github.com/pmrv/fleche/commit/b1f0f86)

**What changed:**

`PickleFileBackend.secret_key` is now `tuple[bytes, ...]` (converted from whatever `normalize_secret_key` returns in `__post_init__`). This makes frozen-dataclass instances like `ValuePickleFile` and `CallPickleFile` fully hashable.

`PerKeyLockMixin` now uses a module-level `weakref.WeakKeyDictionary` (instance → `(WeakValueDict, Lock)`) instead of the `dict[int, tuple]` + `weakref.finalize` approach. The `WeakKeyDictionary` handles eviction automatically when instances are GC'd — no explicit finalize callback needed.

Test classes `PerKeyValueMemory`/`PerKeyCallMemory` add `__hash__ = object.__hash__` because their `storage: dict` field would otherwise make them unhashable (identity hash is fine here since these are test-local throwaway instances).

Three test assertions updated to expect `tuple` instead of `list` for `secret_key`.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-214-20260418-1050)

pmrv:
@claude fix types

pmrv:
**Claude finished @pmrv's task in 3s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24643899187)

---
### Done

- [x] Read changed files to identify type issues
- [x] Fix type annotations
- [x] Run tests — all 767 pass
- [x] Commit and push → [`446d374`](https://github.com/pmrv/fleche/commit/446d374)

**What changed:**

`security.py`:
- `normalize_secret_key` now accepts `Sequence[bytes | str]` (not just `list`) — updates both the type annotation and the runtime `isinstance` check so passing `tuple[bytes, ...]` no longer raises `TypeError`
- `SignedBytes.keys` changed from `list[bytes]` to `Sequence[bytes]`, matching what `PickleFileBackend` passes

`thread_safe.py`:
- `_per_instance_locks` now has full type parameters: `WeakKeyDictionary[PerKeyLockMixin, tuple[WeakValueDictionary[Digest | str, threading.RLock], _PicklableLock]]`

`test_secure_storage.py`:
- Updated error-message match from `"or list"` → `"or sequence"`
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-214-20260418-1050)

PR #351, opened 2026-04-18
Fix SQL code_digest round-trip causing KeyError in cache.load(table().index)

### LINKED ISSUE #319: query().table() index doesn't match actual call storage keys

pmrv:
When loading a Call from SQL, code_digest was returned as a plain str, but at save time it was a Digest. digest(Digest(x)) == x (identity) while digest(str(x)) == SHA256("str"+x), so to_lookup_key() produced a different hash after SQL round-trip, breaking cache.load(table().index.iloc[0]).

Fix: wrap code_digest in Digest() in Sql.get(), matching the existing treatment of result and argument values.

Closes #319.

### DIFF
--- a/src/fleche/storage/sql.py
+++ b/src/fleche/storage/sql.py
@@ -236,7 +236,7 @@ class Sql(PerKeyLockMixin, CallStorage):
             metadata={row.name: (row.data or {}) for row in meta_rows},
             module=call_model.module,
             version=call_model.version,
-            code_digest=call_model.code_digest,
+            code_digest=Digest(call_model.code_digest) if call_model.code_digest is not None else None,
             result=(
                 Digest(call_model.result) if call_model.result is not None else None
             ),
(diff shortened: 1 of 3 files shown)


pmrv:
@claude test_sql_call_digest_persistence break, because your change changes call digests, update the old test.

pmrv:
**Claude finished @pmrv's task in 1m 39s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24607209717)

---
### Updated `test_sql_call_digest_persistence`

- [x] Find `test_sql_call_digest_persistence` — in `tests/unit/storage/test_sql_query.py:256`
- [x] Update test: changed `code_digest="some_code_digest"` → `code_digest=Digest("some_code_digest")`
- [x] Test passes locally
- [x] Committed and pushed

**Why this was needed:** The SQL fix makes `code_digest` come back as `Digest` on load. The original `Call` in the test used a plain `str`, so `digest(str(x))` ≠ `digest(Digest(x))`, causing the round-trip lookup key assertion to fail. Using `Digest(...)` from the start aligns the types.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-319-20260418-1429)

PR #357, opened 2026-04-19
Add convenience methods to QueryIterator

### LINKED ISSUE #354: QueryIterator needs some more concenience
- [ ] .first()
- [ ] .last() ?? 
- [ ] .only() -> first() but raises error if there are more matches

pmrv:
Implements `.first()`, `.last()`, `.only()`, `.count()`, `.any()`, `.is_empty()`, `.take(n)`, `.skip(n)`, `.filter(predicate)`, `.sorted(key, reverse)`, `.unique_by(key)`, `.groupby(key)`, `.latest()`, `.oldest()`, `.delete_all()` on `QueryIterator`.

Closes #354

### DIFF
--- a/src/fleche/query.py
+++ b/src/fleche/query.py
@@ -1,6 +1,8 @@
+import builtins
 import datetime
+import itertools
 from dataclasses import dataclass
-from typing import Iterable, Iterator, Any, Literal
+from typing import Iterable, Iterator, Any, Literal, Callable
 
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


src/fleche/query.py
```diff
@@ -19,7 +21,156 @@ class QueryIterator(Iterable[call.LazyCall]):
...
+        """Return the total number of matching calls."""
+        return builtins.sum(1 for _ in self)
+
+    def any(self) -> bool:
+        """Return True if at least one matching call exists (short-circuits)."""
+        for _ in self:
+            return True
+        return False
+
+    def is_empty(self) -> bool:
```
pmrv: empty

src/fleche/query.py
```diff
@@ -19,7 +21,156 @@ class QueryIterator(Iterable[call.LazyCall]):
...
+        Args:
+            key: a callable taking a LazyCall, or a string argument name to sort by
+            reverse: if True, sort in descending order
+        """
+        if isinstance(key, str):
+            arg_name = key
+            key = lambda c: c.arguments[arg_name]
+        return QueryIterator(builtins.sorted(self, key=key, reverse=reverse))
+
+    def unique_by(self, key: "str | Callable[[call.LazyCall], Any]") -> "QueryIterator":
```
pmrv: unique

src/fleche/query.py
```diff
@@ -19,7 +21,156 @@ class QueryIterator(Iterable[call.LazyCall]):
...
+
+        Raises:
+            IndexError: if there are no matching calls
+        """
+        calls = builtins.list(self)
+        if not calls:
+            raise IndexError("QueryIterator is empty")
+        return builtins.min(calls, key=lambda c: c.metadata.get("runtime", {}).get("timestart", float("inf")))
+
+    def delete_all(self) -> None:
```
pmrv: evict

pmrv (commented):
@claude Rename according to comments.

Critique the new tests.

pmrv:
> @claude Rename according to comments.
> 
> Critique the new tests.

@claude do this

pmrv:
**Claude finished @pmrv's task in 2m 7s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24644305361)

---
### Tasks

- [x] Rename methods: `is_empty` → `empty`, `unique_by` → `unique`, `delete_all` → `evict`
- [x] Update tests to match renamed methods
- [x] Commit and push
- [x] Critique the new tests

---

### Test Critique

The new tests in `tests/unit/fleche/test_query_iterator.py` are well-structured and broadly good. Here are the issues and suggestions:

#### Issues

**1. `test_unique_by_callable` uses identical `Call` objects — but duplicate keys are likely rejected by the cache** (`test_query_iterator.py:514–521`)

Two calls with `{"x": 1}` are saved to the same cache. Since the cache is keyed by content digest, the second save likely overwrites/deduplicates to a single entry. This means the test may pass vacuously (2 items from 3 saves becomes 2 from 2). The test should use `{"x": 1, "y": 1}` / `{"x": 1, "y": 2}` like `test_unique_by_argument_name` does, or verify the actual starting count.

**2. `test_first_returns_first_call` / `test_last_returns_last_call` assert ordering but cache order is unspecified** (`test_query_iterator.py:320–340`)

The tests create two calls, collect them into `calls = list(...)`, then feed the same list back to `QueryIterator(iter(calls))` and assert `qi.first() == calls[0]`. This is tautologically true regardless of the implementation — it just tests that `first()` returns the first element of whatever iteration order the cache happens to use. The tests don't verify anything about stable ordering. Either: (a) acknowledge they only test "some element is returned" and drop the exact equality, or (b) add `.sorted()` to make order deterministic before asserting.

**3. `test_skip_more_than_available` doesn't test skipping past a non-empty iterator** (`test_query_iterator.py:435–436`)

It skips 100 on an empty `QueryIterator([])`. This doesn't verify that `skip` on a populated iterator with `n > len` returns empty. Add a case that skips past a real set of items.

**4. `test_filter_all_excluded` tests an empty input, not filtering-to-empty** (`test_query_iterator.py:460–461`)

`QueryIterator([]).filter(lambda c: False)` trivially returns `[]` because the input is already empty. The test should instead start with some items and apply a predicate that rejects all of them. [Fix this →](https://claude.ai/code?q=Replace%20test_filter_all_excluded%20in%20tests%2Funit%2Ffleche%2Ftest_query_iterator.py%20so%20it%20starts%20with%20one%20or%20more%20real%20Call%20objects%20in%20the%20cache%20and%20applies%20a%20predicate%20that%20rejects%20all%20of%20them%2C%20rather%20than%20testing%20an%20already-empty%20QueryIterator&repo=pmrv/fleche)

**5. `test_take_zero` tests an empty input, not `take(0)` on real data** (`test_query_iterator.py:422–423`)

Same pattern as above — `QueryIterator([]).take(0)` doesn't exercise the slicing logic at all. Should start with items and assert `take(0)` produces nothing.

#### Minor nits

- `test_unique_by_argument_name` and `test_unique_by_callable` kept the old `_by` suffix in their function names after the method rename. These should be `test_unique_argument_name` and `test_unique_callable` for consistency.
- `latest`/`oldest` tests only verify the result via `c.arguments["x"]`. It's slightly cleaner to check `c.metadata["runtime"]["timestart"]` directly, since that's the field being compared — the argument value is an indirect proxy.
- The `_make_cache_with_calls` helper is defined but never used in the new tests (all tests build the cache manually inline). Either use it or remove it.

#### What's good

- Good coverage of both string-key and callable variants for `sorted`, `unique`, `groupby`.
- `test_evict_only_removes_matched` is a solid isolation test.
- Section headers make the file easy to navigate.
- The `latest`/`oldest` empty-raises tests are clean and correct.
 |

pmrv:
> Two calls with {"x": 1} are saved to the same cache. Since the cache is keyed by content digest, the second save likely overwrites/deduplicates to a single entry. This means the test may pass vacuously (2 items from 3 saves becomes 2 from 2). The test should use {"x": 1, "y": 1} / {"x": 1, "y": 2} like test_unique_by_argument_name does, or verify the actual starting count.

The test tries to unique on the arguments, so your suggestion doesn't quite work.   Change the result field in the second call to ensure both calls are saved.

pmrv:
> The tests create two calls, collect them into calls = list(...), then feed the same list back to QueryIterator(iter(calls)) and assert qi.first() == calls[0]. This is tautologically true regardless of the implementation — it just tests that first() returns the first element of whatever iteration order the cache happens to use. The tests don't verify anything about stable ordering. Either: (a) acknowledge they only test "some element is returned" and drop the exact equality, or (b) add .sorted() to make order deterministic before asserting.

No, this is ok.  query itself does not guarantee any sorting, so first() should only be tested against the underlying iterable as done in the test.  However, add a note to first/last's docstring, that users should sort the iteration iff they should require stable order.

pmrv:
> > The tests create two calls, collect them into calls = list(...), then feed the same list back to QueryIterator(iter(calls)) and assert qi.first() == calls[0]. This is tautologically true regardless of the implementation — it just tests that first() returns the first element of whatever iteration order the cache happens to use. The tests don't verify anything about stable ordering. Either: (a) acknowledge they only test "some element is returned" and drop the exact equality, or (b) add .sorted() to make order deterministic before asserting.
> 
> No, this is ok. query itself does not guarantee any sorting, so first() should only be tested against the underlying iterable as done in the test. However, add a note to first/last's docstring, that users should sort the iteration iff they should require stable order.

No actually, I like the thought you had.  Let's do this.  Drop first last.  Reconceptualize any(). It will now return the first element of the iterator found and the docstring should hint that users may `sorted(reverse=...)` as necessary to obtain what they want.

pmrv:
> test_skip_more_than_available, test_filter_all_excluded, test_take_zero

Correct assessment.

pmrv:
> latest/oldest tests only verify the result via c.arguments["x"]. It's slightly cleaner to check c.metadata["runtime"]["timestart"] directly, since that's the field being compared — the argument value is an indirect proxy.

Rewrite the test to actually implement the invariant outside, i.e. for a given list of calls it calculate argmin/max by runtime itself and verifies that against the implementation.

pmrv:
@claude See my comment above and fix them, your nits are also correct.

pmrv:
**Claude finished @pmrv's task in 7m 30s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24644787087)

---
### Tasks

- [x] Gather context (read source files)
- [x] Drop `first()` and `last()` methods; reconceptualize `any()` to return first element with sorting hint
- [x] Fix `test_unique_callable`: use `version=1`/`version=2` to create distinct cache entries for two `x=1` calls (changing `result` alone doesn't affect the lookup key, which excludes results)
- [x] Fix `test_skip_more_than_available`: populate iterator then skip past all items
- [x] Fix `test_filter_all_excluded`: start with items, filter all out with `lambda c: False`
- [x] Fix `test_take_zero`: start with items, `take(0)` returns nothing
- [x] Rewrite `latest`/`oldest` tests to independently compute argmax/argmin via `max()`/`min()` and verify `to_lookup_key()` matches
- [x] Rename `test_unique_by_argument_name` → `test_unique_argument_name`, `test_unique_by_callable` → `test_unique_callable`
- [x] Remove unused `_make_cache_with_calls` helper
- [x] Run tests and verify — 811 passed, 3 skipped
- [x] Commit and push

### Notes

- `any()` now returns `call.LazyCall | None` (first element or `None` on empty), with a docstring hint that callers should `.sorted(reverse=...)` to control which element is returned. `empty()` was updated to not rely on `any()` since its return type changed.
- The `test_unique_callable` fix uses `version=1`/`version=2` to differentiate two otherwise-identical calls: the result field is excluded from `to_lookup_key()` so changing it doesn't produce distinct cache entries.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-354-20260419-1531)

pmrv (commented):
@claude rebase on main and fix merge conflicts.

pmrv:
> @claude rebase on main and fix merge conflicts.

@claude ping

pmrv:
**Claude finished @pmrv's task in 8m 18s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24667112971)

---
### Tasks

- [x] Gather context
- [x] Fetch latest main and unshallow repository
- [x] Rebase on main and fix conflicts
- [x] Run tests — 838 passed, 3 skipped
- [x] Push changes

Rebased 3 commits on top of `origin/main` (now at `6760187`). The only conflict was in `src/fleche/query.py` — merged the `Literal` import (from main) with the `Callable` import (from branch) and preserved the updated `table()` signature with `Literal[True]` support.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-354-20260419-1531)

PR #360, opened 2026-04-20
test: failing regression test for #352 argument/metadata key clash

### LINKED ISSUE #352: Bug: table(arguments=...) drops arguments silently that clash with metadata tags

pmrv:
Adds two failing regression tests for issue #352: `table(arguments=...)` silently drops arguments whose names match flattened metadata keys.

Fixes #352

### DIFF
--- a/src/fleche/query.py
+++ b/src/fleche/query.py
@@ -62,6 +62,12 @@ class QueryIterator(Iterable[call.LazyCall]):
             }
             if results:
                 row["result"] = c.result
+            md = row.pop("metadata", {}) or {}
+            # Flatten each metadata name's dict into the row first, so argument
+            # clash detection below also catches metadata-produced keys.
+            for data in md.values():
+                if isinstance(data, dict):
+                    row.update(data)
             for a in (c.arguments.keys() if arguments is True else arguments):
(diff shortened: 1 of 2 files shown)


pmrv:
@claude solve it.

pmrv:
**Claude finished @pmrv's task in 4m 4s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24644909135)

---
### Solving issue #352: argument/metadata key clash

**Root cause:** In `query.py:72-76`, metadata was flattened into the row *after* arguments were added. The `row.update(data)` call silently overwrote any argument whose name matched a metadata-produced key. The existing clash detection (which prefixes collisions with `a_`) only guarded against `name`/`module`/`metadata`/`result` keys.

**Fix:** Move metadata flattening to happen *before* arguments are added, so the existing `a_<name>` prefix logic also catches metadata key collisions ([`src/fleche/query.py`](https://github.com/pmrv/fleche/blob/claude/issue-352-20260419-2245/src/fleche/query.py)).

- [x] Gather context
- [x] Understand the bug from the regression test
- [x] Locate the relevant source code (`src/fleche/query.py:65-76`)
- [x] Implement the fix
- [x] Run tests to verify (789 passed, 3 skipped)
- [x] Commit and push

Both regression tests now pass and the full test suite is green.
 |

PR #366, opened 2026-04-20
refactor: extract _resolve_prefix to deduplicate expand() ambiguity logic

### LINKED ISSUE #324: Refactor: dedupe `expand()` ambiguity logic between base.py and sql.py
## Location

- `src/fleche/storage/base.py:86-109` — `KeyManagement.expand`
- `src/fleche/storage/sql.py:254-287` — `Sql.expand`

## Problem

`Sql.expand` reimplements the full expand/ambiguity algorithm from `KeyManagement.expand`. The two bodies are structurally identical — length guards, "no match → `KeyError`", "two matches → find divergence point and raise `AmbiguousDigestError` with the same message" — the only real difference is how candidates are fetched (in-memory `list()` + `startswith` vs. `LIKE 'prefix%' LIMIT 2`).

If the ambiguity error message or the divergence-index computation ever changes (e.g. off-by-one on `i+1`, localization, richer error payload), it has to be updated in two places and drift is easy to miss.

## Suggested refactor

Extract a private helper in `base.py` that, given up to two candidate keys and the short prefix, returns the unambiguous `Digest` or raises `AmbiguousDigestError` / `KeyError`. Both `KeyManagement.expand` and `Sql.expand` become thin wrappers that just *fetch candidates* and delegate to the helper.

```python
def _resolve_prefix(key: str, candidates: list[str]) -> Digest: ...
```

## Pros

- Single source of truth for the ambiguity error message + divergence computation.
- Keeps the SQL-side `LIKE ... LIMIT 2` optimisation intact.
- Very small diff, pure extraction.

## Cons / risks

- Minimal. One extra level of indirection in each `expand`. No API change, no new dependency.

pmrv:
Moves the shared length-guard / KeyError / AmbiguousDigestError logic into a single module-level helper `_resolve_prefix(key, candidates)` in base.py.

KeyManagement.expand and Sql.expand both become thin wrappers that only fetch up to two candidate keys and delegate to the helper, keeping the SQL LIKE+LIMIT 2 optimisation intact.

Closes #324

### DIFF
--- a/src/fleche/storage/base.py
+++ b/src/fleche/storage/base.py
@@ -18,6 +18,28 @@ class AmbiguousDigestError(ValueError):
     pass
 
 
+def _resolve_prefix(key: str, candidates: list[str]) -> Digest:
+    """Return the unique Digest for *key* prefix, or raise KeyError / AmbiguousDigestError.
+
+    *candidates* must contain at most two entries (the two lexicographically
+    smallest keys that start with *key*); callers are responsible for fetching
+    them efficiently (e.g. via a ``LIKE … LIMIT 2`` query for SQL backends).
+    """
+    if not candidates:
+        raise KeyError(key)
+    if len(candidates) == 1:
+        return Digest(candidates[0])
+    m1, m2 = candidates[0], candidates[1]
+    for i, (c1, c2) in enumerate(zip(m1, m2)):
+        if c1 != c2:
+            break
+    else:
+        i = min(len(m1), len(m2))
+    raise AmbiguousDigestError(
+        f"Short digest {key} is ambiguous; need at least {i+1} characters."
+    )
+
+
 class KeyManagement(ABC):
     """Abstract base providing key-management helpers for any keyed storage.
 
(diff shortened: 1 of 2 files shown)


pmrv:
@claude Fix types, you're getting confused between str/Digest.

pmrv:
**Claude finished @pmrv's task in 22s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24667895773)

---
### Fixed str/Digest type confusion

- [x] Read changed files to identify type issues
- [x] Fix type annotations
- [x] Run tests (778 passed, 3 skipped)
- [x] Commit and push

**Changes made:**

`base.py` — `_resolve_prefix`:
- `candidates: list[str]` → `candidates: list[Digest]` (since `self.list()` returns `Iterable[Digest]`, candidates are already `Digest` objects)
- `return Digest(candidates[0])` → `return candidates[0]` (already a `Digest`, no re-wrapping needed)

`sql.py` — `Sql.expand`:
- `[r[0] for r in rows]` → `[Digest(r[0]) for r in rows]` (SQL returns raw strings; wrap as `Digest` before passing to `_resolve_prefix`)
 |

PR #375, opened 2026-04-21
feat(state): proxy .fleche namespace on BoundWrapper

### LINKED ISSUE #374: BoundWrapper should proxy .fleche namespace
## Problem

`BoundWrapper` does not forward attribute access to the wrapped function, so `bound.fleche.digest(...)` raises `AttributeError`. Users must instead write `bound.func.fleche.digest(...)`, which is not discoverable.

Noted during analysis in #222.

## Solution

Add a `fleche` property to `BoundWrapper` that returns `self.func.fleche`, making the helper namespace accessible as `bound.fleche.*`.

pmrv:
Add a fleche property to BoundWrapper that forwards to self.func.fleche, so bound.fleche.digest(...) etc. work without needing bound.func.fleche.

Closes #374

### DIFF
--- a/src/fleche/state.py
+++ b/src/fleche/state.py
@@ -1,7 +1,7 @@
 from contextlib import AbstractContextManager
 import contextvars
 from contextvars import ContextVar, Token
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from types import SimpleNamespace
 from typing import overload, Callable
 
@@ -127,6 +127,16 @@ def project(name):
     return tags(project=name)
 
 
+def _reconstruct_bound_wrapper(func, bound_cache, bound_meta):
+    """Module-level factory used by BoundWrapper.__reduce__ (contextvars.Context is not picklable)."""
+    def _setup():
+        _CACHE.set(bound_cache)
+        _METADATA.set(bound_meta)
+    ctx = contextvars.copy_context()
+    ctx.run(_setup)
+    return BoundWrapper(func, bound_cache, bound_meta, ctx)
+
+
 @dataclass(frozen=True, eq=True)
 class BoundWrapper:
     """Utility class that freezes global state for the cache and metadata config.
@@ -139,6 +149,7 @@ class BoundWrapper:
     func: Callable
     cache: caches.BaseCache
     meta: tuple[metadata.MetaData, ...]
+    ctx: contextvars.Context = field(compare=False, repr=False)
 
     @classmethod
     def bind(cls, func):
@@ -152,20 +163,23 @@ class BoundWrapper:
 
         Returns:
             :class:`.BoundWrapper`: instance with the bound cache and metadata state"""
-        return cls(func, _CACHE.get(), _METADATA.get())
+        return cls(func, _CACHE.get(), _METADATA.get(), contextvars.copy_context())
 
     @property
     def fleche(self):
         """Return a .fleche namespace whose helpers run in the bound cache/meta context."""
+        if not hasattr(self.func, 'fleche'):
+            raise AttributeError(
+                f"BoundWrapper.fleche requires a fleche-decorated function; "
+                f"{self.func!r} has no .fleche namespace"
+            )
         ns = self.func.fleche  # ty: ignore[unresolved-attribute]
-        return SimpleNamespace(**{
-            name: BoundWrapper(helper, self.cache, self.meta)
-            for name, helper in vars(ns).items()
-        })
+        def _bind_helpers():
+            return {name: BoundWrapper.bind(helper) for name, helper in vars(ns).items()}
+        return SimpleNamespace(**self.ctx.run(_bind_helpers))
 
     def __call__(self, *args, **kwargs):
-        def _call():
-            _CACHE.set(self.cache)
-            _METADATA.set(self.meta)
-            return self.func(*args, **kwargs)
-        return contextvars.copy_context().run(_call)
+        return self.ctx.run(self.func, *args, **kwargs)
+
+    def __reduce__(self):
+        return (_reconstruct_bound_wrapper, (self.func, self.cache, self.meta))
--- a/src/fleche/state.py
+++ b/src/fleche/state.py
@@ -125,6 +125,33 @@ def project(name):
     return tags(project=name)
 
 
+class _BoundFlecheNamespace:
+    """Wraps a .fleche SimpleNamespace so every helper activates the bound cache/meta state."""
+
+    __slots__ = ('_ns', '_cache', '_meta')
+
+    def __init__(self, ns, bound_cache, bound_meta):
+        self._ns = ns
+        self._cache = bound_cache
+        self._meta = bound_meta
+
+    def __getattr__(self, name):
+        helper = getattr(self._ns, name)
+        bound_cache = self._cache
+        bound_meta = self._meta
+
+        def _bound(*args, **kwargs):
+            token_cache = _CACHE.set(bound_cache)
+            token_meta = _METADATA.set(bound_meta)
+            try:
+                return helper(*args, **kwargs)
+            finally:
+                _METADATA.reset(token_meta)
+                _CACHE.reset(token_cache)
+
+        return _bound
+
+
 @dataclass(frozen=True, eq=True)
 class BoundWrapper:
     """Utility class that freezes global state for the cache and metadata config.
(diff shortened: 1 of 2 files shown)


pmrv:
@claude I think this won't actually work the way I want it.  Calling the proxied helper does not actually activate the bound cache configuration.  You may haves more luck bind()'ing the helpers as well, but may that's quite verbose.  Investigate if you can pull some fancy descriptor magic.

pmrv:
**Claude finished @pmrv's task in 6m 8s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24726798345)

---
### Fix: `BoundWrapper.fleche` now activates bound cache/meta on every helper call

- [x] Read current implementation (`state.py`, `wrapper.py`)
- [x] Understand the problem
- [x] Investigate descriptor magic approach
- [x] Implement fix
- [x] Run tests (840 passed, 0 failures)
- [x] Push changes

---

**Root cause:** Helpers like `make_contains`, `make_load`, `make_query`, `make_rerun` all call `state._CACHE.get()` at invocation time. Proxying the `.fleche` `SimpleNamespace` just gave direct access to those same closures — they still resolved the cache from the current context, not from the `BoundWrapper`.

**Fix: `_BoundFlecheNamespace`** (added to `state.py`)

```python
class _BoundFlecheNamespace:
    __slots__ = ('_ns', '_cache', '_meta')

    def __init__(self, ns, bound_cache, bound_meta):
        self._ns = ns
        self._cache = bound_cache
        self._meta = bound_meta

    def __getattr__(self, name):
        helper = getattr(self._ns, name)
        bound_cache = self._cache
        bound_meta = self._meta

        def _bound(*args, **kwargs):
            token_cache = _CACHE.set(bound_cache)
            token_meta = _METADATA.set(bound_meta)
            try:
                return helper(*args, **kwargs)
            finally:
                _METADATA.reset(token_meta)
                _CACHE.reset(token_cache)

        return _bound
```

`BoundWrapper.fleche` now returns `_BoundFlecheNamespace(self.func.fleche, self.cache, self.meta)`. Every attribute access on it intercepts via `__getattr__`, wraps the underlying helper in a closure that temporarily sets `_CACHE`/`_METADATA` to the bound values, then delegates — exactly mirroring what `BoundWrapper.__call__` already does.

The old identity test (`assert bound.fleche is my_func.fleche`) was replaced with a correctness test: `bound.fleche.contains(5)` finds the result in the bound cache even when the active context is a different (empty) cache.
 |

pmrv:
@claude Let's take this opportunity to simplify the wrapper to just call `contextvars.Context.run`.  There's some awkward deduplication here as well: Really you want `BoundWrapper.bind(func.fleche.helper)`, but you don't want to proxy .fleche on them.

pmrv:
**Claude finished @pmrv's task in 6m 35s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24728077958)

---
### Simplifying `_BoundFlecheNamespace` with `contextvars.Context.run`

- [x] Read current implementation
- [x] Understand approach
- [x] Refactor `BoundWrapper.__call__` to use `contextvars.copy_context().run()`
- [x] Replace `_BoundFlecheNamespace` with `SimpleNamespace` of `BoundWrapper`-wrapped helpers
- [x] Run tests — 840 passed, 0 failures
- [x] Commit and push

---

**What changed** ([`273438f`](https://github.com/pmrv/fleche/commit/273438f))

`_BoundFlecheNamespace` (25 lines) is gone. Two things replace it:

**`BoundWrapper.__call__`** — manual set/reset token dance becomes:
```python
def __call__(self, *args, **kwargs):
    def _call():
        _CACHE.set(self.cache)
        _METADATA.set(self.meta)
        return self.func(*args, **kwargs)
    return contextvars.copy_context().run(_call)
```
`copy_context()` creates an isolated copy — no tokens, no try/finally, and the caller's context is never touched.

**`BoundWrapper.fleche`** — `_BoundFlecheNamespace` is replaced by building a fresh `SimpleNamespace` at access time, wrapping each helper as a `BoundWrapper` directly:
```python
@property
def fleche(self):
    ns = self.func.fleche  # ty: ignore[unresolved-attribute]
    return SimpleNamespace(**{
        name: BoundWrapper(helper, self.cache, self.meta)
        for name, helper in vars(ns).items()
    })
```
Each bound helper runs through the same `copy_context().run()` path in `__call__`. No separate wrapping logic, no separate class.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-222-20260421-1354)

src/fleche/state.py
```diff
@@ -152,11 +154,18 @@ def bind(cls, func):
...
-        token_meta = _METADATA.set(self.meta)
-        try:
+        def _call():
+            _CACHE.set(self.cache)
+            _METADATA.set(self.meta)
             return self.func(*args, **kwargs)
-        finally:
-            _METADATA.reset(token_meta)
-            _CACHE.reset(token_cache)
+        return contextvars.copy_context().run(_call)
```
pmrv: No, you should instead copy the context at bind time, then use its run method

src/fleche/state.py
```diff
@@ -152,11 +154,18 @@ def bind(cls, func):
...
         return cls(func, _CACHE.get(), _METADATA.get())
 
+    @property
+    def fleche(self):
+        """Return a .fleche namespace whose helpers run in the bound cache/meta context."""
+        ns = self.func.fleche  # ty: ignore[unresolved-attribute]
+        return SimpleNamespace(**{
+            name: BoundWrapper(helper, self.cache, self.meta)
+            for name, helper in vars(ns).items()
+        })
```
pmrv: As I mentioned in the last comment, now you have the problem, that e.g. func.fleche.digest has a fleche property.  We don't want that.  Rebuilding the name space is not bad, though, try and raise an attribute error or equivalent in case BoundWrapper.func is not a flechette.

pmrv (commented):
@claude see comments

pmrv:
> @claude see comments

@claude Jo.

pmrv:
**Claude finished @pmrv's task in 9m 53s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24730296279)

---
### Review comments addressed — [`1f26c5d`](https://github.com/pmrv/fleche/commit/1f26c5d)

- [x] Read review comments
- [x] Read current `state.py` implementation
- [x] Fix 1: Copy context at bind time (`self.ctx`); use `self.ctx.run()` in `__call__`
- [x] Fix 2: Raise `AttributeError` explicitly when `self.func` is not a flechette; build helper namespace via `self.ctx.run()`
- [x] Handle pickle: added `__reduce__` + module-level `_reconstruct_bound_wrapper`
- [x] Tests — 842 passed, 3 skipped
- [x] Committed and pushed

---

**What changed**

**`bind()` / `__call__`**

Context is copied once at bind time and stored as `self.ctx`:

```python
@classmethod
def bind(cls, func):
    return cls(func, _CACHE.get(), _METADATA.get(), contextvars.copy_context())

def __call__(self, *args, **kwargs):
    return self.ctx.run(self.func, *args, **kwargs)
```

No more `_call` closure, no manual set/reset tokens on every invocation. `copy_context()` captures `_CACHE`/`_METADATA` at bind time; `self.ctx.run()` re-enters that snapshot for each call.

**`fleche` property**

```python
@property
def fleche(self):
    if not hasattr(self.func, 'fleche'):
        raise AttributeError(
            f"BoundWrapper.fleche requires a fleche-decorated function; "
            f"{self.func!r} has no .fleche namespace"
        )
    ns = self.func.fleche  # ty: ignore[unresolved-attribute]
    def _bind_helpers():
        return {name: BoundWrapper.bind(helper) for name, helper in vars(ns).items()}
    return SimpleNamespace(**self.ctx.run(_bind_helpers))
```

- Explicit `AttributeError` for non-flechettes — `bound.fleche.contains.fleche` now fails with a clear message instead of a cryptic `AttributeError: 'function' object has no attribute 'fleche'`.
- Helpers are bound by running `BoundWrapper.bind()` inside `self.ctx`. Inside that run, `copy_context()` copies `self.ctx` (which has `_CACHE = self.cache`), giving each helper its own isolated copy of the bound context.

**Pickle**

`contextvars.Context` is not picklable, so `ctx` is excluded via `__reduce__`:

```python
def __reduce__(self):
    return (_reconstruct_bound_wrapper, (self.func, self.cache, self.meta))
```

`_reconstruct_bound_wrapper` rebuilds a fresh context with the bound values on load.

Two new tests added: `test_bound_wrapper_fleche_raises_for_plain_function` and `test_bound_wrapper_fleche_helpers_have_no_fleche`.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-222-20260421-1354)

pmrv:
Too complicated and of questionable use, we'll just document it away.

PR #377, opened 2026-04-21
Add .bind() helper method on wrapped functions

pmrv:
$(cat <<'EOF'
## Summary

- Adds a `.bind(*args, **kwargs)` helper to every `@fleche`-decorated function (also accessible via `.fleche.bind`)
- With no arguments, returns a `BoundWrapper` that freezes the current cache and metadata state — equivalent to calling `BoundWrapper.bind(func)` directly but ergonomically available on the wrapper
- With positional/keyword arguments, pre-applies them via `functools.partial` before binding, enabling partial application with frozen state

## Usage

```python
@fleche
def add(a, b):
    return a + b

# Bind current cache/meta state
with cache(my_cache):
    bound = add.bind()          # BoundWrapper(func=add, cache=my_cache, ...)

# Partial application + bind
with cache(my_cache):
    add_one = add.bind(a=1)     # pre-applies a=1

result = add_one(b=9)           # calls add(1, 9), stored in my_cache
```

## Test plan

- [ ] `test_bind_helper_no_args_returns_bound_wrapper` — returns `BoundWrapper` pointing at the wrapper
- [ ] `test_bind_helper_uses_bound_cache` — cache at bind time is used, not the active cache at invocation time
- [ ] `test_bind_helper_partial_args` — positional partial application works
- [ ] `test_bind_helper_partial_kwargs` — keyword partial application works
- [ ] `test_bind_helper_partial_stores_in_bound_cache` — cache isolation holds with partial args
- [ ] `test_bind_helper_accessible_via_fleche_namespace` — `.fleche.bind is .bind`

https://claude.ai/code/session_015BnHsfMLicmVDSi6sBo1KA
EOF
)

### DIFF
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -1,6 +1,6 @@
 import os
 from pathlib import Path
-from functools import wraps
+from functools import wraps, partial
 from inspect import signature
 from types import SimpleNamespace
 from typing import Any, Callable, Dict, Iterable, TypeVar, Annotated, get_type_hints, get_origin, get_args
@@ -249,6 +249,14 @@ def make_rerun(func, wrapper):
     return _rerun_func
 
 
+def make_bind(wrapper):
+    """Build the `.bind` helper that creates a :class:`.BoundWrapper` with optionally pre-applied arguments."""
+    def _bind_func(*args, **kwargs):
+        target = partial(wrapper, *args, **kwargs) if (args or kwargs) else wrapper
+        return state.BoundWrapper.bind(target)
+    return _bind_func
+
+
 def make_wrapper(func, policy, meta, isolate, get_call):
     """Build the cached wrapper returned by :func:`fleche`."""
     @wraps(func)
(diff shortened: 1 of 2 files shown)


pmrv (commented):
@claude already use the prefered syntax func.fleche.bind everywhere

pmrv:
**Claude finished @pmrv's task in 2m 9s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24727099790)

---
### Tasks

- [x] Gather context
- [x] Read wrapper.py and test file to identify usages
- [x] Update all `func.bind(...)` calls to use `func.fleche.bind(...)` syntax
- [x] Commit and push changes

All 5 direct `.bind(...)` calls in the new test section (lines 258, 274, 295, 310, 326) have been updated to use `func.fleche.bind(...)`. The docstrings for those tests were updated accordingly too (e.g. `""".bind() with no args..."` → `""".fleche.bind() with no args..."`). The final test (`test_bind_helper_accessible_via_fleche_namespace`) that asserts `my_func.bind is my_func.fleche.bind` was left as-is since it specifically tests that both forms are the same object.

PR #379, opened 2026-04-21
test(storage): add mixin-level tests; bring base.py/thread_safe.py/destructuring.py to 100%

pmrv:
Add targeted mixin-level tests for ValueMixin, CallMixin, and base class helpers. Brings the three core mixin files from 78-99% to 100% coverage.

Closes #282 (second step)

### DIFF
--- a/tests/unit/storage/test_thread_safe.py
+++ b/tests/unit/storage/test_thread_safe.py
@@ -199,8 +199,9 @@ def test_picklable_lock_pickle_roundtrip():
     restored = pickle.loads(pickle.dumps(lock))
     assert isinstance(restored, _PicklableLock)
     # Restored lock must be acquirable (not stuck in acquired state).
-    with restored:
-        pass
+    # Use blocking=False so a stuck lock raises AssertionError instead of deadlocking.
+    assert restored._lock.acquire(blocking=False), "lock should be free after unpickle"
+    restored._lock.release()
 
 
 def test_picklable_rlock_pickle_roundtrip():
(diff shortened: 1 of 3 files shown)


tests/unit/storage/test_thread_safe.py
```diff
@@ -186,3 +186,28 @@ def test_per_key_lock_released_when_unused():
...
+
+def test_picklable_lock_pickle_roundtrip():
+    """_PicklableLock.__reduce__ produces a fresh unlocked lock on unpickle."""
+    import pickle
+    lock = _PicklableLock()
+    restored = pickle.loads(pickle.dumps(lock))
+    assert isinstance(restored, _PicklableLock)
+    # Restored lock must be acquirable (not stuck in acquired state).
+    with restored:
+        pass
```
pmrv: This will deadlock us on failure

pmrv (commented):
@claude

pmrv:
**Claude finished @pmrv's task in 1m 27s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24729558158)

---
### Fixed deadlock issue in test_thread_safe.py

- [x] Gather context
- [x] Identify the deadlock issue at line 203
- [x] Fix the deadlock
- [x] Commit and push

**Root cause:** `with restored:` at lines 202–203 and 211–213 blocks indefinitely if the lock is stuck in an acquired state — which is exactly the failure condition being tested for. On test failure this would deadlock instead of raising `AssertionError`.

**Fix (commit [29b9098](https://github.com/pmrv/fleche/commit/29b9098)):** Replaced both `with restored: pass` blocks with `acquire(blocking=False)`:

```python
assert restored._lock.acquire(blocking=False), "lock should be free after unpickle"
restored._lock.release()
```

Applied to both `test_picklable_lock_pickle_roundtrip` and `test_picklable_rlock_pickle_roundtrip`. If a lock is ever stuck in an acquired state, the test now fails immediately with `AssertionError` rather than hanging the test suite.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-282-20260420-1525)

PR #380, opened 2026-04-21
Add wrap_executor to automatically bind fleche functions at submission

pmrv:
## Summary
Introduces `fleche.wrap_executor()`, a utility function that monkey-patches an executor's `submit` method to automatically handle fleche-decorated functions. This eliminates the need for users to manually call `BoundWrapper.bind()` and enables cache-hit short-circuiting without executor overhead.

## Key Changes
- **New module `src/fleche/executor.py`**: Implements `wrap_executor()` which:
  - Intercepts `executor.submit()` calls to detect fleche-decorated functions
  - Returns cached results via completed `Future` objects without touching the executor
  - Automatically binds non-cached calls to carry cache/metadata state into workers
  - Intelligently splits keyword arguments between executor-reserved parameters (e.g., `resource_dict`) and function payload arguments
  
- **Updated `src/fleche/__init__.py`**: Exports `wrap_executor` in the public API

- **Comprehensive test coverage** in `tests/integration/test_parallel_execution.py`:
  - `test_wrap_executor_passthrough_non_fleche`: Non-fleche functions pass through unchanged
  - `test_wrap_executor_thread_cache_miss`: Cache misses bind state and submit normally
  - `test_wrap_executor_cache_hit_skips_submit`: Cache hits return completed futures without calling submit
  - `test_process_executor_wrap_executor`: Works with ProcessPoolExecutor and file-backed caches
  - `test_wrap_executor_splits_submit_kwargs`: Correctly partitions executor-specific kwargs from function arguments
  - `test_executorlib_wrap_executor`: Compatible with executorlib.SingleNodeExecutor

## Implementation Details
- Uses introspection (`inspect.signature`) to detect keyword-only parameters on executor's `submit` method, enabling compatibility with custom executors that declare their own parameters
- Handles both in-memory and file-backed caches transparently
- Works with any executor implementing the standard `submit(func, *args, **kwargs)` interface
- Non-invasive: patches the instance's `submit` attribute rather than subclassing

https://claude.ai/code/session_01SwPyQTm6foVSXLAY17KwYL

### DIFF
--- a/src/fleche/executor.py
+++ b/src/fleche/executor.py
@@ -1,0 +1,92 @@
+"""Executor wrapper that intercepts ``submit`` for fleche-decorated functions.
+
+Motivation: users passing a :func:`fleche.fleche`-decorated function to
+``executor.submit(...)`` have to remember to call :meth:`.BoundWrapper.bind`
+themselves to carry the active cache/metadata state into the worker.  They also
+pay the submit/serialisation cost even when the result is already cached.
+
+:func:`wrap_executor` patches the ``submit`` method of an executor *instance*
+(we cannot subclass, since callers pass us instances of third-party executors)
+so that:
+
+* non-fleche callables are forwarded unchanged,
+* fleche callables whose result is already cached are returned via an
+  already-completed :class:`~concurrent.futures.Future` without touching the
+  executor, and
+* otherwise the call is bound via :meth:`.BoundWrapper.bind` and submitted to
+  the original ``submit``.
+
+Executors that declare their own keyword-only parameters on ``submit``
+(e.g. ``resources=`` or ``resource_dict=``) have those split off from the
+caller's ``**kwargs`` and forwarded to the underlying ``submit`` while the
+remaining keyword arguments are bound as part of the function payload.
+"""
+
+from concurrent.futures import Future
+from inspect import signature, Parameter
+from types import SimpleNamespace
+
+
+__all__ = ["wrap_executor"]
+
+
+def _is_fleche_function(func) -> bool:
+    return isinstance(getattr(func, "fleche", None), SimpleNamespace)
+
+
+def _split_submit_kwargs(submit_method, kwargs):
+    """Partition ``kwargs`` into (executor-reserved, function-payload).
+
+    Any keyword-only parameter declared on ``submit_method``'s signature is
+    reserved for the executor; everything else (including arguments matched by
+    a ``**kwargs`` catch-all) is forwarded to the callable.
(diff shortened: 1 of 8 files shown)


tests/integration/test_parallel_execution.py
```diff
@@ -276,7 +276,177 @@ def test_executorlib_bound_wrapper():
...
+    from executorlib import SingleNodeExecutor
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        values_dir = f"{tmpdir}/values"
+        calls_dir = f"{tmpdir}/calls"
+
+        file_cache = Cache(
+            ValuePickleFile.with_pickle(root=values_dir),
+            CallPickleFile.with_pickle(root=calls_dir),
+        )
```
pmrv: @claude I think we have a fixture for that, if not create one.

pmrv: **Claude finished @pmrv's task in 2m 22s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24730176599)

---
### Tasks

- [x] Understand the request
- [x] Read line 444 context — `test_executorlib_wrap_executor` and `test_process_executor_wrap_executor` both created a file-backed `Cache` manually with `tempfile.TemporaryDirectory()`
- [x] Check `tests/fixtures.py` — no `file_cache` fixture existed yet
- [x] Add `file_cache` fixture to `tests/fixtures.py`
- [x] Update the two new tests to use the fixture
- [x] Commit and push changes

Added a `file_cache` fixture to `tests/fixtures.py` using pytest's built-in `tmp_path`:

```python
@pytest.fixture
def file_cache(tmp_path):
    yield Cache(
        ValuePickleFile.with_pickle(root=tmp_path / "values"),
        CallPickleFile.with_pickle(root=tmp_path / "calls"),
    )
```

Both `test_process_executor_wrap_executor` and `test_executorlib_wrap_executor` now take `file_cache` as a parameter instead of manually setting up `tempfile.TemporaryDirectory()`.

src/fleche/executor.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if func.fleche.contains(*args, **func_kwargs):
+            fut: Future = Future()
+            fut.set_result(func.fleche.load(*args, **func_kwargs))
+            return fut
+
+        bound = func.fleche.bind(*args, **func_kwargs)
+        return original_submit(bound, **submit_kwargs)
+
+    submit._fleche_wrapped = True
```
pmrv:
```suggestion
    submit._fleche_wrapped = True  # type: ignore
```

PR #381, opened 2026-04-21
Document the three executor-interop features

pmrv:
Stacked on top of #380.

## Summary

Document all three executor-related features side-by-side in `docs/parallel_execution.rst` and the `ConcurrentExecution` notebook:

1. **`BoundWrapper.bind(func)`** (existing) — explicit state capture into a picklable callable.
2. **Future pass-through** (existing) — a fleche-decorated function returns a `Future`; fleche caches on completion via a done-callback.
3. **`wrap_executor(executor)`** (new in #380) — monkey-patches `submit` so fleche functions are bound on miss and cache hits return an already-completed `Future` without touching the executor.

## Changes

- `docs/parallel_execution.rst`:
  - New introductory overview listing all three APIs and when to pick each.
  - New `wrap_executor` section with ProcessPool example and an executorlib example that demonstrates `resource_dict=` being correctly split off from the function payload.
  - Reworked "When to prefer each pattern" table into a 3-column comparison (BoundWrapper / Future pass-through / wrap_executor).
  - Updated Quick Reference to recommend `wrap_executor` as the lightest-touch option.
- `notebooks/ConcurrentExecution.ipynb`: restructured around the three APIs. New cells for Future pass-through and `wrap_executor` (thread pool, process pool with file cache, kwarg splitting). Verified end-to-end via `nbclient` — all output assertions match.
- `docs/notebooks/ConcurrentExecution.ipynb`: new symlink, wired into `docs/index.rst` Notebooks toctree.

## Test plan

- [x] Executed the notebook end-to-end with `nbclient`; all `print(...)` outputs match their inline expectations.
- [x] `nbformat.validate` passes on the rewritten notebook.
- [x] No code changes under `src/` — the `src/fleche` test suite is untouched and still passes on the base branch.

### DIFF
--- a/docs/parallel_execution.rst
+++ b/docs/parallel_execution.rst
@@ -5,6 +5,21 @@ Fleche-decorated functions work with Python's ``concurrent.futures`` executors
 and other process-/thread-based parallelism libraries.  This page explains the
 recommended patterns for each executor type.
 
+Three complementary APIs cover the common cases:
+
+- :class:`~fleche.BoundWrapper` — freeze the active cache and metadata into a
+  picklable callable that a worker can invoke without any further setup.  Use
+  when you own the submission site and want explicit control.
+- **Future pass-through** — a fleche-decorated function that internally
+  returns a :class:`~concurrent.futures.Future` is cached automatically when
+  the future completes.  Use when the decorated function *is* the one that
+  submits work.
+- :func:`~fleche.wrap_executor` — a one-line wrapper around an executor
+  instance that transparently binds fleche-decorated functions on ``submit``
+  **and** short-circuits cache hits without ever touching the executor.  Use
+  when you want the most compact call site, or to avoid submit/serialise
+  overhead on cached inputs.
+
 .. contents:: On this page
    :local:
    :depth: 2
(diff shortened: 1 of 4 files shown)


docs/parallel_execution.rst
```diff
@@ -188,33 +203,112 @@ automatically caches the result once the future completes:
...
+               heavy_computation, 5,
+               resource_dict={"cores": 4},   # goes to SingleNodeExecutor.submit
+           )
+           assert future.result() == 125
+
+.. note::
+
+   :func:`~fleche.wrap_executor` mutates the instance it is given; it does
+   not return a proxy.  Wrap once per executor — wrapping a second time
+   stacks another interception layer on top of the first.
```
pmrv: @claude Work in #380 to make the wrapping idempotent.

pmrv: **Claude finished @pmrv's task in 1m 54s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24732021401)

---
### Making `wrap_executor` idempotent

- [x] Read the relevant docs line and understand the idempotency concern
- [x] Find the `wrap_executor` implementation in the base branch (#380)
- [x] Implement idempotency guard
- [x] Add test for idempotency
- [x] Update the docs note
- [x] Commit and push

---

Done. Here's what changed across 3 files in commit `3ddd510`:

**`src/fleche/executor.py`** — The patched `submit` closure now carries a `_fleche_wrapped = True` attribute. `wrap_executor` checks for it at entry and returns early if found, so a second call is a no-op.

```python
_WRAPPED_SENTINEL = "_fleche_wrapped"

def wrap_executor(executor):
    if getattr(executor.submit, _WRAPPED_SENTINEL, False):
        return executor
    ...
    setattr(submit, _WRAPPED_SENTINEL, True)
    executor.submit = submit
    return executor
```

**`tests/integration/test_parallel_execution.py`** — New `test_wrap_executor_idempotent` that wraps a `ThreadPoolExecutor` twice and asserts the `submit` attribute is the same object after both calls.

**`docs/parallel_execution.rst`** — Updated the note at line 279 from "wrapping a second time stacks another interception layer" to "calling it on an already-wrapped executor is a no-op."

All 12 integration tests pass (4 skipped due to missing `executorlib`).

pmrv: Done in #380 as `0b0eeb1` "Make wrap_executor idempotent": the patched `submit` carries a `_fleche_wrapped = True` marker and `wrap_executor` short-circuits if it sees it, so a second call is a no-op. Added `test_wrap_executor_is_idempotent` and reworded the note here from "wrapping a second time stacks another interception layer" to "calling it on an already-wrapped executor is a no-op".

Note: a parallel `claude[bot]` run had put an equivalent fix (`3ddd510`) directly on the docs branch; I force-pushed this branch to place the implementation commit on the `submit` branch instead (per the review), and `3ddd510` is now superseded.

---

PR #385, opened 2026-04-22
test(call): cover QueryCall.matches metadata filter branches

pmrv:
## Summary

Running branch coverage over the full test suite flagged `src/fleche/call.py` as
the lowest-covered module after `__init__.py`. The largest cohesive gap was
lines **204-214** — the metadata-filter block of `QueryCall.matches()` — which
no existing test exercised (all prior `test_query_call_matches*` cases leave
`self.metadata` empty).

This PR adds one orthogonal, singly-focused test `test_query_call_matches_metadata`
in `tests/unit/call/test_matches.py`. It exercises every branch of the metadata
block in a single feature-scoped test:

- wildcard filter `filters=None` on a present metadata name → accept
- filter value equal to stored → accept
- filter value differs from stored → reject
- presence-only filter (`vv is None`), key present → accept
- presence-only filter (`vv is None`), key absent → reject
- required metadata name absent on the candidate → reject

## Coverage report (before, full suite with `--cov-branch`)

```
Name                                      Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------------------------------------
src/fleche/__init__.py                       12      3      2      1    71%   5-6, 43
src/fleche/_version.py                       11      0      0      0   100%
src/fleche/caches.py                        300     36     74      8    88%
src/fleche/call.py                          131     13     48      5    83%   58->60, 85, 172, 173->175, 192, 204-214
src/fleche/config.py                        172      7     78      6    95%
src/fleche/digest.py                        132     18     72     10    84%
src/fleche/metadata.py                       33      1      0      0    97%
src/fleche/query.py                         109      0     46      1    99%
src/fleche/security.py                       57      3     26      2    94%
src/fleche/state.py                          52      1     10      1    97%
src/fleche/storage/__init__.py               10      0      0      0   100%
src/fleche/storage/bagofholding_file.py      40      0      8      1    98%
src/fleche/storage/base.py                  114      0     32      0   100%
src/fleche/storage/destructuring.py         106      0     44      3    98%
src/fleche/storage/file.py                   69      2      8      0    97%
src/fleche/storage/memory.py                 24      0      4      0   100%
src/fleche/storage/pickle_file.py            82      4     16      0    96%
src/fleche/storage/sql.py                   237     10     86      6    94%
src/fleche/storage/thread_safe.py            54      0      4      1    98%
src/fleche/storage/void.py                   20      1      4      0    96%
src/fleche/wrapper.py                       226      1     54      1    99%
-------------------------------------------------------------------------------------
TOTAL                                      1991    100    616     46    94%
```

After this PR, lines `204-214` drop out of the `call.py` `Missing` column (confirmed
by a focused run of `tests/unit/call/`). Remaining gaps in `call.py` (`58->60, 85,
172, 173->175, 192`) are unrelated to metadata matching and left for follow-up.

## Test plan

- [x] `pytest tests/unit/call/test_matches.py -v` — 3 passed
- [x] Confirmed that the three assertions covering the three major branch points
      (name missing / presence-only / value equality) each fail independently if
      the corresponding assertion is negated.

https://claude.ai/code/session_01UPAcVakLTyEbvkuFeMTN2R

### DIFF
--- a/tests/unit/digest/test_digest.py
+++ b/tests/unit/digest/test_digest.py
@@ -32,26 +32,6 @@ def test_custom_digest():
     assert isinstance(digest(CustomMethod("method")), Digest)
 
 
-def test_custom_digest_priority():
-    """Test that __digest__ has priority over default implementations."""
-
-    class CustomStr(str):
-        def __digest__(self):
-            return "custom_str_digest"
-
-    class CustomInt(int):
-        def __digest__(self):
-            return "custom_int_digest"
-
-    class CustomDict(dict):
-        def __digest__(self):
-            return "custom_dict_digest"
-
-    assert digest(CustomStr("hello")) == "custom_str_digest"
-    assert digest(CustomInt(42)) == "custom_int_digest"
-    assert digest(CustomDict(a=1)) == "custom_dict_digest"
-
-
 def test_supported_types():
     """Test that all supported types can be digested without raising an exception."""
 
--- a/tests/unit/digest/test_digest.py
+++ b/tests/unit/digest/test_digest.py
@@ -32,6 +32,26 @@ def test_custom_digest():
     assert isinstance(digest(CustomMethod("method")), Digest)
 
 
+def test_custom_digest_priority():
+    """Test that __digest__ has priority over default implementations."""
+
+    class CustomStr(str):
+        def __digest__(self):
+            return "custom_str_digest"
+
+    class CustomInt(int):
+        def __digest__(self):
+            return "custom_int_digest"
+
+    class CustomDict(dict):
+        def __digest__(self):
+            return "custom_dict_digest"
+
+    assert digest(CustomStr("hello")) == "custom_str_digest"
+    assert digest(CustomInt(42)) == "custom_int_digest"
+    assert digest(CustomDict(a=1)) == "custom_dict_digest"
+
+
 def test_supported_types():
     """Test that all supported types can be digested without raising an exception."""
 
--- a/tests/unit/fleche/test_ignore_digest.py
+++ b/tests/unit/fleche/test_ignore_digest.py
@@ -1,7 +1,10 @@
 import logging
 
+import pytest
+
 from fleche import fleche, cache
 from fleche.caches import Cache
+from fleche.digest import Unhashable
 from fleche.storage import ValueMemory, CallMemory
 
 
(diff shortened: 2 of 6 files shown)


pmrv:
Step 2 landed in a follow-up commit on the same branch (0750e5b). Couldn't open a separate PR — session is scoped to this single branch — but happy to split if reviewers prefer.

**What the second commit does:** removes 5 coverage-redundant tests sampled at random from the unit suite.

## Knockout measurements (unit tests only, `--ignore=tests/unit/test_pickle.py`)

| Run | Stmts | Miss | Branch | BrPart | Cover |
|---|---|---|---|---|---|
| Baseline | 1991 | 95 | 616 | 47 | 94% |
| All 6 deselected | 1991 | **97** | 616 | 47 | 94% |
| 5/6 deselected | 1991 | 95 | 616 | 47 | 94% |

The +2-stmt delta in the all-6 row is entirely from `test_cache_function_loads_by_name` (only unit test instantiating `ValueBagOfHoldingH5File` through the config loader). **That test stays.** The other 5 produce zero delta — every module's `(Stmts, Miss, Branch, BrPart)` tuple is identical to the baseline:

- `test_cache_stack_evict_tolerates_key_missing_from_some_layers` — the `except (Rejected, KeyError)` branch in `CacheStack.evict` is never actually reached (the underlying `Cache.evict` silently drops missing keys via `MemoryBackend.pop(..., None)`), so removing the test doesn't lose any branch.
- `test_count_empty` — `QueryIterator.count` is `sum(1 for _ in self)`; the non-empty `test_count_returns_correct_number` hits the same bytecode.
- `test_digest_helper_in_worker` — the `.digest` helper and `ProcessPoolExecutor` pickling paths are already exercised by the surrounding tests in the same file.
- `test_unignored_unhashable_raises_but_ignored_is_ok` — the `ignore=` code path is covered by the file's other 9 tests.
- `test_custom_digest_priority` — `__digest__` dispatch is hit by every object that implements it (`LazyArguments`, `LazyCall`, `Digested` / `DigestedIterable` / `DigestedDict`) plus the preceding `test_custom_digest_method` in the same file.

I applied the rule strictly ("if coverage was not impacted, remove"). Two of these (`test_cache_stack_evict_tolerates_...` as an evict-tolerance contract, `test_digest_helper_in_worker` as an issue #172 regression) could be argued as semantic contracts worth preserving even without coverage impact — happy to revert either individually on review.

---

tests/unit/digest/test_digest.py
```diff
@@ -32,26 +32,6 @@ def __digest__(self):
     assert isinstance(digest(CustomMethod("method")), Digest)
 
 
-def test_custom_digest_priority():
```
pmrv, line 35: keep

tests/unit/fleche/test_ignore_digest.py
```diff
@@ -96,26 +93,6 @@ def f_tuple(a, b):
     assert k3a == k3b
 
 
-def test_unignored_unhashable_raises_but_ignored_is_ok():
```
pmrv, line 99: keep

pmrv (commented):
@claude

pmrv:
**Claude finished @pmrv's task in 2m 4s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24781736281)

---
### Done

- [x] Read git diff to identify deleted tests at indicated lines
- [x] Restore `test_custom_digest_priority` to `tests/unit/digest/test_digest.py` (after `test_custom_digest`, at line 35)
- [x] Restore `test_unignored_unhashable_raises_but_ignored_is_ok` to `tests/unit/fleche/test_ignore_digest.py` (after `test_ignore_string_list_tuple_equivalent`, at line 96) — also restored the `import pytest` and `from fleche.digest import Unhashable` imports that were removed alongside it
- [x] Both tests pass (`2 passed in 0.85s`)
- [x] Committed and pushed as `e7d8307`

PR #386, opened 2026-04-22
docs: clarify BoundWrapper is a plain callable without .fleche namespace

pmrv:
Document that bind() returns a plain BoundWrapper callable that does not expose the .fleche helper namespace. Users who need helpers like digest() or query() should access them via bound.func.fleche.*.

### DIFF
--- a/src/fleche/state.py
+++ b/src/fleche/state.py
@@ -127,12 +127,17 @@ def project(name):
 
 @dataclass(frozen=True, eq=True)
 class BoundWrapper:
-    """Utility class that freezes global state for the cache and metadata config.
+    """A plain callable that freezes cache and metadata state at construction time.
 
-    Essentially acts like an early binding closure.
+    :class:`.BoundWrapper` is intentionally a minimal wrapper: it captures the active
+    :class:`.BaseCache` and metadata tuple and restores them around every call to
+    :attr:`.func`, but it does **not** expose the ``fleche`` helper namespace
+    (``digest``, ``call``, ``load``, ``contains``, ``query``, ``rerun``).  Those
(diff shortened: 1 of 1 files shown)


src/fleche/state.py
```diff
@@ -127,12 +127,17 @@ def project(name):
...
-    """Utility class that freezes global state for the cache and metadata config.
+    """A plain callable that freezes cache and metadata state at construction time.
 
-    Essentially acts like an early binding closure.
+    :class:`.BoundWrapper` is intentionally a minimal wrapper: it captures the active
+    :class:`.BaseCache` and metadata tuple and restores them around every call to
+    :attr:`.func`, but it does **not** expose the ``fleche`` helper namespace
+    (``digest``, ``call``, ``load``, ``contains``, ``query``, ``rerun``).  Those
+    helpers live on the decorated function itself and can still be reached via
+    ``bound.func.fleche.*``.
```
pmrv: @claude no, do not advertise internal attributes of boundwrapper.  Users still have the option to call the methods on the main function.

pmrv: **Claude finished @pmrv's task in 1m 3s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24781803704)

---
### Tasks

- [x] Gather context
- [x] Read `src/fleche/state.py` to understand what's being advertised at line 137
- [x] Remove `bound.func` internal attribute references, point users to original decorated function
- [x] Commit and push

Done. Replaced both occurrences of `bound.func.fleche.*` in the `BoundWrapper` class docstring and `bind()` method docstring with "the original decorated function", avoiding exposure of the internal `.func` attribute.

PR #395, opened 2026-04-22
fix(benchmarks): realign scripts with current storage API

pmrv:
## Summary

The benchmark scripts in `benchmarks/` were broken against the current codebase — they still imported `Memory`, `PickleFile`, `BagOfHoldingH5File` from `fleche.storage`, which the Value*/Call* split removed. On top of that, the random nested-values workload produced inputs that plain pickle and H5Bag could not serialize, and `run_benchmarks.py` swallowed any sub-script failure into an empty result list so the resulting report was silently missing entire tables.

- Update imports to `ValueMemory` / `CallMemory` / `ValuePickleFile` / `ValueBagOfHoldingH5File` so every sub-script starts.
- Replace the hypothesis `st_nested_values` workload with a seeded RNG over backend-safe primitives; the previous random dataclasses broke stdlib pickle and random text/binary broke H5Bag, depending on the run.
- Drop the dead "stop-gap" `contains` fallbacks (superseded by the real API) and the blanket `try/except` wrappers that silenced genuine backend errors.
- Make `run_benchmarks.run_script` abort with a traceback on any sub-script failure (non-zero exit or invalid JSON). A broken script can no longer silently elide its whole table.

## Test plan

- [x] `python -m benchmarks.run_benchmarks` prints **Digest**, **Value Storage** (small_strings, nested_structures, numpy_arrays × 8 configs × 5 benchmarks), **Call Storage** (calls × 2 × 5), **Integration** (3 workloads × 6 configs × 4) — 213 rows in `results.csv`, no stderr errors.
- [x] `python benchmark_storage.py`, `python benchmark_integration.py`, `python benchmark_digest.py` each exit 0 with complete JSON.
- [x] A synthetic failing sub-script makes `run_benchmarks` raise a `RuntimeError` naming the script, instead of producing a truncated report.

https://claude.ai/code/session_015vN2GYk2Bcm63aAgrgKv6f

### DIFF
--- a/benchmarks/benchmark_digest.py
+++ b/benchmarks/benchmark_digest.py
@@ -6,7 +6,7 @@ from hypothesis import strategies as st
 from hypothesis.extra.numpy import arrays as st_arrays
 
 from fleche.digest import digest
-from utils import st_nested_values, st_nested_values, generate_examples
+from utils import st_nested_values, generate_examples
 
 import timeit
 
--- a/benchmarks/benchmark_integration.py
+++ b/benchmarks/benchmark_integration.py
@@ -8,7 +8,13 @@ import numpy as np
 
 from fleche import fleche, cache
 from fleche.caches import Cache, SizeLimitedCache
-from fleche.storage import Memory, PickleFile, Sql, BagOfHoldingH5File
+from fleche.storage import (
+    ValueMemory,
+    CallMemory,
+    ValuePickleFile,
+    ValueBagOfHoldingH5File,
+    Sql,
+)
 
 
 # Define some test functions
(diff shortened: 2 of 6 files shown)


pmrv:
@claude I don't like you dropping the hypothesis strategies.  From the value sub strategies you may assemble one instead that is compatible with all backends.  Keep some options open to control depth and size of data.

pmrv:
**Claude finished @pmrv's task in 6m 58s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24794864889)

---
### Restoring Hypothesis Strategies (Backend-Compatible)

- [x] Read current `benchmark_storage.py` and `utils.py` to understand the situation
- [x] Review git history to understand original hypothesis strategies and constraints
- [x] Understand backend serialization constraints
- [x] Add `st_backend_safe_nested(max_depth, max_size)` to `utils.py`
- [x] Replace seeded RNG in `benchmark_storage.py` with hypothesis-based generation
- [x] Commit and push changes

---

**What changed:**

`benchmarks/utils.py` — new function `st_backend_safe_nested(max_depth=3, max_size=6)` assembled from the safe sub-strategies:

```python
_backend_safe_leaf_strategies = [
    st.none(),
    st.booleans(),
    st.integers(),
    st.floats(allow_nan=False, allow_infinity=False),
    st.text(
        alphabet=string.ascii_letters + string.digits + string.punctuation + " ",
        max_size=50,
    ),
]
```

Containers: `st.lists(children, max_size=max_size)` and `st.dictionaries(str-key, children, max_size=max_size)` via `st.recursive`. `max_leaves = max_depth * max_size` so both parameters meaningfully control the output.

Excluded from the full `st_nested_values`:
- `st.binary()` — H5Bag can't store raw byte strings
- `st.text()` (full unicode) — surrogates/null bytes break HDF5 strings
- `make_dataclass` outputs — not importable by stdlib pickle at unpickling time
- `Call` objects — memoization metadata, not stored values

`benchmarks/benchmark_storage.py` — the seeded-RNG block replaced by a small helper that drives the strategy with `@given`:

```python
def _generate_backend_safe_nested(count=50, max_depth=3, max_size=6):
    collected = []

    @settings(
        max_examples=count,
        suppress_health_check=list(HealthCheck),
        database=InMemoryExampleDatabase(),
    )
    @given(val=st_backend_safe_nested(max_depth=max_depth, max_size=max_size))
    def _collect(val):
        collected.append(val)

    _collect()
    return collected
```

`st_nested_values` is untouched — `benchmark_digest.py` still uses it (no backend round-trip constraints there).
 | [Branch](https://github.com/pmrv/fleche/tree/claude/fix-benchmark-scripts-6KbhB)

PR #396, opened 2026-04-22
Use qualified names and improve version extraction for function calls

pmrv:
## Summary
This PR improves how function metadata is extracted and stored in `Call` objects by using qualified names instead of simple function names, and by leveraging `pyiron_snippets.versions` for more robust version extraction.

## Key Changes

- **Added `_extract_version_info()` helper function** that uses `pyiron_snippets.versions` utilities to extract module, qualified name, and version information from functions. This provides better handling of edge cases like modules and gracefully falls back to `None` when a module is not importable.

- **Updated `Call.from_call()` and `QueryCall.from_call()`** to use the new helper function, replacing direct attribute access (`func.__name__`, `func.__module__`) with qualified names via `get_qualname()`. This ensures that nested functions and methods are properly identified.

- **Changed version type annotation** from `int | None` to `str | int | None` across `Call`, `LazyCall`, and `QueryCall` classes to accommodate string versions returned by `pyiron_snippets.versions.get_version()`.

- **Updated test fixtures** to set `__qualname__` attributes on mock functions to match the new behavior, ensuring tests properly simulate real function metadata.

- **Updated integration and unit tests** to use `func.__wrapped__.__qualname__` when constructing expected call names, reflecting the change from simple names to qualified names.

## Implementation Details

- The `_extract_version_info()` function prioritizes `func.__version__` if it exists, falling back to the extracted version from the module. This maintains backward compatibility while providing better version detection.
- The helper function gracefully handles `ModuleNotFoundError` by returning `None` for the version rather than propagating the exception, making the code more robust.
- All test updates ensure that the caching and querying behavior works correctly with qualified names, which better distinguish between functions with the same simple name in different scopes.

https://claude.ai/code/session_016CCgE8GWx9W1BTnZYXWLEC

### DIFF
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -8,7 +8,7 @@ from pyiron_snippets.versions import VersionInfo, get_module, get_qualname
 from . import digest
 
 
-def _extract_version_info(func) -> tuple[str, str, str | None]:
+def _extract_version_info(func) -> tuple[str, str, str | int | None]:
     """Extract ``(name, module, version)`` from ``func`` via :mod:`pyiron_snippets.versions`.
 
     Uses :meth:`VersionInfo.of` to introspect ``func``.  When the module is not
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -1,11 +1,15 @@
+import logging
 from dataclasses import dataclass, field, replace
-from typing import Any
+from typing import Any, Callable
 from inspect import signature
 from collections.abc import Mapping
 
 from pyiron_snippets.versions import VersionInfo, get_module, get_qualname
 
 from . import digest
+from .digest import Digest
+
+logger = logging.getLogger("fleche.call")
 
 
 def _extract_version_info(func) -> tuple[str, str, str | int | None]:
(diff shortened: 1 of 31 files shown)


src/fleche/call.py
```diff
@@ -3,9 +3,28 @@
 from inspect import signature
 from collections.abc import Mapping
 
+from pyiron_snippets.versions import get_module, get_qualname, get_version
+
 from . import digest
 
 
+def _extract_version_info(func) -> tuple[str, str, str | None]:
```
pmrv: Use `VersionInfo.of()`

pmrv: I also want to ensure that versions defined directly on the function have priority.  Iirc this is somehow implemented in snipppets, check their code first how to do it.

pmrv (commented):
> * **Changed version type annotation** from `int | None` to `str | int | None` across `Call`, `LazyCall`, and `QueryCall` classes to accommodate string versions returned by `pyiron_snippets.versions.get_version()`.

@claude Check explicitly that this doesn't mess up the SQL backend.  It will likely require change of table format and extra tests that check the 

See also the other comments.

pmrv:
**Claude finished @pmrv's task in 43s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24838090755)

---
### Done — addressing review feedback on PR #396

- [x] Read AGENTS.md and understand repo guidelines
- [x] Read `src/fleche/call.py` and understand current implementation
- [x] Investigated `pyiron_snippets.versions` — fetched source from GitHub; confirmed `VersionInfo.of()` does **not** have built-in function-level `__version__` priority (it is purely module-based)
- [x] Inspected SQL backend — `CallModel.version` was `Integer` — **confirmed broken for string versions**
- [x] Implemented changes (840 passed, 4 skipped ✅)
- [x] Pushed to branch

---

### Changes made

**`src/fleche/call.py`** — use `VersionInfo.of()`

Replaced the three individual imports (`get_module`, `get_qualname`, `get_version`) with `VersionInfo.of()`. The fallback on `ModuleNotFoundError` still uses `get_module`/`get_qualname` (kept in import) since `VersionInfo.of()` has no partial-result API when version resolution fails. Function-level `__version__` priority is implemented explicitly via `getattr(func, "__version__", info.version)` — `VersionInfo.of()` has no built-in mechanism for this, so fleche must continue to handle it.

**`src/fleche/storage/sql.py`** — fix SQL backend for `str | int | None` versions

`CallModel.version` was `Mapped[int] = mapped_column(Integer, ...)`. Storing a string version (`"1.2.3"`) in an `Integer` column silently corrupts data (SQLite's type affinity would let it through, but strict backends like PostgreSQL would reject it outright; and even SQLite would break on query comparisons).

Fix: change column to `String`, store as `json.dumps(version)`, load as `json.loads(version)`. Using `json.dumps`/`json.loads` rather than a raw `JSON` column avoids SQLAlchemy's inconsistent JSON equality-comparison behaviour in WHERE clauses (which caused queries to return 0 rows for string versions in testing). The query condition compares against `json.dumps(template.version)` to match the stored encoding. Type is fully preserved on round-trip: `int 42` → `"42"` → `42`, `str "1.2.3"` → `'"1.2.3"'` → `"1.2.3"`.

**`tests/unit/storage/test_sql_query.py`** — three new SQL version tests

- `test_sql_string_version_roundtrip` — string version saves and loads with correct type
- `test_sql_int_version_roundtrip` — integer version saves and loads with correct type (regression)
- `test_sql_query_by_string_version` — querying by string version finds only the matching call
 | [Branch](https://github.com/pmrv/fleche/tree/claude/extract-function-metadata-92IhO)

pmrv:
@claude provide a script to update existing sqlite databases.  Do not check it into the repo, just post here for convenience.

pmrv:
**Claude finished @pmrv's task in 2m 2s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24839264448)

---
Here's a migration script for existing SQLite databases. The schema change is:

- **Column type**: `version INTEGER` → `version TEXT`  
- **Data encoding**: raw integer `42` → JSON-encoded string `"42"` (i.e. `json.dumps(42)`)

This matters for correctness: SQLite's INTEGER type affinity would silently re-coerce `"42"` back to integer `42`, breaking the `WHERE version = '"42"'` query comparisons the new code relies on.

```python
#!/usr/bin/env python3
"""
Migrate a fleche SQLite database so that calls.version stores
JSON-encoded text instead of raw integers.

  Schema change:  version INTEGER  →  version TEXT
  Data change:    42               →  "42"   (json.dumps encoding)

Usage:
    python migrate_fleche_db.py path/to/fleche.db [--dry-run]
"""

import argparse
import json
import shutil
import sqlite3
import sys
from datetime import datetime
from pathlib import Path


def migrate(db_path: str, dry_run: bool = False) -> None:
    path = Path(db_path)
    if not path.exists():
        sys.exit(f"Error: {db_path!r} does not exist")

    con = sqlite3.connect(db_path)
    con.row_factory = sqlite3.Row
    try:
        if con.execute(
            "SELECT sql FROM sqlite_master WHERE type='table' AND name='calls'"
        ).fetchone() is None:
            sys.exit("No 'calls' table found — is this a fleche database?")

        col_info = {
            row["name"]: row["type"].upper()
            for row in con.execute("PRAGMA table_info(calls)")
        }
        current_type = col_info.get("version", "MISSING")
        if current_type == "MISSING":
            sys.exit("No 'version' column in 'calls' — nothing to migrate.")

        print(f"Current 'version' column declared type: {current_type}")

        rows = con.execute(
            "SELECT key, name, module, version, code_digest, result FROM calls"
        ).fetchall()
        print(f"Total rows in calls: {len(rows)}")
        non_null = [(r, r["version"]) for r in rows if r["version"] is not None]
        print(f"Rows with non-NULL version: {len(non_null)}")

        # Work out the new encoded value for each non-NULL version.
        new_versions: dict[str, str] = {}
        changed = []
        for row, v in non_null:
            if isinstance(v, int):
                encoded = json.dumps(v)               # 42  →  "42"
            elif isinstance(v, str):
                try:
                    parsed = json.loads(v)
                    encoded = v if json.dumps(parsed) == v else json.dumps(v)
                except (json.JSONDecodeError, ValueError):
                    try:
                        encoded = json.dumps(int(v))  # "42" as text  →  "42"
                    except ValueError:
                        encoded = json.dumps(v)
            else:
                encoded = json.dumps(v)

            new_versions[row["key"]] = encoded
            if str(v) != encoded:
                changed.append((row["key"], v, encoded))

        print(f"Values that need re-encoding: {len(changed)}")
        if changed:
            print("Sample conversions (first 5):")
            for key, old, new in changed[:5]:
                print(f"  key={key[:12]}…  {old!r}  →  {new!r}")

        if dry_run:
            print("\n[Dry run] No changes written.")
            return

        if current_type in ("TEXT", "VARCHAR") and not changed:
            print("Nothing to do — already up to date.")
            return

        backup = path.with_suffix(
            f".pre-migration-{datetime.now():%Y%m%d%H%M%S}.db"
        )
        shutil.copy2(path, backup)
        print(f"\nBackup created: {backup}")

        # SQLite can't ALTER COLUMN type; recreate the table instead.
        con.execute("PRAGMA foreign_keys=OFF")
        con.execute("BEGIN")
        try:
            con.execute("""
                CREATE TABLE calls_new (
                    key         TEXT NOT NULL PRIMARY KEY,
                    name        TEXT NOT NULL,
                    module      TEXT,
                    version     TEXT,
                    code_digest TEXT,
                    result      TEXT
                )
            """)
            for row in rows:
                key = row["key"]
                # None if the original version was NULL
                new_ver = new_versions.get(key)
                con.execute(
                    "INSERT INTO calls_new VALUES (?,?,?,?,?,?)",
                    (key, row["name"], row["module"], new_ver,
                     row["code_digest"], row["result"]),
                )
            con.execute("DROP TABLE calls")
            con.execute("ALTER TABLE calls_new RENAME TO calls")

            fk_issues = con.execute("PRAGMA foreign_key_check").fetchall()
            if fk_issues:
                con.execute("ROLLBACK")
                sys.exit(f"Foreign key violations after migration: {fk_issues}")

            con.execute("COMMIT")
            print(
                f"Migration complete: {len(rows)} rows migrated, "
                f"{len(changed)} version values re-encoded."
            )
        except Exception:
            con.execute("ROLLBACK")
            raise
        finally:
            con.execute("PRAGMA foreign_keys=ON")
    finally:
        con.close()


if __name__ == "__main__":
    p = argparse.ArgumentParser(
        description=__doc__,
        formatter_class=argparse.RawDescriptionHelpFormatter,
    )
    p.add_argument("db_path", help="Path to the fleche SQLite database file")
    p.add_argument(
        "--dry-run",
        action="store_true",
        help="Report what would change without modifying the database",
    )
    args = p.parse_args()
    migrate(args.db_path, dry_run=args.dry_run)
```

### Why full table recreation (not just `UPDATE`)?

SQLite's [type affinity](https://www.sqlite.org/datatype3.html) rules mean that even if you `UPDATE calls SET version = '42'` on an INTEGER-declared column, SQLite will silently coerce the text `"42"` back to integer `42` on storage. The new code queries with `WHERE version = '"42"'` (the JSON-dumps form), which would never match an integer `42` under SQLite's comparison rules (TEXT > INTEGER). Table recreation changes the column declaration to TEXT, giving it TEXT affinity so the string values are stored as-is.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/extract-function-metadata-92IhO)

pmrv:
@claude fix type problems

pmrv:
**Claude finished @pmrv's task in 5m 27s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24839599585)

---
### Done — fixing type problems

- [x] Gather context from CI failure
- [x] Read source files and identify root causes
- [x] Fix `_extract_version_info` return type (`str | None` → `str | int | None`) — `getattr(func, "__version__", ...)` returns `int` when version is set via `@fleche(version=42)`
- [x] Fix `wrapper.py` `fleche()` parameter (`version: int | None` → `str | int | None`) — consistent with the broader type system change
- [x] Remove unused `# ty: ignore` from `wrapper.py:222` (`wraps(func)(_query_func)`)
- [x] Remove unused `# type: ignore` from `config.py:250` (`_asdict_init_only` return)
- [x] Committed and pushed (`ca6af44`)

### Changes

| File | Change |
|------|--------|
| `src/fleche/call.py` | Return type of `_extract_version_info`: `tuple[str, str, str \| None]` → `tuple[str, str, str \| int \| None]` |
| `src/fleche/wrapper.py` | `fleche(version: int \| None)` → `fleche(version: str \| int \| None)`; removed unused `# ty: ignore` |
| `src/fleche/config.py` | Removed unused `# type: ignore` |
 | [Branch](https://github.com/pmrv/fleche/tree/claude/extract-function-metadata-92IhO)

pmrv:
@claude still

pmrv:
**Claude finished @pmrv's task in 14m 58s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24840145725)

---
### Done — fixing remaining CI type errors

- [x] Read CI log and identify actual errors
- [x] Root cause identified: main added `DigestedCall` class (with `version: int | None`) after the PR branch diverged. The auto-merge kept `DigestedCall.version: int | None` while `LazyCall.version` and `Call.version` were widened to `str | int | None` by the PR — causing two `ty` errors when those values were passed to `DigestedCall`
- [x] Fix `call.py`: incorporated main's `DigestedCall` class into the PR branch with `version: str | int | None`; also kept main's `LazyCall.to_lookup_key()` (which uses `DigestedCall`) — eliminating both type errors
- [x] Fix `caches.py`: removed unused `# ty: ignore` comment at the relevant line
- [x] 840 tests pass locally
- [x] Pushed (`87219eb`)

### What changed

| File | Change |
|------|--------|
| `src/fleche/call.py` | Added `DigestedCall` class (from main) with `version: str \| int \| None`; incorporated its `_to_digested`, `stash`, `digest`, `fetch` machinery; updated `LazyCall.to_lookup_key()` to use `DigestedCall` |
| `src/fleche/caches.py` | Removed unused `# ty: ignore` comment |
 | [Branch](https://github.com/pmrv/fleche/tree/claude/extract-function-metadata-92IhO)

PR #400, opened 2026-04-23
feat: structural deduplication for dataclasses in CAS (#3)

### LINKED ISSUE #3: Structural deduplication in CAS
Should support:
- [x] tuples/lists
- [x] dicts
- [ ] dataclasses

pmrv:
Closes #3

Adds `DigestedDataclass` to `destructuring.py`, parallel to `DigestedIterable` and `DigestedDict`.  Field values are stored as separate CAS entries so instances sharing the same sub-objects deduplicate automatically.

Reconstruction bypasses `__init__` / `__post_init__` via `object.__new__` + `object.__setattr__`, which handles every field variant transparently: regular fields, `init=False` fields, and `InitVar` fields.

Two-level fallback ensures dataclasses are never rejected.  38 new tests.

### DIFF
--- a/src/fleche/storage/destructuring.py
+++ b/src/fleche/storage/destructuring.py
@@ -1,5 +1,7 @@
+import hashlib
 from abc import ABC, abstractmethod
 from collections import Counter
+import dataclasses as _dataclasses
 from dataclasses import dataclass
 from numbers import Number
 from typing import Any, Callable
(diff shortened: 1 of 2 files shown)


pmrv:
@claude Debate pros/cons of storing `fields()` vs. `__getstate__()`.  Can you point me to the code in cpython that generates `__getstate__` for dataclasses?

pmrv:
**Claude finished @pmrv's task in 2m 3s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24880607162)

---
### `fields()` vs `__getstate__()` for CAS dataclass storage

- [x] Read implementation and form analysis
- [x] Research CPython `__getstate__` generation for dataclasses

---

#### Where CPython generates `__getstate__` for dataclasses

CPython **only synthesizes `__getstate__`** for dataclasses declared with `slots=True` (added in Python 3.10).  For ordinary non-slotted dataclasses, no `__getstate__` is generated; pickle falls back to `obj.__dict__` via the default protocol.

The generated pair lives in [`Lib/dataclasses.py`](https://github.com/python/cpython/blob/main/Lib/dataclasses.py) — search for `_dataclass_getstate`:

```python
def _dataclass_getstate(self):
    return [getattr(self, f.name) for f in fields(self)]

def _dataclass_setstate(self, state):
    for field, value in zip(fields(self), state):
        # object.__setattr__ because frozen=True raises FrozenInstanceError
        object.__setattr__(self, field.name, value)
```

These are injected as methods when `_process_class` rebuilds the class with `__slots__`.  The takeaway: **CPython's own `__getstate__` for dataclasses is just a thin wrapper around `fields()`**, which is a strong signal that `fields()` is the right abstraction at this level.

---

#### Debate: `fields()` vs `__getstate__()`

**`fields()` — current approach**

| | |
|---|---|
| ✅ | **CAS-appropriate semantics.** Two instances that agree on all declared field values are logically identical. `fields()` captures exactly that set — nothing more, nothing less. |
| ✅ | **Stable key space.** The field names and order are fixed at class-definition time. `__dict__` can silently gain extra keys (lazy attributes, `__post_init__` temporaries) that would break content-addressing. |
| ✅ | **Correct for `init=False`.** `fields()` includes them; `getattr` fetches their post-`__post_init__` values. Reconstruction via `object.__setattr__` replays those computed values directly, bypassing `__post_init__` entirely — which is what you want. |
| ✅ | **No pickle coupling.** `__getstate__` is a pickle hook; pulling it into CAS storage merges two unrelated serialization concerns. |
| ✅ | **Custom `__getstate__` can't interfere.** A class author might define `__getstate__` to omit or transform fields for pickling efficiency (e.g. dropping a large computed cache). Using `fields()` ignores such customizations and stores the full logical state. |
| ⚠️ | **Misses non-field instance attributes.** If `__post_init__` does `self._cache = {}` without a `field()` declaration, `fields()` silently skips it. This is an unusual pattern but valid Python. |
| ⚠️ | **`InitVar` docstring is slightly misleading.** `InitVar` fields are correctly absent from `fields()` (they're constructor-only parameters, not instance attributes), so they're simply not stored or restored — which is correct. The docstring saying they're "handled uniformly" overstates this; they're handled by *being invisible* to `fields()`. |

**`__getstate__()`**

| | |
|---|---|
| ✅ | **Captures everything the class considers state.** For classes with a well-designed custom `__getstate__`, using it would respect the author's intent about what constitutes the object's serializable state. |
| ✅ | **Handles non-field instance attributes** set in `__post_init__`. |
| ❌ | **No synthesized `__getstate__` for non-slotted dataclasses.** You'd have to fall back to `obj.__dict__`, which is a superset of the declared fields and can contain ephemeral keys. |
| ❌ | **`__dict__` is CAS-hostile.** Two instances that are semantically equal can have different `__dict__` contents (e.g. one has a `_computed` cache key, the other doesn't), producing different CAS keys even though they represent the same value. |
| ❌ | **Custom `__getstate__` may be lossy.** If a class excludes fields from `__getstate__` for pickling purposes (common for large computed arrays), those values vanish from CAS too, making `mend()` reconstruct a different object than what was stored. |
| ❌ | **Format is not standardized.** `__getstate__` can return a dict, list, tuple, or anything. `mend()` would need to handle all cases or impose a contract that user classes can violate. |
| ❌ | **`__setstate__` side effects.** Using `__setstate__` for reconstruction might trigger `__post_init__`-style logic inside `__setstate__`, re-deriving computed values instead of restoring the ones we stored — defeating deduplication. |
| ❌ | **When it does exist for slotted dataclasses, it's just `fields()` anyway** — so you'd pay the extra complexity for no benefit in the common case. |

---

#### Verdict

`fields()` is the right choice for CAS.  The core CAS invariant is *same content → same key*; `fields()` defines "content" as the declared fields, which is stable, predictable, and matches the dataclass contract.  `__getstate__` is a pickle hook whose semantics are deliberately flexible in ways that are harmful for content-addressing.  The fact that CPython itself uses `fields()` inside the synthesized `__getstate__` for slotted dataclasses confirms this is the natural primitive.

The one gap worth acknowledging: **non-field instance attributes** (set in `__post_init__` outside any `field()` declaration) are invisible to the current implementation.  Whether that matters depends on whether fleche users are expected to have such attributes — for a typical scientific/numerical workflow this seems unlikely, but a comment in the docstring acknowledging the limitation would be prudent.

PR #423, opened 2026-04-24
feat: add Cache.gc() to evict values unreachable from any call

pmrv:
Baseline brute-force mark-and-sweep: scans every DigestedCall in call storage, collects the set of value digests referenced by arguments and result, then transitively follows destructured sub-references (DigestedIterable/DigestedDict/DigestedDataclass) until the frontier drains. Any value key not in the reachable set is evicted; call records are untouched.

To walk the reference graph without mending, factored the wrapper- matching logic out of DestructuringMixin.count_reuses into a new sub_digests(key) helper that returns the direct digest children of a raw entry. Value storages without this helper (e.g. ValueVoid) fall back to direct-reference reachability only.

https://claude.ai/code/session_01QJ4a61zNH95zyfJen4PzwB

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -342,6 +342,53 @@ class Cache(BaseCache):
                 self.save(call)
                 self.calls.evict(key)
 
+    def gc(self) -> set[Digest]:
+        """Evict value entries not reachable from any stored call.
+
+        Brute-force mark-and-sweep: walks every call record to build the set
+        of directly-referenced value digests, then transitively follows
+        destructured sub-references (via :meth:`DestructuringMixin.sub_digests`
+        when available), and evicts every ``values`` key outside the reachable
+        set.  Call records are left untouched.
+
+        Returns:
+            The set of digests that were evicted from value storage.
+        """
+        reachable: set[Digest] = set()
+        for key in self.calls.list():
+            try:
+                dc = self.calls.load(key)
+            except KeyError:
+                continue
+            if isinstance(dc.result, Digest):
+                reachable.add(dc.result)
+            for v in dc.arguments.values():
+                if isinstance(v, Digest):
+                    reachable.add(v)
+
+        sub_digests = getattr(self.values, "sub_digests", None)
+        if sub_digests is not None:
+            frontier = set(reachable)
+            while frontier:
+                key = frontier.pop()
+                try:
+                    children = sub_digests(key)
+                except KeyError:
+                    continue
+                new = children - reachable
+                reachable |= new
+                frontier |= new
+
+        evicted: set[Digest] = set()
+        for key in list(self.values.list()):
+            if key not in reachable:
+                try:
+                    self.values.evict(key)
+                    evicted.add(key)
+                except KeyError:
+                    continue
+        return evicted
+
 
 @dataclass(frozen=True)
 class ReadOnlyCache(BaseCache):
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -10,6 +10,7 @@ import pandas as pd
 from . import digest as _digest
 from .digest import Digest  # type hint convenience
 from . import storage
+from .storage.destructuring import HasChildDigests
 from .call import Call, DigestedCall, LazyCall, QueryCall
 from . import call
 from . import query
@@ -347,8 +348,9 @@ class Cache(BaseCache):
 
         Brute-force mark-and-sweep: walks every call record to build the set
         of directly-referenced value digests, then transitively follows
-        destructured sub-references (via :meth:`DestructuringMixin.sub_digests`
-        when available), and evicts every ``values`` key outside the reachable
+        destructured sub-references (via :meth:`DestructuringMixin.child_digests`
+        on storages that satisfy :class:`HasChildDigests`), and evicts every
+        ``values`` key outside the reachable
         set.  Call records are left untouched.
 
         Returns:
@@ -366,13 +368,12 @@ class Cache(BaseCache):
                 if isinstance(v, Digest):
                     reachable.add(v)
 
-        sub_digests = getattr(self.values, "sub_digests", None)
-        if sub_digests is not None:
+        if isinstance(self.values, HasChildDigests):
             frontier = set(reachable)
             while frontier:
                 key = frontier.pop()
                 try:
-                    children = sub_digests(key)
+                    children = self.values.child_digests(key)
                 except KeyError:
                     continue
                 new = children - reachable
--- a/tests/unit/caches/test_gc.py
+++ b/tests/unit/caches/test_gc.py
@@ -6,177 +6,136 @@ import pytest
 
 from fleche.call import Call
 from fleche.caches import Cache
-from fleche.digest import Digest, digest
-from fleche.storage import ValueVoid, CallVoid
-from fleche.storage.memory import CallMemory, ValueMemory
+from fleche.digest import digest
+from fleche.storage.base import ValueMixin
+from fleche.storage.memory import CallMemory, MemoryBackend, ValueMemory
 
 
-def _make_cache() -> Cache:
-    return Cache(values=ValueMemory({}), calls=CallMemory({}))
+@dataclass(frozen=True)
+class FlatValueMemory(ValueMixin, MemoryBackend):
+    """Memory-backed value storage without destructuring.
+
+    Stores values opaquely — no ``DestructuringMixin``, so it does not satisfy
+    :class:`fleche.storage.destructuring.HasChildDigests`.  Used to exercise
+    the GC path that walks only direct call references.
+    """
 
 
-def test_gc_empty_cache_is_noop():
-    c = _make_cache()
-    assert c.gc() == set()
+@pytest.fixture(params=["destructuring", "flat"])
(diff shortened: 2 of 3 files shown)


src/fleche/caches.py
```diff
@@ -342,6 +342,53 @@ def redigest(self) -> None:
...
+                dc = self.calls.load(key)
+            except KeyError:
+                continue
+            if isinstance(dc.result, Digest):
+                reachable.add(dc.result)
+            for v in dc.arguments.values():
+                if isinstance(v, Digest):
+                    reachable.add(v)
+
+        sub_digests = getattr(self.values, "sub_digests", None)
```
pmrv: @claude Awkward and API breaking.  Consider alternatives.

pmrv: **Claude finished @pmrv's task in 1m 35s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24904601867)

---
### Analysis of `getattr` probe at `caches.py:369`

- [x] Read `src/fleche/caches.py` around line 369
- [x] Read `src/fleche/storage/destructuring.py`
- [x] Read tests for context
- [x] Propose alternatives

---

The line in question:

```python
sub_digests = getattr(self.values, "sub_digests", None)
if sub_digests is not None:
```

**Why it's awkward:** `self.values` is typed as `ValueStorage`, which has no `sub_digests` declaration. The `getattr` probe is effectively runtime duck-typing — a magic string check that bypasses the type system, hides the contract, and confuses type checkers.

**Why it's API-adjacent:** `sub_digests` is a new public method added to `DestructuringMixin`. If it were added to `ValueStorage` as an abstract method (the obvious fix), every existing `ValueStorage` implementation would need updating — that *would* be breaking. The current `getattr` workaround avoids that breakage at the cost of clarity.

---

### Alternatives

#### Option A — `typing.Protocol` (recommended)

Define an optional-capability protocol in `storage/destructuring.py` (or `storage/base.py`):

```python
from typing import Protocol, runtime_checkable
from ..digest import Digest

@runtime_checkable
class SubDigestsProvider(Protocol):
    def sub_digests(self, key: Digest) -> set[Digest]: ...
```

Then in `Cache.gc()`:

```python
if isinstance(self.values, SubDigestsProvider):
    frontier = set(reachable)
    while frontier:
        key = frontier.pop()
        try:
            children = self.values.sub_digests(key)
        except KeyError:
            continue
        new = children - reachable
        reachable |= new
        frontier |= new
```

No `getattr`, no magic string, mypy understands the narrowing, and the interface is documented explicitly. No changes to `ValueStorage` → truly non-breaking.

---

#### Option B — default method on `ValueStorage`

Add a non-abstract default directly to the `ValueStorage` ABC (no existing implementation breaks because the default is provided):

```python
# in storage/base.py, on ValueStorage:
def sub_digests(self, key: Digest) -> set[Digest]:
    """Direct digest children of the stored entry.  Returns empty set for flat storages."""
    return set()
```

`DestructuringMixin` already has the real override. `Cache.gc()` then unconditionally calls `self.values.sub_digests(key)` — no branching needed at all, and the while-loop simplifies to:

```python
frontier = set(reachable)
while frontier:
    key = frontier.pop()
    try:
        children = self.values.sub_digests(key)
    except KeyError:
        continue
    new = children - reachable
    reachable |= new
    frontier |= new
```

This is the cleanest `gc()` body, but it does add a method to `ValueStorage`'s surface area — a minor API expansion (not breaking, since it has a default).

---

#### Option C — delegate entirely to value storage

Move the transitive-closure step into `ValueStorage` with a `reachable_from(roots: set[Digest]) -> set[Digest]` method. The base returns `roots` unchanged; `DestructuringMixin` does the BFS internally. `Cache.gc()` becomes:

```python
reachable = ...  # direct refs from call records
reachable = self.values.reachable_from(reachable)
```

Most encapsulated, but adds a slightly wider method to the ABC and means the GC traversal logic moves into storage rather than staying in the cache layer.

---

**My recommendation:** **Option A** — the `@runtime_checkable Protocol` approach. It eliminates the `getattr` without touching `ValueStorage`, keeps the interface explicit and type-checkable, and matches how Python's standard library expresses optional structural capabilities (e.g. `Sized`, `Iterable`). [Fix this →](https://claude.ai/code?q=In%20src%2Ffleche%2Fcaches.py%20on%20branch%20claude%2Fgarbage-collection-values-7wvkI%2C%20replace%20the%20getattr%20probe%20at%20line%20369%20with%20a%20%40runtime_checkable%20Protocol%20%28SubDigestsProvider%29%20defined%20in%20storage%2Fdestructuring.py%20and%20an%20isinstance%20check%20in%20Cache.gc%28%29.%20DestructuringMixin%20already%20has%20sub_digests%20so%20it%20will%20satisfy%20the%20protocol%20automatically.&repo=pmrv/fleche)

pmrv: @claude Go with A but name the method better.

pmrv: **Claude finished @pmrv's task in 8m 4s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24907593288)

---
### Implementing Option A with better method name

- [x] Read task and gather context
- [x] Read current code in `caches.py`, `storage/destructuring.py`, `storage/base.py`
- [x] Choose better method name: `child_digests` (clearer graph-traversal semantics than `sub_digests`)
- [x] Define `@runtime_checkable Protocol HasChildDigests` with `child_digests` in `destructuring.py`
- [x] Rename `DestructuringMixin.sub_digests` → `child_digests`
- [x] Update `Cache.gc()`: replace `getattr` probe with `isinstance(self.values, HasChildDigests)`
- [x] Update test comments
- [x] Run tests (886 passed)
- [x] Commit and push

---

Done. Changes in commit `3f85ab3`:

- **`storage/destructuring.py`**: Added `@runtime_checkable class HasChildDigests(Protocol)` with a `child_digests` method declaration. Renamed `DestructuringMixin.sub_digests` → `child_digests`. Any storage with `child_digests` satisfies the protocol automatically.

- **`caches.py`**: Imported `HasChildDigests`; replaced `getattr(self.values, "sub_digests", None)` with `isinstance(self.values, HasChildDigests)` and calls `self.values.child_digests(key)` directly — no magic string, mypy-visible narrowing.

- **`tests/unit/caches/test_gc.py`**: Updated two comments that referenced `sub_digests`.
 ·

tests/unit/caches/test_gc.py
```diff
@@ -0,0 +1,182 @@
...
+
+from fleche.call import Call
+from fleche.caches import Cache
+from fleche.digest import Digest, digest
+from fleche.storage import ValueVoid, CallVoid
+from fleche.storage.memory import CallMemory, ValueMemory
+
+
+def _make_cache() -> Cache:
+    return Cache(values=ValueMemory({}), calls=CallMemory({}))
```
pmrv: Reuse a fixture, if none found in global helpers, make one.

tests/unit/caches/test_gc.py
pmrv, line 1: Your tests implicitly always work on destructing storages,  also test non destructing ones.

tests/unit/caches/test_gc.py
```diff
@@ -0,0 +1,182 @@
...
+    c.gc()
+
+    assert c.contains(key_a)
+    assert c.contains(key_b)
+
+
+def test_gc_on_non_destructuring_value_storage():
+    """Value storages without child_digests() (e.g. ValueVoid) are handled gracefully."""
+
+    @dataclass(frozen=True)
```
pmrv: Make a new MemoryStorage like class here where you just drop the destructing

pmrv (commented):
fix issues then rebase on latest main.

tests/unit/caches/test_gc.py
pmrv: Switched to the existing `clean_cache` fixture from `tests/fixtures.py` for the destructuring-only tests. For tests that should exercise both storage flavours, added a parametrised `gc_cache` fixture in `test_gc.py` itself (since it needs the test-local `FlatValueMemory`). bd5c1da

---

tests/unit/caches/test_gc.py
pmrv: The five storage-agnostic tests (`test_gc_empty_cache_is_noop`, `test_gc_retains_values_referenced_by_calls`, `test_gc_evicts_orphan_value`, `test_gc_preserves_call_records`, `test_gc_returns_evicted_keys`) are now parametrised over `["destructuring", "flat"]` via `gc_cache`, so each runs once on `ValueMemory` (with `DestructuringMixin`) and once on the new `FlatValueMemory`. The three destructuring-specific tests stay on `clean_cache` since their assertions exercise sub-tree reachability that flat storage can't model. bd5c1da

---

tests/unit/caches/test_gc.py
pmrv: Done. Replaced the duck-typed `_Stub` with a real `FlatValueMemory(ValueMixin, MemoryBackend)` — same composition as `ValueMemory` but without `DestructuringMixin`, so it doesn't satisfy `HasChildDigests`. It feeds the new parametrised `gc_cache` fixture, which exercises the storage-agnostic GC tests on both flavours. bd5c1da

---

PR #425, opened 2026-04-27
Add support for attrs-decorated classes in digest and storage

pmrv:
## Summary
This PR extends fleche's digest and destructuring storage systems to support `attrs`-decorated classes (both `@attrs.define` and `@attr.s` styles) with feature parity to dataclasses.

## Key Changes

- **Digest support for attrs**: Added `is_attrs_instance()` and `attrs_field_items()` helper functions to detect and extract fields from attrs instances. The digest algorithm treats attrs classes identically to dataclasses with the same name and field layout, ensuring cache compatibility when swapping implementations.

- **Destructuring storage integration**: Extended `DigestedDataclass.sunder()` to handle attrs instances by decomposing them into per-field representations, enabling efficient nested storage and round-trip serialization.

- **Merkle-tree property**: Attrs instances support the same merkle-tree property as dataclasses—replacing a sub-value with its digest produces an identical overall digest.

- **@fleche decorator compatibility**: Attrs instances now work seamlessly as cache keys and return values in `@fleche`-decorated functions.

- **Test coverage**: Added comprehensive test suite covering:
  - Basic digest operations (stability, field distinction, class distinction)
  - Both new-style (`@attrs.define`) and old-style (`@attr.s`) attrs classes
  - Frozen and mutable variants
  - Nested attrs structures and collections
  - Parity with equivalent dataclass digests
  - Integration with `@fleche` caching
  - Destructuring storage round-trips

## Implementation Details

- The digest format for attrs classes mirrors dataclasses exactly (class name + field dict), ensuring identical hashes for equivalent structures regardless of implementation.
- Attrs support is gracefully degraded when the `attrs` library is not installed (returns False from `is_attrs_instance()`).
- The destructuring storage treats attrs instances as `DigestedDataclass` objects, using the same per-field decomposition strategy as dataclasses.

https://claude.ai/code/session_016qhy9egJNeWEEV7Pr6qRNA

### DIFF
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -11,7 +11,29 @@ from collections.abc import Iterable
 from typing import Any, TypeVar, Callable, Type, Generic
 import numpy as np
 
+try:
+    import attr as _attr
+except ImportError:
+    _attr = None
+
 logger = logging.getLogger("fleche.digest")
+
+
+def is_attrs_instance(value: Any) -> bool:
+    """Return True iff *value* is an instance of an ``attrs``-decorated class.
+
+    Returns False (rather than raising) when ``attrs`` is not installed.
+    """
+    return (
+        _attr is not None
+        and not isinstance(value, type)
+        and _attr.has(type(value))
+    )
+
+
+def attrs_field_items(value: Any) -> list[tuple[str, Any]]:
+    """Return ``[(name, value), ...]`` for every attribute of an attrs instance."""
+    return [(f.name, getattr(value, f.name)) for f in _attr.fields(type(value))]
 
 
 class Unhashable(Exception):
@@ -239,6 +261,10 @@ def _digest(value: Any) -> Digest:
                 lambda f: (f.name, getattr(value, f.name)), dataclasses.fields(value)
             )
             m.update(digest(dict(fields)).encode())
+        case _ if is_attrs_instance(value):
+            # mirror the dataclass digest format so an attrs class and a dataclass
+            # with the same name + field layout hash identically.
+            m.update(digest(dict(attrs_field_items(value))).encode())
         case Iterable():
             for v in value:
                 m.update(digest(v).encode())
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -12,7 +12,8 @@ from typing import Any, TypeVar, Callable, Type, Generic
 import numpy as np
 
 try:
-    import attr as _attr
+    import attr as _attr_module
+    _attr: types.ModuleType | None = _attr_module
 except ImportError:
     _attr = None
 
@@ -33,6 +34,7 @@ def is_attrs_instance(value: Any) -> bool:
 
 def attrs_field_items(value: Any) -> list[tuple[str, Any]]:
     """Return ``[(name, value), ...]`` for every attribute of an attrs instance."""
+    assert _attr is not None
     return [(f.name, getattr(value, f.name)) for f in _attr.fields(type(value))]
 
 
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -11,31 +11,9 @@ from collections.abc import Iterable
 from typing import Any, TypeVar, Callable, Type, Generic
 import numpy as np
 
-try:
-    import attr as _attr_module
-    _attr: types.ModuleType | None = _attr_module
-except ImportError:
-    _attr = None
+from . import _attrs
 
 logger = logging.getLogger("fleche.digest")
-
-
-def is_attrs_instance(value: Any) -> bool:
-    """Return True iff *value* is an instance of an ``attrs``-decorated class.
-
-    Returns False (rather than raising) when ``attrs`` is not installed.
-    """
-    return (
-        _attr is not None
-        and not isinstance(value, type)
-        and _attr.has(type(value))
-    )
-
-
-def attrs_field_items(value: Any) -> list[tuple[str, Any]]:
-    """Return ``[(name, value), ...]`` for every attribute of an attrs instance."""
-    assert _attr is not None
-    return [(f.name, getattr(value, f.name)) for f in _attr.fields(type(value))]
 
 
 class Unhashable(Exception):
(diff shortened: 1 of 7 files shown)


src/fleche/digest.py
```diff
@@ -11,9 +11,33 @@
...
+try:
+    import attr as _attr_module
+    _attr: types.ModuleType | None = _attr_module
+except ImportError:
+    _attr = None
+
 logger = logging.getLogger("fleche.digest")
 
 
+def is_attrs_instance(value: Any) -> bool:
```
pmrv: move to separate module

src/fleche/digest.py
```diff
@@ -239,6 +263,10 @@ def _digest(value: Any) -> Digest:
                 lambda f: (f.name, getattr(value, f.name)), dataclasses.fields(value)
             )
             m.update(digest(dict(fields)).encode())
+        case _ if is_attrs_instance(value):
+            # mirror the dataclass digest format so an attrs class and a dataclass
+            # with the same name + field layout hash identically.
+            m.update(digest(dict(attrs_field_items(value))).encode())
```
pmrv: Not sure about hash equivalence between data class and attrs. Maybe nice backwards compact feature but could turn out to be a foot gun.  Present detailed arguments for and against.

src/fleche/storage/destructuring.py
```diff
@@ -113,26 +113,31 @@ def mend(self, storage: 'DestructuringMixin'):
...
         # Fall back to whole-object storage on any structural problem.
-        try:
-            dc_fields = _dataclasses.fields(value)
-        except TypeError:
+        if _dataclasses.is_dataclass(value) and not isinstance(value, type):
+            try:
+                items = [(f.name, getattr(value, f.name)) for f in _dataclasses.fields(value)]
+            except (AttributeError, TypeError):
+                return value, float("inf")
+        elif digest.is_attrs_instance(value):
```
pmrv: Feels like you actually want separate suggested subclasses with this many cases inside sunder

pmrv (commented):
review

src/fleche/digest.py
pmrv: As requested, the trade-off — keeping in place pending your call:

**Where the equivalence shows up.** A `@dataclass` and an `@attrs.define` class with the same `__name__` and same field layout currently produce the same digest. This is because both go through `SHA256(type.__name__ + digest({field: value}))`, and the marker subclasses `DigestedDataclass` / `DigestedAttrs` follow suit.

**For:**
1. *Migration friendliness.* If a project converts a `@dataclass` to `@attrs.define` (or vice versa), every existing cache entry keyed by an instance survives the swap. fleche is a *persistent* cache and stable keys across refactors are one of the primary value props.
2. *Both are record types.* The digest is content-based: it hashes attribute names and values, not `__init__`/`__eq__`/`__hash__`. attrs and dataclasses deliver the same record contract, so by the digest's own lens, identical records have identical digests. Symmetric with how `digest(1) == digest(1.0) == digest(1+0j)` already.
3. *Future flexibility is easy either way.* The `DigestedFields`/`DigestedDataclass`/`DigestedAttrs` split (now landed) means the per-type `__digest__` could diverge with a one-liner if we ever decided to.

**Against:**
1. *Foot-gun in mixed projects.* If both forms of the same class name coexist (e.g. mid-migration), they would silently share cache slots. This *can* matter if the two forms compute differently — e.g. an attrs class with validators may reject inputs the dataclass form accepts, so a result cached under "dataclass-like attributes" might not match what the attrs form would have produced.
2. *Loss of class identity.* Users may assume "different class → different cache entry"; we silently break that for what is, from their perspective, two different declarations.
3. *Different runtime semantics.* slots, frozen, converters, validators, attrs's eq-key/order-key — none of these affect the digest. Equivalence under digest doesn't imply equivalence under the language. A user reading from cache could get an instance constructed via `object.__new__` that bypasses the validator, which is already true for dataclasses but at least with dataclasses the cache key was scoped by the *dataclass* lookup.
4. *Pre-existing class-name-only quirk.* Two dataclasses in different modules with the same name *already* collide today. Extending that to attrs widens the foot-gun, even if it doesn't introduce a new category.

**Mitigation if you want them distinct.** One-line change: in `_digest`'s attrs case, mix in a discriminator (e.g. `m.update(b"attrs")`) before the field map, and have `DigestedAttrs.__digest__` do the same. Keeps Merkle-transparency, breaks accidental cross-kind collisions. Optional: also mix in `type(value).__module__` to fix the dataclass-name-only quirk while you're there — though that one *is* a behaviour change for existing dataclass entries.

**My lean.** Keep the equivalence and document it explicitly — migration friendliness is concrete, the foot-gun is rare and bounded to mixed-form projects, and the existing dataclass-name-only collision behaviour is the bigger fish. Happy to flip it if you'd rather err on the side of stricter scoping.

---

src/fleche/storage/destructuring.py
```diff
@@ -110,30 +113,47 @@ def mend(self, storage: 'DestructuringMixin'):
             object.__setattr__(obj, name, self.get(storage, val_or_digest))
         return obj
 
+    @staticmethod
+    def _field_items(value: Any) -> list[tuple[str, Any]]:
+        raise NotImplementedError
```
pmrv: abstractmethod instead

src/fleche/storage/destructuring.py
```diff
@@ -110,30 +113,47 @@ def mend(self, storage: 'DestructuringMixin'):
...
+        if not items:
             return value, 0
 
-        try:
-            field_values = [getattr(value, f.name) for f in dc_fields]
-        except AttributeError:
-            return value, float("inf")
-
+        names = [n for n, _ in items]
+        field_values = [v for _, v in items]
```
pmrv: Could be a zip

src/fleche/digest.py
```diff
@@ -239,6 +241,10 @@ def _digest(value: Any) -> Digest:
                 lambda f: (f.name, getattr(value, f.name)), dataclasses.fields(value)
             )
             m.update(digest(dict(fields)).encode())
+        case _ if _attrs.is_attrs_instance(value):
+            # mirror the dataclass digest format so an attrs class and a dataclass
```
pmrv, line 245: I am in favor of keeping this equivalence, but we need to clearly document this.  Write a new doc page on digest equivalence.  Should be structured like a general intro to the digest function, but only fill in the part of dataclasses/attrs classes for now.

pmrv (commented):
review

PR #431, opened 2026-04-27
Cache per-function statics in Call.from_call (-27% on hit path)

pmrv:
## Summary

cProfile of the cache-hit hot path (50k calls of a `@fleche` wrapped
`add(x, y)`, Memory backend) showed `Call.from_call` accounting for **9.1s
of 24.6s (37%)**.  Three pure-of-`func` operations were being recomputed
on every invocation:

| Operation | Time/50k calls | % of hit path |
|---|---|---|
| `inspect.signature(func)` | 1.66 s | 6.7 % |
| `digest(func.__code__)` recursion | baked into 12.6 s of `digest()` | — |
| `VersionInfo.of(func)` / `get_module` | 0.95 s | 3.8 % |

This PR wraps `signature()` and `digest(func.__code__)` in
`lru_cache(maxsize=None)` helpers keyed on `func`.  Both are constants
of the function definition, so subsequent invocations are a dict
lookup.

`_extract_version_info()` is **left uncached** because
`tests/unit/fleche/test_fleche.py::test_fleche_with_module` deliberately
mutates `mock_function.__module__` between two `@fleche` wraps and
expects each to produce a distinct cache key — that case relies on
the version-info path running fresh.

## Numbers

50k-call cache-hit benchmark (in-process, no I/O):

| | Before | After | Δ |
|---|---|---|---|
| total wallclock | 24.6 s | 18.0 s | **−27 %** |
| `Call.from_call` cumulative | 9.1 s | 1.9 s | **5×** |

Standard `benchmark_integration.py` deltas (Memory backend, median of 3):

| workload / fn | before | after | Δ |
|---|---|---|---|
| `lightweight` / `contains_hit`  |  97 µs |  54 µs | **−44 %** |
| `lightweight` / `hit`           | 110 µs |  89 µs | **−19 %** |
| `compute_heavy` / `contains_hit`|  84 µs |  54 µs | **−37 %** |
| `compute_heavy` / `hit`         | 120 µs |  89 µs | **−26 %** |

## Caveats

`_func_signature` / `_func_code_digest` are keyed by `func` identity, so
modules that hot-swap `func.__code__` or `func.__signature__` after first
use will see stale values.  The two cached helpers expose
`.cache_clear()` for test suites that need to reset.  All 885 unit +
regression tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BvAnFkvRZvJRTrQbZk4aH1)_

### DIFF
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -1,7 +1,8 @@
 import logging
 from dataclasses import dataclass, field, replace
+from functools import lru_cache
 from typing import Any, Callable
-from inspect import signature
+from inspect import Signature, signature
 from collections.abc import Mapping
 
 from pyiron_snippets.versions import VersionInfo, get_module, get_qualname
@@ -10,6 +11,34 @@ from . import digest
 from .digest import Digest
 
 logger = logging.getLogger("fleche.call")
+
+
+@lru_cache(maxsize=None)
+def _func_signature(func) -> Signature:
+    """Memoised :func:`inspect.signature` lookup.
+
+    ``signature()`` is a non-trivial introspection call (~30 µs) and is
+    invoked on every cache hit by :meth:`Call.from_call` /
+    :class:`ArgumentPolicy`.  Caching keyed on ``func`` removes ~6% from the
+    cache-hit hot path.  Tests that mutate ``func.__signature__`` can call
+    ``_func_signature.cache_clear()`` to drop the cache.
+    """
+    return signature(func)
+
+
+@lru_cache(maxsize=None)
+def _func_code_digest(func) -> Digest | None:
+    """Memoised digest of ``func.__code__``.
+
+    ``digest(func.__code__)`` recursively hashes the code object's bytecode,
+    constants, names, etc., which is the single biggest contributor to
+    :meth:`Call.from_call` time on the cache-hit path.  Cached by ``func``
+    identity; modules that hot-swap ``__code__`` should call
+    ``_func_code_digest.cache_clear()``.
+    """
+    if not hasattr(func, "__code__"):
+        return None
+    return digest.digest(func.__code__)
 
 
 def _extract_version_info(func) -> tuple[str, str, str | int | None]:
@@ -46,7 +75,7 @@ def bind(func, args, kwargs, apply_defaults=False, partial=False):
         :attr:`inspect.BoundArguments.arguments` — an ``OrderedDict``
         containing the supplied (and, when requested, defaulted) values.
     """
-    sig = signature(func)
+    sig = _func_signature(func)
     if partial:
         bound = sig.bind_partial(*args, **kwargs)
     else:
@@ -75,12 +104,13 @@ class Call:
 
     @classmethod
     def from_call(cls, func, *args, **kwargs):
-        arguments = dict(bind(func, args, kwargs, apply_defaults=True))
+        sig = _func_signature(func)
+        bound = sig.bind(*args, **kwargs)
+        bound.apply_defaults()
         qualname, module, version = _extract_version_info(func)
-        call = cls(qualname, arguments, module=module)
+        call = cls(qualname, dict(bound.arguments), module=module)
         call.version = getattr(func, "__version__", version)
-        if hasattr(func, "__code__"):
-            call.code_digest = digest.digest(func.__code__)
+        call.code_digest = _func_code_digest(func)
         return call
 
     def to_lookup_key(self) -> "Digest":
@@ -315,9 +345,10 @@ class QueryCall:
 
     @classmethod
     def from_call(cls, func, *args, **kwargs):
+        sig = _func_signature(func)
         bound_args = bind(func, args, kwargs, partial=True)
         # Unspecified arguments default to None (wildcard)
-        arguments = {name: bound_args.get(name) for name in signature(func).parameters}
+        arguments = {name: bound_args.get(name) for name in sig.parameters}
         qualname, module, version = _extract_version_info(func)
         call = cls(qualname, arguments, module=module)
         call.version = getattr(func, "__version__", version)
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -30,6 +30,29 @@ def _cached_version_info(func) -> tuple[str, str, str | int | None]:
     return _extract_version_info(func)
 
 
+def _func_statics(func) -> tuple[Signature, str, str, str | int | None, Digest | None]:
+    """Return ``(signature, qualname, module, version, code_digest)`` for *func*.
+
+    Uses the per-function ``lru_cache`` helpers when *func* is hashable; falls
+    back to direct introspection in a single ``except`` for callables with
+    ``__hash__ = None``.  Both :meth:`Call.from_call` and
+    :meth:`QueryCall.from_call` consume this tuple (the latter discards
+    ``code_digest``).
+    """
+    try:
+        return (
+            _cached_signature(func),
+            *_cached_version_info(func),
+            _cached_code_digest(func),
+        )
+    except TypeError:
+        return (
+            signature(func),
+            *_extract_version_info(func),
+            digest.digest(func.__code__) if hasattr(func, "__code__") else None,
+        )
+
+
 def _extract_version_info(func) -> tuple[str, str, str | int | None]:
     """Extract ``(name, module, version)`` from ``func`` via :mod:`pyiron_snippets.versions`.
 
@@ -97,18 +120,7 @@ class Call:
 
     @classmethod
     def from_call(cls, func, *args, **kwargs):
-        try:
-            sig = _cached_signature(func)
-            qualname, module, version = _cached_version_info(func)
-            code_digest = _cached_code_digest(func)
-        except TypeError:
-            # Unhashable callable (e.g. instance with __hash__ = None) —
-            # bypass the per-function caches and recompute directly.
-            sig = signature(func)
-            qualname, module, version = _extract_version_info(func)
-            code_digest = (
-                digest.digest(func.__code__) if hasattr(func, "__code__") else None
-            )
+        sig, qualname, module, version, code_digest = _func_statics(func)
         bound = sig.bind(*args, **kwargs)
         bound.apply_defaults()
         call = cls(qualname, dict(bound.arguments), module=module)
@@ -348,13 +360,7 @@ class QueryCall:
 
     @classmethod
     def from_call(cls, func, *args, **kwargs):
-        try:
-            sig = _cached_signature(func)
-            qualname, module, version = _cached_version_info(func)
-        except TypeError:
-            # Unhashable callable (e.g. instance with __hash__ = None).
-            sig = signature(func)
-            qualname, module, version = _extract_version_info(func)
+        sig, qualname, module, version, _code_digest = _func_statics(func)
         bound_args = bind(func, args, kwargs, partial=True)
         # Unspecified arguments default to None (wildcard)
         arguments = {name: bound_args.get(name) for name in sig.parameters}
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -18,19 +18,6 @@ def _cached_signature(func) -> Signature:
     return signature(func)
 
 
-def _func_signature(func) -> Signature:
-    """Cached :func:`inspect.signature` lookup with a fallback for
-    callables that cannot be hashed (e.g. instances with
-    ``__hash__ = None``).  Caches the per-function statics that
-    :meth:`Call.from_call` and :class:`ArgumentPolicy.check_required`
-    would otherwise re-introspect on every cache hit.
-    """
-    try:
-        return _cached_signature(func)
-    except TypeError:
-        return signature(func)
-
-
 @lru_cache(maxsize=1000)
 def _cached_code_digest(func) -> Digest | None:
     if not hasattr(func, "__code__"):
@@ -38,35 +25,12 @@ def _cached_code_digest(func) -> Digest | None:
     return digest.digest(func.__code__)
 
 
-def _func_code_digest(func) -> Digest | None:
-    """Cached :func:`fleche.digest.digest` of ``func.__code__``, with a
-    fallback for unhashable callables.  Avoids re-running the recursive
-    code-object digest on every cache-hit invocation.
-    """
-    try:
-        return _cached_code_digest(func)
-    except TypeError:
-        if not hasattr(func, "__code__"):
-            return None
-        return digest.digest(func.__code__)
-
-
 @lru_cache(maxsize=1000)
 def _cached_version_info(func) -> tuple[str, str, str | int | None]:
-    return _compute_version_info(func)
+    return _extract_version_info(func)
 
 
 def _extract_version_info(func) -> tuple[str, str, str | int | None]:
-    """Cached ``(name, module, version)`` lookup, with a fallback for
-    unhashable callables.  Wraps :func:`_compute_version_info`.
-    """
-    try:
-        return _cached_version_info(func)
-    except TypeError:
-        return _compute_version_info(func)
-
-
-def _compute_version_info(func) -> tuple[str, str, str | int | None]:
     """Extract ``(name, module, version)`` from ``func`` via :mod:`pyiron_snippets.versions`.
 
     Uses :meth:`VersionInfo.of` to introspect ``func``.  When the module is not
@@ -100,7 +64,11 @@ def bind(func, args, kwargs, apply_defaults=False, partial=False):
         :attr:`inspect.BoundArguments.arguments` — an ``OrderedDict``
         containing the supplied (and, when requested, defaulted) values.
     """
-    sig = _func_signature(func)
+    try:
+        sig = _cached_signature(func)
+    except TypeError:
+        # Unhashable callable (e.g. instance with __hash__ = None).
+        sig = signature(func)
     if partial:
         bound = sig.bind_partial(*args, **kwargs)
     else:
@@ -129,13 +97,23 @@ class Call:
 
     @classmethod
     def from_call(cls, func, *args, **kwargs):
-        sig = _func_signature(func)
+        try:
+            sig = _cached_signature(func)
+            qualname, module, version = _cached_version_info(func)
+            code_digest = _cached_code_digest(func)
+        except TypeError:
+            # Unhashable callable (e.g. instance with __hash__ = None) —
+            # bypass the per-function caches and recompute directly.
+            sig = signature(func)
+            qualname, module, version = _extract_version_info(func)
+            code_digest = (
+                digest.digest(func.__code__) if hasattr(func, "__code__") else None
+            )
         bound = sig.bind(*args, **kwargs)
         bound.apply_defaults()
-        qualname, module, version = _extract_version_info(func)
         call = cls(qualname, dict(bound.arguments), module=module)
         call.version = getattr(func, "__version__", version)
-        call.code_digest = _func_code_digest(func)
+        call.code_digest = code_digest
         return call
 
     def to_lookup_key(self) -> "Digest":
@@ -370,11 +348,16 @@ class QueryCall:
 
     @classmethod
     def from_call(cls, func, *args, **kwargs):
-        sig = _func_signature(func)
+        try:
+            sig = _cached_signature(func)
+            qualname, module, version = _cached_version_info(func)
+        except TypeError:
+            # Unhashable callable (e.g. instance with __hash__ = None).
+            sig = signature(func)
+            qualname, module, version = _extract_version_info(func)
         bound_args = bind(func, args, kwargs, partial=True)
         # Unspecified arguments default to None (wildcard)
         arguments = {name: bound_args.get(name) for name in sig.parameters}
-        qualname, module, version = _extract_version_info(func)
         call = cls(qualname, arguments, module=module)
         call.version = getattr(func, "__version__", version)
         return call
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -13,35 +13,60 @@ from .digest import Digest
 logger = logging.getLogger("fleche.call")
 
 
-@lru_cache(maxsize=None)
-def _func_signature(func) -> Signature:
-    """Memoised :func:`inspect.signature` lookup.
-
-    ``signature()`` is a non-trivial introspection call (~30 µs) and is
-    invoked on every cache hit by :meth:`Call.from_call` /
-    :class:`ArgumentPolicy`.  Caching keyed on ``func`` removes ~6% from the
-    cache-hit hot path.  Tests that mutate ``func.__signature__`` can call
-    ``_func_signature.cache_clear()`` to drop the cache.
-    """
+@lru_cache(maxsize=1000)
+def _cached_signature(func) -> Signature:
     return signature(func)
 
 
-@lru_cache(maxsize=None)
-def _func_code_digest(func) -> Digest | None:
-    """Memoised digest of ``func.__code__``.
-
-    ``digest(func.__code__)`` recursively hashes the code object's bytecode,
-    constants, names, etc., which is the single biggest contributor to
-    :meth:`Call.from_call` time on the cache-hit path.  Cached by ``func``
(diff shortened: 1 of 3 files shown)


pmrv:
## Top-5 perf-issues triage — context for this PR

This PR addresses **issue #4** of a five-item triage I did against
`benchmark_digest.py` and `benchmark_integration.py`.  The remaining
four are not fixed here — I'm leaving them as separate comments on this
PR with cProfile evidence + a proposed fix each, so they can be picked
up individually.

The five hotspots (ordered by impact on the cache-hit / digest paths):

1. **Memory backend `copy.deepcopy` on every `get`** — ~24 % of hit-path time
2. **Cache-hit path re-runs `signature` / `VersionInfo.of` / code digest** — *fixed in this PR* (~10 % of hit path)
3. **SQL `save` does `SELECT` + two `COMMIT`s per call record** — ~50 % of `Sql.save`
4. **Iterable/dict digest: hex-encode round-trip per child** — ~12 % of long-iterable digests
5. **`hasattr(value, "__digest__")` probed on every recursive `_digest`** — ~2 % across the board

Each is detailed in its own comment below with the exact cProfile excerpt
and a proposed fix.  The numbers come from
`python -m cProfile` on a pristine main checkout (commit `722f4a3c`)
running representative workloads — the same workloads each benchmark
script already exercises.

— *Claude*

---

pmrv:
## Issue 1 — `ValueMemory.get` deep-copies every load (~24 % of hit path)

### Where

`src/fleche/storage/memory.py:25` — `MemoryBackend.get` (and the dict it
backs) returns `copy.deepcopy(self.data[key])` to defend against callers
mutating cached values.

### Evidence

cProfile, `for _ in range(50_000): f(1, 2)` on a pre-warmed Memory cache
(`@fleche def f(x, y): return x + y`):

```
   ncalls  tottime  cumtime  filename:lineno(function)
    50000    0.379   24.800  src/fleche/wrapper.py:257(wrapper)
    50000    0.061    6.536  src/fleche/caches.py:267(load)
    50000    0.148    6.301  src/fleche/storage/base.py:247(load)
   100000    0.096    5.946  src/fleche/storage/memory.py:25(get)
1850000/100000 2.679    5.850 /usr/lib/python3.11/copy.py:128(deepcopy)
250000/50000   0.637    5.374 /usr/lib/python3.11/copy.py:259(_reconstruct)
150000/50000   0.570    4.811 /usr/lib/python3.11/copy.py:227(_deepcopy_dict)
```

`deepcopy` accounts for **5.85 s of 24.6 s (23.7 %)** on the hit path — a
hot-path tax most users never opt into deliberately.  Fixing this alone
would push the Memory `hit` benchmark from ~110 µs into the ~80 µs range.

### Proposed fix

Make the defensive copy opt-in:

```python
@dataclass(frozen=True)
class MemoryBackend(StorageBackend):
    data: dict = field(default_factory=dict)
    deep_copy: bool = False  # default: hand back the stored reference

    def get(self, key):
        v = self.data[key]
        return copy.deepcopy(v) if self.deep_copy else v
```

Most cached values are content-addressed by a hash of their bytes, so
mutating one effectively invalidates the cache key that addresses it —
the value either round-trips through `pickle` (file backends) or is
rebuilt via destructuring (`DigestedDataclass.mend` etc.).  A user who
genuinely caches a mutable container they then plan to mutate can pass
`deep_copy=True`.

Drawbacks:
- Default behaviour change.  Anyone relying on "Memory = isolated
  copies" semantics will see their code start sharing references.  This
  could be guarded behind `ValueMemory(safe=True)` as default and a
  documented `safe=False` fast-path.
- An alternative that doesn't change defaults: skip `deepcopy` when
  `type(v)` is in a known-immutable allow-list (`str | int | float |
  bytes | tuple | frozenset | None`).  Keeps the safety guarantee and
  still buys most of the win for typical small return values.

### Risk

Medium — it's a public-facing semantic change.  Worth a discussion
before implementing.

---

pmrv:
## Issue 2 — Cache-hit path re-runs per-function statics (THIS PR)

### Where

`src/fleche/call.py:76` — `Call.from_call` recomputes
`inspect.signature(func)`, `_extract_version_info(func)` and
`digest(func.__code__)` on every invocation, even though all three are
constants of `func`.

### Evidence (before this PR)

Same 50k-call hit benchmark:

```
   ncalls  tottime  cumtime  filename:lineno(function)
    50000    0.214    9.125  src/fleche/call.py:76(from_call)
    50000    0.101    7.725  src/fleche/call.py:86(to_lookup_key)
    50000    0.037    1.662  /usr/lib/python3.11/inspect.py:3261(signature)
    50000    0.040    1.625  /usr/lib/python3.11/inspect.py:3007(from_callable)
    50000    0.260    1.586  /usr/lib/python3.11/inspect.py:2435(_signature_from_callable)
    50000    0.066    0.945  src/fleche/call.py:15(_extract_version_info)
    50000    0.130    0.872  pyiron_snippets/versions.py:54(of)
```

`from_call` cumulatively burns **9.13 s of 24.64 s (37 %)**, of which
about 1.66 s is `inspect.signature`, 0.94 s is `VersionInfo.of`, and the
remainder is the recursive code-object digest threaded through `digest`.

### Fix (implemented here)

Two `lru_cache(maxsize=None)` helpers — `_func_signature(func)` and
`_func_code_digest(func)` — both keyed on `func` identity.
`_extract_version_info` is left uncached because
`tests/unit/fleche/test_fleche.py::test_fleche_with_module` relies on
re-extraction picking up `func.__module__` mutations.

### After-fix profile (same workload)

```
   ncalls  tottime  cumtime  filename:lineno(function)
    50000    0.197    1.899  src/fleche/call.py:105(from_call)
    50000    0.063    0.916  src/fleche/call.py:44(_extract_version_info)
   (no inspect.signature / from_callable in top 20)
```

`from_call` cumulative drops 9.1 s → 1.9 s (**5×**).  Total hit-path
wallclock 24.6 s → 18.0 s (**−27 %**).

### Risk

Low — caches are keyed on `func` identity, opt-out via `.cache_clear()`.
Tests pass green; semantic change limited to functions that hot-swap
`__code__` or `__signature__` after first wrap.

---

pmrv:
## Issue 3 — `Sql.save` does a `SELECT` + two `COMMIT`s per call (~50 % of `Sql.save`)

### Where

`src/fleche/storage/sql.py:282` — `Sql.save` calls
`self._contains(key)` first (line 246, runs a `SELECT … FROM calls
WHERE digest = ?` via `Session.get`), then inserts the call row and the
arguments rows in two separate flushes/commits.

### Evidence

cProfile, 499 in-memory `Sql.save(DigestedCall)` operations on a
`sqlite:///:memory:`:

```
   ncalls  tottime  cumtime  filename:lineno(function)
      499    0.004    1.362  src/fleche/storage/sql.py:282(save)
      499    0.010    0.955  src/fleche/storage/sql.py:181(put)
  998/499    0.008    0.614  sqlalchemy/orm/session.py:1304(commit)
     1497    0.005    0.573  sqlalchemy/orm/session.py:4301(flush)
      499    0.002    0.256  src/fleche/storage/base.py:97(contains)
      499    0.006    0.239  src/fleche/storage/sql.py:246(_contains)
      499    0.001    0.224  sqlalchemy/orm/session.py:3569(get)
```

Two key tells:

1. `commit` is called **998 times for 499 saves** — each save does two
   commits (call row, then arguments rows).  Each commit on a real
   filesystem-backed SQLite forces a WAL fsync, but even on
   `:memory:` the per-commit overhead totals 0.61 s (45 % of save time).
2. `_contains` is invoked **once per save** — a redundant `SELECT`
   immediately before an `INSERT`.  That's another 0.24 s (18 %).

Together: **~63 % of `Sql.save` time is spent on the pre-check + extra
commit**, not on the actual write.

### Proposed fix

1. Replace the `_contains` pre-check with `INSERT OR IGNORE` (SQLite) /
   `ON CONFLICT (digest) DO NOTHING` (PostgreSQL).  SQLAlchemy:
   ```python
   from sqlalchemy.dialects.sqlite import insert as sqlite_insert
   stmt = sqlite_insert(CallTable).values(...).on_conflict_do_nothing(
       index_elements=["digest"]
   )
   session.execute(stmt)
   ```
   Saves a round-trip per save.  Same digest collisions are silently
   skipped (which is what `_contains`-and-skip does today).

2. Coalesce the call-row insert and the arguments-row inserts into a
   single transaction with one `commit`.  `Session.add`/`add_all` then
   one `commit` at the end of `save` is sufficient — SQLAlchemy will
   flush in dependency order on commit.

Expected impact: ~50 % reduction in `Sql.save` time.  In the integration
benchmark, `Pickle+Sql/lightweight/miss` is 19 ms today; this should
roughly halve the SQL-side contribution.

### Risk

Low – `ON CONFLICT DO NOTHING` matches the existing
"contains-then-skip" semantics exactly.  Single-commit-per-save is a
durability change for users who were relying on the call-row landing
before the arguments rows; that's a niche case but worth a sanity check
in the regression tests for `tests/regression/test_sql_concurrent_save.py`.

---

pmrv:
## Issue 4 — Iterable / dict digest hex-encodes every child (~12 % of `digest()` for big iterables)

### Where

`src/fleche/digest.py:243` (Iterable case) and `:204-205` (dict case):

```python
case dict():
    sorted_items = sorted(...)
    for k_digest, k, v in sorted_items:
        m.update(k_digest.encode())          # ← 64 hex chars → 64 utf-8 bytes
        m.update(digest(v).encode())         # ← same again

case Iterable():
    for v in value:
        m.update(digest(v).encode())         # ← every recursion eats this
```

Each child digest is computed as 32 raw bytes, formatted to 64 hex
chars, returned as a `Digest` (str), then re-encoded to 64 utf-8 bytes
to feed back into the parent SHA-256.  The hex round-trip is pure waste.

### Evidence

cProfile, `digest(generate_examples(st.lists(st.integers(), min_size=100)))`
(50 outer calls; 946k recursive `_digest` calls):

```
   ncalls  tottime  cumtime  filename:lineno(function)
946900/50    2.869    4.796  src/fleche/digest.py:133(_digest)
  2835600    0.394    0.394  {method 'update' of '_hashlib.HASH' objects}
   946900    0.386    0.386  {method 'hexdigest' of '_hashlib.HASH' objects}
  1893750    0.205    0.205  {method 'encode' of 'str' objects}
   946900    0.203    0.203  {built-in method _hashlib.openssl_sha256}
```

`hexdigest` (0.39 s) + `str.encode` of those hex strings (~0.10 s of the
0.21 s total) ≈ **0.49 s of 4.80 s = 10 %** spent translating between hex
strings and the bytes the parent hash actually consumes.

### Proposed fix

Introduce an internal `_digest_bytes(value) -> bytes` returning the raw
32-byte SHA-256, and rewrite the recursive cases to use it:

```python
def _digest_bytes(value):
    m = hashlib.sha256()
    ...
    case Iterable():
        for v in value:
            m.update(_digest_bytes(v))       # 32 raw bytes, no hex
    ...
    return m.digest()                        # raw bytes

def digest(value):
    return Digest(_digest_bytes(value).hex())
```

Saves the hex format + encode on every recursive call and halves the
bytes fed into each parent hash.

### Caveat

**This changes the wire format of every digest.**  All persisted caches
become unreadable.  Has to be coordinated with a `hash_version` bump
and probably a one-shot migration via `Cache.redigest`.  Worth doing,
but only as a deliberate, documented release boundary — not a stealth
optimisation.

### Lower-risk alternative

Cache the bytes form alongside the hex form:

```python
class Digest(str):
    @property
    def _bytes(self):
        try: return self.__bytes_cache
        except AttributeError:
            self.__bytes_cache = bytes.fromhex(self)
            return self.__bytes_cache
```

Then `m.update(digest(v)._bytes)` instead of `digest(v).encode()`.  The
hex string is still the wire format (no breakage), but the inner hash
sees half the bytes and the encode call disappears.  Saves ~half the
hex-conversion cost — call it 5–6 % on this workload — without any
backwards-compat issue.

### Risk

- Format change (raw bytes path): **High** — invalidates every persisted
  cache.  Needs `hash_version` discipline.
- Bytes-cache approach: **Low** — `Digest` is a `str` subclass already;
  adding a lazy bytes-cache attribute is invisible to users.

---

pmrv:
## Issue 5 — `hasattr(value, "__digest__")` probed on every recursive `_digest` (~2 %)

### Where

`src/fleche/digest.py:160`:

```python
match value:
    case Digest():
        return value
    case _ if hasattr(value, "__digest__"):
        return Digest(value.__digest__())
    case str(): ...
    case int(): ...
```

The `__digest__` opt-in protocol is checked *before* the cheap built-in
cases.  For every `int`, `str`, `float`, `dict`, `list` (the bulk of
real-world digest traffic), Python performs an attribute lookup that
must walk the type chain to fail.

### Evidence

cProfile, same 946k-`_digest`-call workload:

```
   ncalls  tottime   filename:lineno(function)
   951950    0.101   {built-in method builtins.hasattr}
```

That's **0.10 s of 4.80 s = 2.1 %** spent on a check that returns
`False` for every built-in.  Small in relative terms, but it's also the
single most-called branch in the digest hot loop.

### Proposed fixes (cheapest first)

**Option A** — reorder the `match` so built-ins are checked first:

```python
match value:
    case Digest():
        return value
    case str():       ...        # ← cheap, common, return early
    case int():       ...
    case bool():      ...
    case None:        ...
    case dict():      ...
    case _ if hasattr(value, "__digest__"):
        return Digest(value.__digest__())
    case Iterable():  ...
```

Cost: zero.  Behaviour change: tiny — a `dict`/`list` subclass that
defines `__digest__` would currently use `__digest__` and after this
change would use the dict/list path instead.  Whether that's a feature
or a bug depends on intent; would want a quick poll of users defining
`__digest__` on container subclasses.

**Option B** — type-cache the answer:

```python
_HAS_DIGEST_CACHE: dict[type, bool] = {}

def _has_digest_dunder(t):
    cached = _HAS_DIGEST_CACHE.get(t)
    if cached is None:
        cached = hasattr(t, "__digest__")
        _HAS_DIGEST_CACHE[t] = cached
    return cached
```

`hasattr` is replaced with a dict lookup keyed on `type(value)`.  No
semantic change.  ~30 ns instead of ~100 ns per call.

### Risk

- Option A: low–medium, depending on whether anyone relies on
  `__digest__` overriding container behaviour on subclasses.
- Option B: low — pure caching, identical semantics.

Probably worth combining with **Issue 4**'s `_digest_bytes` refactor
since both touch the same dispatch hot loop — the rearrangement is
cheaper to land as one diff with one round of testing than two.

---

pmrv:
> ## Issue 3 — `Sql.save` does a `SELECT` + two `COMMIT`s per call (~50 % of `Sql.save`)
> 
> ### Where
> 
> `src/fleche/storage/sql.py:282` — `Sql.save` calls
> `self._contains(key)` first (line 246, runs a `SELECT … FROM calls
> WHERE digest = ?` via `Session.get`), then inserts the call row and the
> arguments rows in two separate flushes/commits.
> 
> ### Evidence
> 
> cProfile, 499 in-memory `Sql.save(DigestedCall)` operations on a
> `sqlite:///:memory:`:
> 
> ```
>    ncalls  tottime  cumtime  filename:lineno(function)
>       499    0.004    1.362  src/fleche/storage/sql.py:282(save)
>       499    0.010    0.955  src/fleche/storage/sql.py:181(put)
>   998/499    0.008    0.614  sqlalchemy/orm/session.py:1304(commit)
>      1497    0.005    0.573  sqlalchemy/orm/session.py:4301(flush)
>       499    0.002    0.256  src/fleche/storage/base.py:97(contains)
>       499    0.006    0.239  src/fleche/storage/sql.py:246(_contains)
>       499    0.001    0.224  sqlalchemy/orm/session.py:3569(get)
> ```
> 
> Two key tells:
> 
> 1. `commit` is called **998 times for 499 saves** — each save does two
>    commits (call row, then arguments rows).  Each commit on a real
>    filesystem-backed SQLite forces a WAL fsync, but even on
>    `:memory:` the per-commit overhead totals 0.61 s (45 % of save time).
> 2. `_contains` is invoked **once per save** — a redundant `SELECT`
>    immediately before an `INSERT`.  That's another 0.24 s (18 %).
> 
> Together: **~63 % of `Sql.save` time is spent on the pre-check + extra
> commit**, not on the actual write.
> 
> ### Proposed fix
> 
> 1. Replace the `_contains` pre-check with `INSERT OR IGNORE` (SQLite) /
>    `ON CONFLICT (digest) DO NOTHING` (PostgreSQL).  SQLAlchemy:
>    ```python
>    from sqlalchemy.dialects.sqlite import insert as sqlite_insert
>    stmt = sqlite_insert(CallTable).values(...).on_conflict_do_nothing(
>        index_elements=["digest"]
>    )
>    session.execute(stmt)
>    ```
>    Saves a round-trip per save.  Same digest collisions are silently
>    skipped (which is what `_contains`-and-skip does today).
> 
> 2. Coalesce the call-row insert and the arguments-row inserts into a
>    single transaction with one `commit`.  `Session.add`/`add_all` then
>    one `commit` at the end of `save` is sufficient — SQLAlchemy will
>    flush in dependency order on commit.
> 
> Expected impact: ~50 % reduction in `Sql.save` time.  In the integration
> benchmark, `Pickle+Sql/lightweight/miss` is 19 ms today; this should
> roughly halve the SQL-side contribution.
> 
> ### Risk
> 
> Low – `ON CONFLICT DO NOTHING` matches the existing
> "contains-then-skip" semantics exactly.  Single-commit-per-save is a
> durability change for users who were relying on the call-row landing
> before the arguments rows; that's a niche case but worth a sanity check
> in the regression tests for `tests/regression/test_sql_concurrent_save.py`.
> 
> ---

implement in new pr

src/fleche/call.py
```diff
@@ -75,12 +104,13 @@ class Call:
 
     @classmethod
     def from_call(cls, func, *args, **kwargs):
-        arguments = dict(bind(func, args, kwargs, apply_defaults=True))
+        sig = _func_signature(func)
+        bound = sig.bind(*args, **kwargs)
+        bound.apply_defaults()
         qualname, module, version = _extract_version_info(func)
```
pmrv: Also cache the version info.  Fix the tests to construct independent functions where they relied on this behavior previously.

src/fleche/call.py
```diff
@@ -12,6 +13,34 @@
...
+    invoked on every cache hit by :meth:`Call.from_call` /
+    :class:`ArgumentPolicy`.  Caching keyed on ``func`` removes ~6% from the
+    cache-hit hot path.  Tests that mutate ``func.__signature__`` can call
+    ``_func_signature.cache_clear()`` to drop the cache.
+    """
+    return signature(func)
+
+
+@lru_cache(maxsize=None)
+def _func_code_digest(func) -> Digest | None:
```
pmrv: This will break if func is a callable object that doesn't hash (in the python sense, not the fleche sense).   Add a layer around these functions to ensure fleche won't break for such callables and add an example to the tests.

src/fleche/call.py
```diff
@@ -12,6 +13,34 @@
 logger = logging.getLogger("fleche.call")
 
 
+@lru_cache(maxsize=None)
```
pmrv: `maxsize=1000` for all functions.

src/fleche/call.py
```diff
@@ -12,6 +13,34 @@
...
+def _func_signature(func) -> Signature:
+    """Memoised :func:`inspect.signature` lookup.
+
+    ``signature()`` is a non-trivial introspection call (~30 µs) and is
+    invoked on every cache hit by :meth:`Call.from_call` /
+    :class:`ArgumentPolicy`.  Caching keyed on ``func`` removes ~6% from the
+    cache-hit hot path.  Tests that mutate ``func.__signature__`` can call
+    ``_func_signature.cache_clear()`` to drop the cache.
```
pmrv: Keep historical info (~6%, e.g) to commit messages, you may explain why we cache though.

pmrv (commented):
review

src/fleche/call.py
```diff
@@ -12,7 +13,60 @@
...
+def _func_signature(func) -> Signature:
+    """Cached :func:`inspect.signature` lookup with a fallback for
+    callables that cannot be hashed (e.g. instances with
+    ``__hash__ = None``).  Caches the per-function statics that
+    :meth:`Call.from_call` and :class:`ArgumentPolicy.check_required`
+    would otherwise re-introspect on every cache hit.
+    """
+    try:
+        return _cached_signature(func)
+    except TypeError:
```
pmrv: Cleaner to apply the try/except at the call site of func_signature.  There you can call all three cached getters in a try and fallback on all three base treatments in the except clause.

src/fleche/call.py
```diff
@@ -75,12 +97,23 @@ class Call:
...
+            code_digest = _cached_code_digest(func)
+        except TypeError:
+            # Unhashable callable (e.g. instance with __hash__ = None) —
+            # bypass the per-function caches and recompute directly.
+            sig = signature(func)
+            qualname, module, version = _extract_version_info(func)
+            code_digest = (
+                digest.digest(func.__code__) if hasattr(func, "__code__") else None
+            )
+        bound = sig.bind(*args, **kwargs)
```
pmrv: How come this differs across call classes?  Factor out and unify

pmrv (commented):
Ensure this caching on func identity comes up in the docs.  I can't really see any practical issues from this (e.g. importlib.reload would change func identity too), but maybe you can.  Mention in admonitions if so.

pmrv:
> ## Issue 5 — `hasattr(value, "__digest__")` probed on every recursive `_digest` (~2 %)
> ### Where
> 
> `src/fleche/digest.py:160`:
> 
> ```python
> match value:
>     case Digest():
>         return value
>     case _ if hasattr(value, "__digest__"):
>         return Digest(value.__digest__())
>     case str(): ...
>     case int(): ...
> ```
> 
> The `__digest__` opt-in protocol is checked _before_ the cheap built-in cases. For every `int`, `str`, `float`, `dict`, `list` (the bulk of real-world digest traffic), Python performs an attribute lookup that must walk the type chain to fail.
> ### Evidence
> 
> cProfile, same 946k-`_digest`-call workload:
> 
> ```
>    ncalls  tottime   filename:lineno(function)
>    951950    0.101   {built-in method builtins.hasattr}
> ```
> 
> That's **0.10 s of 4.80 s = 2.1 %** spent on a check that returns `False` for every built-in. Small in relative terms, but it's also the single most-called branch in the digest hot loop.
> ### Proposed fixes (cheapest first)
> 
> **Option A** — reorder the `match` so built-ins are checked first:
> 
> ```python
> match value:
>     case Digest():
>         return value
>     case str():       ...        # ← cheap, common, return early
>     case int():       ...
>     case bool():      ...
>     case None:        ...
>     case dict():      ...
>     case _ if hasattr(value, "__digest__"):
>         return Digest(value.__digest__())
>     case Iterable():  ...
> ```
> 
> Cost: zero. Behaviour change: tiny — a `dict`/`list` subclass that defines `__digest__` would currently use `__digest__` and after this change would use the dict/list path instead. Whether that's a feature or a bug depends on intent; would want a quick poll of users defining `__digest__` on container subclasses.
> 
> **Option B** — type-cache the answer:
> 
> ```python
> _HAS_DIGEST_CACHE: dict[type, bool] = {}
> 
> def _has_digest_dunder(t):
>     cached = _HAS_DIGEST_CACHE.get(t)
>     if cached is None:
>         cached = hasattr(t, "__digest__")
>         _HAS_DIGEST_CACHE[t] = cached
>     return cached
> ```
> 
> `hasattr` is replaced with a dict lookup keyed on `type(value)`. No semantic change. ~30 ns instead of ~100 ns per call.
> ### Risk
> 
>     * Option A: low–medium, depending on whether anyone relies on
>       `__digest__` overriding container behaviour on subclasses.
> 
>     * Option B: low — pure caching, identical semantics.
> 
> 
> Probably worth combining with **Issue 4**'s `_digest_bytes` refactor since both touch the same dispatch hot loop — the rearrangement is cheaper to land as one diff with one round of testing than two.
> 

So the issue here is that we want to easily allow specializing digest even for subclasses.  I don't see another way than to avoid checking first.

pmrv:
> 1. **Memory backend `copy.deepcopy` on every `get`** — ~24 % of hit-path time

Unavoidable without making UX terrible.

PR #435, opened 2026-04-28
Add cross-dialect SQL backend support for Postgres and MySQL

pmrv:
## Summary
This PR extends the `Sql` storage backend to support Postgres and MySQL/MariaDB databases in addition to SQLite, with comprehensive cross-dialect testing and CI integration.

### Core Storage Backend (`src/fleche/storage/sql.py`)
- **Fixed URL handling in `_coerce_sqlite_url()`**: Previously treated any non-`sqlite:` URL as a filesystem path, causing `postgresql://host/db` to be rewritten as `sqlite:///<cwd>/postgresql:/host/db`. Now correctly detects SQLAlchemy URLs by checking for `://` or `sqlite:` prefix and passes them through verbatim.
- **Gated SQLite-specific PRAGMA**: Added dialect check to `_enable_sqlite_foreign_keys()` to prevent `PRAGMA foreign_keys` from being registered on non-SQLite connections, which would crash on first connect.
- **Conditional connect args**: Made `check_same_thread=False` conditional on SQLite dialect, as this pysqlite-only flag raises `TypeError` on other DBAPI drivers.

### Test Infrastructure (`tests/fixtures.py`)
- **Ephemeral database management**: Added `_ephemeral_database()` context manager that creates uniquely-named test databases and cleans them up, with dialect-specific handling for Postgres (admin database, session termination) and MySQL (no database selection needed).
- **External SQL fixtures**: Implemented `postgres_sql` and `mysql_sql` fixtures that skip gracefully when connection URLs aren't provided via environment variables.
- **Parametrized call_storage**: Extended the `call_storage` fixture to include `sql_postgres` and `sql_mysql` variants when their respective `FLECHE_TEST_*_URL` env vars are set, enabling existing tests to run against all backends.

### New Regression Test Suite (`tests/regression/test_sql_non_sqlite_backends.py`)
- **URL pass-through tests**: Verify that Postgres and MySQL URLs are not rewritten as SQLite paths.
- **Dialect detection**: Confirm the correct driver is loaded, not a fallback SQLite.
- **CRUD operations**: Test save/load/list/contains/evict against real database drivers.
- **Query pushdown**: Verify JSON metadata and argument filtering work correctly with each dialect's native JSON column type.
- **Idempotency**: Ensure duplicate saves don't create duplicates.

### CI Integration (`.github/workflows/tests.yml`)
- Added `sql-backends` job that spins up Postgres 16 and MariaDB 11 services.
- Installs dialect-specific drivers (`psycopg2-binary`, `pymysql`).
- Runs cross-backend conformance tests plus all storage unit tests with both external backends active.

## Implementation Details
- The fixture design uses `request.getfixturevalue()` to lazily resolve only the configured backend, preventing unconfigured backends' skip exceptions from cascading.
- Database creation uses `AUTOCOMMIT` isolation level to allow DDL operations.
- Postgres requires connecting to an admin database (`postgres`) to create new databases; MySQL doesn't.
- All external database fixtures are properly cleaned up via context managers and engine disposal.

https://claude.ai/code/session_01RfDKAL5NMMfMHMma9W9GPx

### DIFF
--- a/src/fleche/storage/sql.py
+++ b/src/fleche/storage/sql.py
@@ -93,13 +93,24 @@ with ImportAlarm(
 
 
 def _coerce_sqlite_url(path_or_url: str | None) -> str:
+    """Normalise the user-facing ``url=`` argument.
+
+    Accepts a filesystem path (treated as sqlite), a ``sqlite:`` URL
+    (passed through, parent dir auto-created), or any other SQLAlchemy URL
+    such as ``postgresql://`` / ``mysql+pymysql://`` (passed through
+    verbatim). Only sqlite paths get the parent-dir-create convenience.
+    """
     if path_or_url is None:
         return "sqlite:///:memory:"
 
-    if isinstance(path_or_url, str) and path_or_url.startswith("sqlite:"):
-        url = path_or_url
+    s = str(path_or_url)
+
+    # Already a SQLAlchemy URL of any dialect (driver scheme contains "://"
+    # or the short ``sqlite:foo`` form). Leave it alone.
+    if "://" in s or s.startswith("sqlite:"):
+        url = s
     else:
-        abs_path = Path(str(path_or_url)).absolute()
+        abs_path = Path(s).absolute()
         url = f"sqlite:///{abs_path}"
 
     if url.startswith("sqlite:///"):
@@ -115,6 +126,11 @@ SQLITE_FOREIGN_KEYS_ON = "PRAGMA foreign_keys=ON"
 
 
 def _enable_sqlite_foreign_keys(engine) -> None:
+    # PRAGMA is sqlite-only; running it on Postgres/MySQL connections would
+    # raise at connect time, so gate the listener on the dialect.
+    if engine.dialect.name != "sqlite":
+        return
+
     @event.listens_for(engine, "connect")
     def _set_sqlite_pragma(dbapi_connection, connection_record):
         cursor = dbapi_connection.cursor()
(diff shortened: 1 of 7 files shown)


src/fleche/storage/sql.py
```diff
@@ -41,12 +41,19 @@
 
     Base = declarative_base()
 
+    # MySQL/MariaDB reject ``VARCHAR`` columns without an explicit length, so
+    # every ``String`` here is sized. 255 is the conventional MySQL upper
+    # bound for indexable VARCHARs and comfortably fits Python qualified
+    # names, function/argument names, and the JSON-encoded scalar versions
+    # we actually store.
+    _NAME_LEN = 255
```
pmrv: Can we keep this purely in mysql path then and let everyone else enjoy varchar?

tests/regression/test_sql_non_sqlite_backends.py
```diff
@@ -0,0 +1,205 @@
+"""Cross-dialect smoke + regression tests for the SQLAlchemy ``Sql`` backend.
+
+The active-by-default suite already exercises the sqlite path. This file adds
+a parametrized fixture (`external_sql`) that yields a ``Sql`` storage pointed
+at a fresh database on every supported non-sqlite backend whose connection
+URL was provided via environment variable. Tests are silently skipped when
+no URL is configured, so the file is a no-op for default local runs but
+becomes the cross-backend conformance suite under CI / when a developer
+exports e.g. ``FLECHE_TEST_POSTGRES_URL``.
```
pmrv, line 9: Document this in the doc pages under a developer section

pmrv (commented):
ewview

PR #446, opened 2026-04-30
fix: normalize string keys in Cache.load_value without requiring D() wrapping

### LINKED ISSUE #422: load_value requires explicit fleche.D() wrapping
that's annoying.

pmrv:
load_value now accepts plain hex strings (full digests or short prefixes) directly, delegating to values.load() which already calls _normalize_key() internally. Non-string non-Digest values are still returned as-is.

Closes #422

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -253,10 +253,9 @@ class Cache(BaseCache):
     calls: storage.CallStorage
 
     def load_value(self, key):
-        if not isinstance(key, Digest):
-            return key
-
-        return self.values.load(key)
+        if isinstance(key, str):  # Digest is a str subclass; values.load handles normalization
+            return self.values.load(key)
+        return key
 
     def save(self, call: Call) -> str:
         try:
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 import logging
 import random
+import string
 import threading
 from dataclasses import dataclass, replace, field
 from typing import Iterable, Any, Callable, Literal
@@ -253,7 +254,9 @@ class Cache(BaseCache):
     calls: storage.CallStorage
 
     def load_value(self, key):
-        if isinstance(key, str):  # Digest is a str subclass; values.load handles normalization
+        if isinstance(key, Digest):
+            return self.values.load(key)
+        if isinstance(key, str) and 4 <= len(key) <= _digest.DIGEST_LENGTH and all(c in string.hexdigits for c in key):
             return self.values.load(key)
         return key
 
--- a/tests/unit/caches/test_cache.py
+++ b/tests/unit/caches/test_cache.py
@@ -350,6 +350,49 @@ def test_cache_query_kwargs_and_template_raises():
         c.query(QueryCall(), name="foo")
 
 
+def test_load_value_plain_string_key():
+    """load_value should accept a plain hex string without requiring D() wrapping."""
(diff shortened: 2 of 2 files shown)


src/fleche/caches.py
```diff
@@ -253,10 +253,9 @@ class Cache(BaseCache):
     calls: storage.CallStorage
 
     def load_value(self, key):
-        if not isinstance(key, Digest):
-            return key
-
-        return self.values.load(key)
+        if isinstance(key, str):  # Digest is a str subclass; values.load handles normalization
```
pmrv: @claude What if a random string as "asdfasdf" would be passed?

pmrv: **Claude finished @pmrv's task in 2m 3s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25173779460)

---
### Analysis: passing `"asdfasdf"` to `load_value`

- [x] Read relevant code in `caches.py` and `values.py`
- [x] Trace what happens when a random string like `"asdfasdf"` is passed
- [x] Post answer

---

With the current PR code in `caches.py:255-258`:

```python
def load_value(self, key):
    if isinstance(key, str):  # Digest is a str subclass; values.load handles normalization
        return self.values.load(key)
    return key
```

Passing `"asdfasdf"` would follow this path:

1. `isinstance("asdfasdf", str)` → `True`
2. `self.values.load("asdfasdf")` is called
3. `_normalize_key("asdfasdf")` is called (`storage/base.py:132`) — `len("asdfasdf") = 8 < 64`, so it calls `expand("asdfasdf")`
4. `expand` scans all stored digests for ones starting with `"asdfasdf"` — likely finds none → **raises `KeyError("asdfasdf")`**

This is a **regression from the pre-PR behavior**. The old code was:

```python
def load_value(self, key):
    if not isinstance(key, Digest):
        return key  # plain strings returned as-is
    return self.values.load(key)
```

Before this PR, `"asdfasdf"` would be returned unchanged. Now it raises `KeyError`.

This matters because `load_value` is called when reconstructing a `DigestedCall` — arguments in a `DigestedCall` can be either `Digest` objects (to be loaded from storage) **or inline values** (returned as-is). If an inline string argument happens to not be a `Digest`, the new code would incorrectly try to load it from storage.

**Example of the breakage:** a saved call with a string argument like `"hello"` would fail on reload with `KeyError("hello")`.

The fix should restrict the new string-handling path to strings that actually look like hex digests:

```python
def load_value(self, key):
    if isinstance(key, Digest):
        return self.values.load(key)
    if isinstance(key, str):
        try:
            bytes.fromhex(key)  # only treat as a key if it's valid hex
            return self.values.load(key)
        except ValueError:
            return key
    return key
```

Or more simply, keep the original guard but also allow plain `str` that passes a hex-validity check. [Fix this →](https://claude.ai/code?q=In%20src%2Ffleche%2Fcaches.py%20on%20branch%20claude%2Fissue-422-20260430-1506%2C%20the%20load_value%20method%20at%20line%20255%20now%20calls%20self.values.load%20for%20ALL%20strings%2C%20but%20it%20should%20only%20do%20so%20for%20valid%20hex%20strings%20(Digest%20or%20plain%20hex).%20Non-hex%20inline%20strings%20should%20still%20be%20returned%20as-is.%20Fix%20by%20checking%20isinstance%28key%2C%20Digest%29%20first%2C%20then%20for%20plain%20str%20validate%20it%20looks%20like%20hex%20before%20calling%20values.load%28key%29%2C%20otherwise%20return%20key%20as-is.&repo=pmrv/fleche)

PR #462, opened 2026-05-01
refactor: extract _longest_common_prefix_length helper

### LINKED ISSUE #407: Refactor: share longest-common-prefix helper between `_resolve_prefix` and `_combine_expand`
## Location

- `src/fleche/storage/base.py:21-40` — `_resolve_prefix`
- `src/fleche/caches.py:214-234` — `_combine_expand`

## Problem

Both helpers compute the length of the longest common prefix of two digest strings in order to build the error message *"need at least N characters"* for `AmbiguousDigestError`. The loop is byte-for-byte identical in both places:

```python
m1, m2 = unique[0], unique[1]  # or candidates[0], candidates[1]
for i, (c1, c2) in enumerate(zip(m1, m2)):
    if c1 != c2:
        break
else:
    i = min(len(m1), len(m2))
raise AmbiguousDigestError(
    f"Short digest {key} is ambiguous; need at least {i+1} characters."
)
```

This is slightly tricky code — the `for/else` plus the `min(len(m1), len(m2))` fallback is easy to get wrong in a future tweak. Two copies means two places to chase bugs and two places a subtle mismatch could hide.

(Note: PR #365 / issue #331 consolidated the **fan-out loops** in `Cache.expand/shrink` vs `CacheStack.expand/shrink` and introduced `_combine_expand`/`_combine_shrink`, but the LCP logic inside `_combine_expand` was left identical to `_resolve_prefix` and the two never cross-reference. PR #324 deduped the **use** of `_resolve_prefix` between `base.py` and `sql.py`, but did not cover `caches.py`.)

## Suggested refactor

Extract a module-level helper (either in `storage/base.py` or a small utility module shared by both):

```python
def _longest_common_prefix_length(s1: str, s2: str) -> int:
    """Return the length of the longest common prefix of *s1* and *s2*.

    Used to compute a minimum digest length hint for ``AmbiguousDigestError``.
    """
    for i, (c1, c2) in enumerate(zip(s1, s2)):
        if c1 != c2:
            return i
    return min(len(s1), len(s2))
```

Both `_resolve_prefix` and `_combine_expand` call it, then build their error messages from `lcp + 1`.

Bonus: a one-line hypothesis test (`lcp(s, s) == len(s)`, `lcp(s1, s2) <= min(len(s1), len(s2))`, etc.) pins down the invariants explicitly.

## Pros

- Removes the one remaining LCP copy between `base.py` and `caches.py`.
- Makes the `for/else + min` trick a named, single-responsibility function that reads linearly.
- Easy to unit-test in isolation.
- Leaves the existing error messages unchanged (only the `i` computation moves).

## Cons / risks

- The logic is small; some may find the inline code fine. The win is that the invariant name and fallback branch become explicit.
- No external dependencies, no behaviour change if the extraction preserves `min(len(m1), len(m2))` as the fallback (not `len(m1)` — this matters when the two strings differ in length, a case the error-message hint depends on).

Low. Internal, covered by existing `tests/unit/storage` tests. Worth doing opportunistically when next touching this code path.

pmrv:
Extract the duplicated longest-common-prefix loop from `_resolve_prefix` (`storage/base.py`) and `_combine_expand` (`caches.py`) into a single named helper in `storage/base.py`, imported by `caches.py`.

Also adds 6 focused unit tests for the new helper in `tests/unit/storage/test_short_digest.py`.

Closes #407

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -11,6 +11,7 @@ import pandas as pd
 from . import digest as _digest
 from .digest import Digest  # type hint convenience
 from . import storage
+from .storage.base import _longest_common_prefix_length
 from .storage.destructuring import HasChildDigests
 from .call import Call, DigestedCall, LazyCall, QueryCall
 from . import call
@@ -225,14 +226,9 @@ def _combine_expand(key: "Digest | str", results: "Iterable[Digest]") -> "Digest
         raise KeyError(key)
     if len(unique) == 1:
         return unique[0]
-    m1, m2 = unique[0], unique[1]
-    for i, (c1, c2) in enumerate(zip(m1, m2)):
-        if c1 != c2:
-            break
-    else:
-        i = min(len(m1), len(m2))
+    lcp = _longest_common_prefix_length(unique[0], unique[1])
     raise storage.AmbiguousDigestError(
-        f"Short digest {key} is ambiguous; expands to: {unique}; need at least {i + 1} characters."
+        f"Short digest {key} is ambiguous; expands to: {unique}; need at least {lcp + 1} characters."
     )
 
 
--- a/src/fleche/storage/base.py
+++ b/src/fleche/storage/base.py
@@ -18,6 +18,13 @@ class AmbiguousDigestError(ValueError):
     pass
 
 
+def _longest_common_prefix_length(s1: str, s2: str) -> int:
+    for i, (c1, c2) in enumerate(zip(s1, s2)):
+        if c1 != c2:
+            return i
+    return min(len(s1), len(s2))
+
+
 def _resolve_prefix(key: str, candidates: list[Digest]) -> Digest:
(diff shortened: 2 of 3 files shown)


pmrv:
@claude rebase on main

pmrv:
**Claude finished @pmrv's task in 3m 6s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25218325792)

---
### Rebasing on main

- [x] Fetch latest main
- [x] Extract PR-specific changes (branch had disconnected history — no common ancestor with main)
- [x] Reset branch to origin/main and re-apply PR changes
- [x] Push rebased branch

The branch had no common ancestor with `main` (the 20 prior commits were already merged via other PRs with different SHAs), so a standard `git rebase` wasn't possible. Instead:

1. Saved the PR's commit as a patch
2. Reset the branch to `origin/main` (`fb97628`)
3. Re-applied the three changed files manually (the patch context had shifted)
4. Committed and force-pushed

The branch is now a clean single commit on top of current `main`.

PR #464, opened 2026-05-01
refactor: inline trivial make_* helper factories in wrapper.py

### LINKED ISSUE #458: Refactor: collapse the eight `make_*` helper factories in `wrapper.py`
## Location

`src/fleche/wrapper.py:174-252` — the `make_get_call`, `make_digest`, `make_query`, `make_load`, `make_contains`, `make_rerun`, `make_bind` (and `make_wrapper`) family.

## Problem

The decorator builds 7 separate helpers (`.call`, `.digest`, `.query`, `.load`, `.contains`, `.rerun`, `.bind`) plus the main `wrapper`. Each helper is created by its own top-level `make_*` factory that follows the same shape:

```python
def make_load(func, digest_func):
    @wraps(func)
    def _load_func(*args, **kwargs):
        return state._CACHE.get().load(digest_func(*args, **kwargs)).result
    return _load_func


def make_contains(func, digest_func):
    @wraps(func)
    def _contains_func(*args, **kwargs):
        return state._CACHE.get().contains(digest_func(*args, **kwargs))
    return _contains_func
```

The bodies are 1–5 lines each; the boilerplate (signature, `@wraps`, return) is the bulk of the file. Then `fleche()` itself (lines 380–409) plumbs them all together by name:

```python
get_call = make_get_call(func, policy, hash_version, hash_module, hash_code)
digest_func = make_digest(func, get_call)
query_func = make_query(func, policy)
load_func = make_load(func, digest_func)
contains_func = make_contains(func, digest_func)
wrapper = make_wrapper(func, policy, meta, isolate, get_call)
rerun_func = make_rerun(func, wrapper)
bind_func = make_bind(wrapper)
```

… followed by 7 nearly identical `_attach(wrapper, …, name=…, doc_prefix=…)` lines. Adding a new helper today requires touching three places (factory, plumbing, attach call) and the order of the plumbing list encodes implicit dependencies (e.g. `make_digest` needs `get_call`, `make_load` needs `digest_func`).

## Suggested refactor

Two complementary moves, either of which would be a real improvement on its own:

1. **Replace the factories with a small registry of helper specs.** Each helper becomes a `(name, doc_prefix, ret, builder)` record where `builder(ctx) -> Callable` reads what it needs out of a context dataclass (`func`, `policy`, `wrapper`, `get_call`, `digest_func`, …). `fleche()` walks the registry once, building each helper into a `SimpleNamespace` and `_attach`-ing it. Adding `.foo` becomes a single new spec.

2. **Or simpler: inline the trivial ones.** `make_digest`, `make_load`, `make_contains`, `make_bind` are all 1–3 line closures that do not need their own top-level factory function. They could be local closures inside `fleche()` (or methods on a small `_FlecheHelpers` dataclass), eliminating the factory layer entirely without introducing a registry.

The non-trivial ones (`make_get_call`, `make_query`, `make_rerun`, `make_wrapper`) carry real logic — those stay as named functions, but the rest of the surface shrinks to a handful of lines.

## Pros

- Roughly halves the file (the factory shells are most of the line count).
- Removes the order-sensitive plumbing block in `fleche()`; helper dependencies become explicit through their builder signatures (or local-scope captures).
- Adding a new helper is one entry instead of three coordinated edits.
- No new external dependency.

## Cons / risks

- Closures inside `fleche()` are slightly harder to unit-test in isolation than top-level `make_*` factories. Mitigation: existing helper tests work through the public `wrapper` API, so this is mostly a stylistic concern.
- A registry-based design adds one indirection (the spec list) — readers have to look up what each spec produces. The simpler "inline trivial helpers" variant avoids this trade-off.
- `@wraps(func)` + custom `__name__`/`__qualname__` munging in `_attach` (lines 141–153) must continue to work; closures defined inside `fleche()` already pick up the correct enclosing scope, so this is mechanical.

Low risk, medium reward. All helpers are exercised by integration tests; behaviour is fully determined by inputs. A safe first step is to inline only the trivial four and leave the registry option for a follow-up.

pmrv:
Inline four trivial `make_*` factory functions as local closures inside `fleche()`'s `decorator`, as suggested in #458.

The four removed top-level functions were all 1–3 line bodies with disproportionate boilerplate. The four non-trivial helpers (`make_get_call`, `make_query`, `make_rerun`, `make_wrapper`) remain as named top-level functions.

Closes #458

### DIFF
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -192,14 +192,6 @@ def make_get_call(func, policy, hash_version, hash_module, hash_code):
     return get_call
 
 
-def make_digest(func, get_call):
-    """Build the `.digest` helper that returns the lookup key for given arguments."""
-    @wraps(func)
-    def _digest_func(*args, **kwargs):
-        return get_call(*args, **kwargs).to_lookup_key()
-    return _digest_func
-
-
 def make_query(func, policy):
     """Build the `.query` helper that yields matching calls from the active cache."""
     def _query_func(
@@ -218,22 +210,6 @@ def make_query(func, policy):
     return _query_func
 
 
-def make_load(func, digest_func):
-    """Build the `.load` helper that retrieves a cached result for given arguments."""
-    @wraps(func)
-    def _load_func(*args, **kwargs):
-        return state._CACHE.get().load(digest_func(*args, **kwargs)).result
-    return _load_func
-
-
-def make_contains(func, digest_func):
-    """Build the `.contains` helper that checks cache membership for given arguments."""
-    @wraps(func)
-    def _contains_func(*args, **kwargs):
-        return state._CACHE.get().contains(digest_func(*args, **kwargs))
-    return _contains_func
-
-
 def make_rerun(func, wrapper):
     """Build the `.rerun` helper that forces recursive reevaluation through a :class:`RefreshingCache`."""
     @wraps(func)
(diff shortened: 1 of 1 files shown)


pmrv:
just busy work

PR #466, opened 2026-05-01
refactor(caches): introduce mixin-based design for delegating cache wrappers

### LINKED ISSUE #459: Refactor: extract `DelegatingCache` base for `ReadOnlyCache`/`FilteredCache`/`RefreshingCache`
## Location

`src/fleche/caches.py`:

- `ReadOnlyCache`, lines 394–430
- `FilteredCache`, lines 433–448
- `RefreshingCache`, lines 451–487

## Problem

All three classes wrap a single underlying `cache: BaseCache` and override the same eight-ish methods (`save`, `load`, `load_value`, `contains`, `evict`, `expand`, `shrink`, `_query`). Most overrides just forward to `self.cache`:

```python
# ReadOnlyCache
def load(self, key) -> LazyCall:               return self.cache.load(key)
def expand(self, key):                          return self.cache.expand(key)
def shrink(self, key):                          return self.cache.shrink(key)
def load_value(self, key):                      return self.cache.load_value(key)
def contains(self, key: str) -> bool:           return self.cache.contains(key)
def _query(self, call) -> Iterable[LazyCall]:   return self.cache.query(call)
def save(self, call):                           raise Rejected(self, call)
def evict(self, key):                           raise Rejected("Cannot evict from a ReadOnlyCache", self, key)

# RefreshingCache
def save(self, call) -> str:                    return self.cache.save(call)
def load(self, key) -> LazyCall:                raise KeyError(key)
def load_value(self, key) -> Any:               return self.cache.load_value(key)
def contains(self, key: str) -> bool:           return False
def evict(self, key):                           self.cache.evict(key)
def expand(self, key):                          return self.cache.expand(key)
def shrink(self, key):                          return self.cache.shrink(key)
def _query(self, call):                         return self.cache.query(call)

# FilteredCache (extends ReadOnlyCache)
def load(self, key) -> LazyCall: ...            # filter check then delegate
def _query(self, call): ...                      # filter loop
```

That is roughly 60 lines of mechanical forwarding for an actual customisation surface of:

| Class | Customised methods |
|-------|---------------------|
| `ReadOnlyCache`     | `save` / `evict` (both raise `Rejected`) |
| `RefreshingCache`   | `load` (raise `KeyError`), `contains` (return `False`) |
| `FilteredCache`     | `load`, `_query` (apply predicate) |

A future "log every access" or "redirect saves elsewhere" wrapper would need yet another 8 forwarding methods.

## Suggested refactor

Introduce a thin `DelegatingCache(BaseCache)` mixin (or abstract subclass) whose default implementations of every `BaseCache` method forward to `self.cache`. The three existing wrappers inherit from it and override only the methods that actually differ:

```python
@dataclass(frozen=True)
class DelegatingCache(BaseCache):
    cache: BaseCache

    def save(self, call):       return self.cache.save(call)
    def load(self, key):        return self.cache.load(key)
    def load_value(self, key):  return self.cache.load_value(key)
    def contains(self, key):    return self.cache.contains(key)
    def evict(self, key):       self.cache.evict(key)
    def expand(self, key):      return self.cache.expand(key)
    def shrink(self, key):      return self.cache.shrink(key)
    def _query(self, call):     return self.cache.query(call)


@dataclass(frozen=True)
class ReadOnlyCache(DelegatingCache):
    def save(self, call):  raise Rejected(self, call)
    def evict(self, key):  raise Rejected("Cannot evict from a ReadOnlyCache", self, key)


@dataclass(frozen=True)
class RefreshingCache(DelegatingCache):
    def load(self, key):     raise KeyError(key)
    def contains(self, key): return False
```

`FilteredCache` keeps its current `ReadOnlyCache` parent (read-only by design) and overrides only `load`/`_query`.

## Pros

- Removes ~50 lines of forwarding boilerplate.
- Makes the customisation surface of each wrapper visually unmissable: anything that *isn't* in the class body is "delegate verbatim".
- New wrapper subclasses (logging, observability, encryption proxies) become trivial — write the methods that differ.
- No new external dependency.
- The `BaseCache` API surface is small and stable, so the maintenance cost of keeping `DelegatingCache` synced is minimal.

## Cons / risks

- One extra class in the MRO — slightly more to follow when reading. Mitigation: a one-line docstring stating "default behaviour is forward-to-`self.cache`".
- Subtle behavioural change risk if someone *relies* on a wrapper *not* implementing some method. None of the existing tests appear to depend on this, but the test suite (`tests/unit/caches/`) should be re-run as a whole.
- `BaseCache` is a `@dataclass(frozen=True)` and so is each subclass — `DelegatingCache` must declare `cache: BaseCache` and subclasses must keep `frozen=True` for hashability. The existing classes already do this, so no new constraint.
- `_query` calls `self.cache.query(call)` (the public API) rather than `self.cache._query(call)`, which preserves the wrapping behaviour of `BaseCache.query` (template/kwargs handling, `QueryIterator` wrapping). The current code does the same, so no behavioural change.

Low. Self-contained to `caches.py`. Existing tests on each wrapper class will catch any divergence. Pairs naturally with #407 (the `expand`/`shrink` LCP helper consolidation) since both target this module.

pmrv:
Replaces the monolithic ReadOnlyCache/FilteredCache/RefreshingCache with four composable mixins following the same pattern as storage classes:

- DelegatingCacheMixin: forwards all BaseCache methods to self.cache
- ReadOnlyMixin: raises Rejected for save/evict
- RefreshingMixin: always misses on load/contains
- FilteringMixin: applies predicate to load/_query

Concrete classes compose only the mixins they need:
  ReadOnlyCache(ReadOnlyMixin, DelegatingCacheMixin)
  FilteredCache(FilteringMixin, ReadOnlyMixin, DelegatingCacheMixin)
  RefreshingCache(RefreshingMixin, DelegatingCacheMixin)

Closes #459

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -428,7 +428,7 @@ class CacheWrapper(BaseCache):
         return self.cache.query(call)
 
 
-class ReadOnlyMixin(BaseCache):
+class ReadOnlyMixin(CacheWrapper):
     """Raises :class:`Rejected` for ``save`` and ``evict``."""
 
     def save(self, call: Call):
@@ -439,34 +439,31 @@ class ReadOnlyMixin(BaseCache):
 
 
 @dataclass(frozen=True)
-class ReadOnlyCache(ReadOnlyMixin, CacheWrapper):
+class ReadOnlyCache(ReadOnlyMixin):
     """A cache that can only be read from."""
 
     cache: BaseCache
 
 
-class FilteringMixin(BaseCache):
-    """Filters ``load`` and ``_query`` results by a predicate.
-
-    Requires ``self.cache: BaseCache`` — compose with :class:`CacheWrapper`.
-    """
+class FilteringMixin(CacheWrapper):
+    """Filters ``load`` and ``_query`` results by a predicate."""
 
     predicate: Callable[[Call | LazyCall], bool]  # declared by the concrete class
 
     def load(self, key: str) -> LazyCall:
-        lc = self.cache.load(key)  # type: ignore[attr-defined]
+        lc = self.cache.load(key)
         if not self.predicate(lc):
             raise KeyError(key)
         return lc
 
     def _query(self, call: call.QueryCall) -> Iterable[LazyCall]:
-        for c in self.cache.query(call):  # type: ignore[attr-defined]
+        for c in self.cache.query(call):
             if self.predicate(c):
                 yield c
 
 
 @dataclass(frozen=True)
-class FilteredCache(FilteringMixin, ReadOnlyMixin, CacheWrapper):
+class FilteredCache(FilteringMixin, ReadOnlyMixin):
     """A read-only view of a cache that only exposes calls matching a predicate."""
 
     cache: BaseCache
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -393,15 +393,15 @@ class Cache(BaseCache):
         return evicted
 
 
+@dataclass(frozen=True)
 class CacheWrapper(BaseCache):
     """Forwarding base class: all BaseCache methods delegate to ``self.cache``.
 
-    Combine with behaviour mixins (ReadOnlyMixin, FilteringMixin) and a
-    concrete ``@dataclass(frozen=True)`` subclass that declares
-    ``cache: BaseCache`` as a field.
+    Combine with behaviour mixins (ReadOnlyMixin, FilteringMixin) to build
+    concrete wrapper classes without redeclaring ``cache``.
     """
 
-    cache: BaseCache  # declared by the concrete class
+    cache: BaseCache
 
     def save(self, call: Call) -> str:
         return self.cache.save(call)
@@ -442,13 +442,12 @@ class ReadOnlyMixin(CacheWrapper):
 class ReadOnlyCache(ReadOnlyMixin):
     """A cache that can only be read from."""
 
-    cache: BaseCache
 
-
+@dataclass(frozen=True)
 class FilteringMixin(CacheWrapper):
     """Filters ``load`` and ``_query`` results by a predicate."""
 
-    predicate: Callable[[Call | LazyCall], bool]  # declared by the concrete class
+    predicate: Callable[[Call | LazyCall], bool]
 
     def load(self, key: str) -> LazyCall:
         lc = self.cache.load(key)
@@ -466,9 +465,6 @@ class FilteringMixin(CacheWrapper):
 class FilteredCache(FilteringMixin, ReadOnlyMixin):
     """A read-only view of a cache that only exposes calls matching a predicate."""
 
-    cache: BaseCache
-    predicate: Callable[[Call | LazyCall], bool]
-
 
 @dataclass(frozen=True)
 class RefreshingCache(CacheWrapper):
@@ -481,8 +477,6 @@ class RefreshingCache(CacheWrapper):
     This is necessary to handle nested fleche calls during a rerun,
     otherwise forcing them to re-execute would be awkward.
     """
-
-    cache: BaseCache
 
     def load(self, key: str) -> LazyCall:
         raise KeyError(key)
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -393,17 +393,30 @@ class Cache(BaseCache):
         return evicted
 
 
-@dataclass(frozen=True)
-class ReadOnlyCache(BaseCache):
-    """A cache that can only be read from."""
+class DelegatingCacheMixin(BaseCache):
+    """All BaseCache methods forward to ``self.cache`` by default.
 
-    cache: BaseCache
+    Combine with behaviour mixins (ReadOnlyMixin, RefreshingMixin, FilteringMixin)
+    and a concrete ``@dataclass(frozen=True)`` subclass that declares
+    ``cache: BaseCache`` as a field.
+    """
 
-    def save(self, call: Call):
-        raise Rejected(self, call)
+    cache: BaseCache  # declared by the concrete class
 
-    def load(self, key) -> LazyCall:
+    def save(self, call: Call) -> str:
+        return self.cache.save(call)
+
+    def load(self, key: str) -> LazyCall:
         return self.cache.load(key)
+
+    def load_value(self, key: str) -> Any:
+        return self.cache.load_value(key)
+
+    def contains(self, key: str) -> bool:
+        return self.cache.contains(key)
+
+    def evict(self, key: str | Digest) -> None:
+        self.cache.evict(key)
 
     def expand(self, key: Digest | str) -> Digest:
         return self.cache.expand(key)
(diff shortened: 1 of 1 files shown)


pmrv:
@claude aim for something like

```python
class ReadOnlyCache(ReadOnlyMixin, Cache): ...
```

ie cut out the delegation for the mixin case.

pmrv:
**Claude finished @pmrv's task in 20m 15s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25219085147)

---
### Tasks

- [x] Gather context
- [x] Understand current implementation
- [x] Refactor `caches.py`: remove `DelegatingCacheMixin`, change inheritance to `(Mixin, Cache)`, update `FilteringMixin` to use `super()`, add `ReadOnlySizeLimitedCache`
- [x] Update `config.py` for new class hierarchy
- [x] Update tests
- [x] Install deps and run tests (844 passed, 11 skipped)
- [x] Commit and push

### Summary

Replaced `DelegatingCacheMixin` with direct `Cache` inheritance across all delegating wrapper classes:

```python
class ReadOnlyCache(ReadOnlyMixin, Cache): ...
class FilteredCache(FilteringMixin, ReadOnlyMixin, Cache): ...
class RefreshingCache(RefreshingMixin, Cache): ...
```

`FilteringMixin` now uses `super().load()` / `super()._query()` instead of `self.cache.*`. Factory methods `readonly()`, `filter()`, and `refreshing()` are now on `Cache` (and `SizeLimitedCache.readonly()`), creating instances by passing `self.values, self.calls` directly.

A `ReadOnlySizeLimitedCache(ReadOnlyMixin, SizeLimitedCache)` was added to handle the `read_only: True` + `max_size` config combination without wrapping. `wrapper.py` (`make_rerun`) was updated to call `cache.refreshing()` on the active `Cache` instance.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-459-20260501-1433)

pmrv:
@claude I actually liked the previous commit with 'DelegatingCacheMixin' better.  Drop the last commit.  `DelegatingCacheMixin` is not a true mixin, so rename.  The `RefreshingCache` is better mapped with a subclass of the delegating one, dropping the (empty) mixin.

pmrv:
**Claude finished @pmrv's task in 5m 5s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25322402184)

---
### Tasks

- [x] Gather context
- [x] Drop last commit (reset to `5631f56`)
- [x] Rename `DelegatingCacheMixin` → `CacheWrapper`
- [x] Refactor `RefreshingCache` as direct subclass of `CacheWrapper` (drop `RefreshingMixin`)
- [x] Run tests (195 passed)
- [x] Commit and push

### Summary

Dropped the previous commit and applied two targeted changes on top of `5631f56`:

1. **`DelegatingCacheMixin` → `CacheWrapper`** — it's a forwarding base class, not a true mixin, so the name now reflects that.
2. **`RefreshingMixin` removed** — its two-method body is inlined directly into `RefreshingCache(CacheWrapper)`, which is cleaner and avoids a mixin that only ever had one consumer.

```python
class ReadOnlyCache(ReadOnlyMixin, CacheWrapper): ...
class FilteredCache(FilteringMixin, ReadOnlyMixin, CacheWrapper): ...

@dataclass(frozen=True)
class RefreshingCache(CacheWrapper):
    cache: BaseCache

    def load(self, key: str) -> LazyCall:
        raise KeyError(key)

    def contains(self, key: str) -> bool:
        return False
```
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-459-20260501-1433)

pmrv:
@claude both filter and readonly mixin should probably derive from cachewrapper.  Present your arguments if you think differently.

pmrv:
**Claude finished @pmrv's task in 2m 16s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25323389466)

---
### Tasks

- [x] Gather context
- [x] Read current `caches.py` and analyze class hierarchy
- [x] Formulate and post design analysis

---

### Design analysis: should `ReadOnlyMixin` and `FilteringMixin` derive from `CacheWrapper`?

**Short answer: I agree with the proposal.** Here's the reasoning for each, plus the one trade-off to note.

---

#### `FilteringMixin(CacheWrapper)` — clear yes

`FilteringMixin` already accesses `self.cache` directly in both `load` and `_query` (with `# type: ignore[attr-defined]` comments to silence the type checker). It is already tightly coupled to `CacheWrapper` by contract — the current `BaseCache` parent is a lie. Switching to `CacheWrapper`:

- Makes the dependency explicit and honest
- Removes the `# type: ignore` workarounds (resolving the CI type-check failures noted in the CI summary)
- `self.cache` is properly declared in `CacheWrapper`, so `ty` won't complain

No downside here.

---

#### `ReadOnlyMixin(CacheWrapper)` — yes, with one trade-off acknowledged

`ReadOnlyMixin` only overrides `save`/`evict`, never touches `self.cache`. So it's not *forced* to derive from `CacheWrapper` — it's a purely behavioral mixin. The argument for the current `(BaseCache)` derivation is that it remains composable with non-wrapper concrete caches, e.g. a hypothetical:

```python
class ReadOnlySizeLimitedCache(ReadOnlyMixin, SizeLimitedCache): ...
```

With `ReadOnlyMixin(BaseCache)`, the MRO is `ReadOnlyMixin → SizeLimitedMixin → Cache`, which works correctly. With `ReadOnlyMixin(CacheWrapper)`, the MRO becomes `ReadOnlyMixin → CacheWrapper → SizeLimitedMixin → Cache`, which inserts `CacheWrapper`'s forwarding methods (requiring `self.cache`) before `SizeLimitedCache`'s own methods — a runtime error.

**However**: the dropped commit that introduced `ReadOnlySizeLimitedCache` was explicitly discarded. The current codebase has no non-wrapper use of `ReadOnlyMixin`. Keeping the theoretical flexibility at the cost of a messier concrete API (forcing `CacheWrapper` to always be listed explicitly) is not a good trade.

The MRO also works out correctly when both derive from `CacheWrapper`:

```python
# FilteredCache(FilteringMixin, ReadOnlyMixin)
# C3: FilteredCache → FilteringMixin → ReadOnlyMixin → CacheWrapper → BaseCache
```

`ReadOnlyMixin.save()` still appears *before* `CacheWrapper.save()`, so the read-only constraint is correctly enforced without having to explicitly repeat `CacheWrapper` in the bases.

---

#### Summary

| Mixin | Proposed parent | Verdict |
|---|---|---|
| `FilteringMixin` | `CacheWrapper` | Yes — already implicitly coupled; fixes type errors |
| `ReadOnlyMixin` | `CacheWrapper` | Yes — cleaner API; the one composability loss is a currently-unused pattern that was already intentionally dropped |

Resulting concrete classes:

```python
@dataclass(frozen=True)
class ReadOnlyCache(ReadOnlyMixin): ...

@dataclass(frozen=True)  
class FilteredCache(FilteringMixin, ReadOnlyMixin): ...

@dataclass(frozen=True)
class RefreshingCache(CacheWrapper): ...  # unchanged
```

Happy to implement this if you'd like.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-459-20260501-1433)

pmrv:
@claude go, ensure ty check is successful before you report back

pmrv:
**Claude finished @pmrv's task in 6m 27s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25325612616)

---
### Tasks

- [x] Gather context
- [x] Read `caches.py` and understand current structure
- [x] Make `FilteringMixin(CacheWrapper)` and `ReadOnlyMixin(CacheWrapper)`
- [x] Simplify concrete classes (drop explicit `CacheWrapper` from bases)
- [x] Run `ty` type checker — no `unresolved-attribute` errors (remaining errors are pre-existing unresolved-import for optional deps resolved in CI env)
- [x] Run tests — 849 passed, 11 skipped
- [x] Commit and push

### Summary

Both mixins now derive from `CacheWrapper`, making `self.cache` properly typed and removing the `# type: ignore[attr-defined]` workarounds that `ty` wasn't honouring anyway. Concrete classes drop the redundant `CacheWrapper` explicit base:

```python
class ReadOnlyMixin(CacheWrapper): ...
class FilteringMixin(CacheWrapper): ...

class ReadOnlyCache(ReadOnlyMixin): ...
class FilteredCache(FilteringMixin, ReadOnlyMixin): ...
class RefreshingCache(CacheWrapper): ...  # unchanged
```

The MRO for `FilteredCache` is `FilteredCache → FilteringMixin → ReadOnlyMixin → CacheWrapper → BaseCache`, which correctly enforces filtering before read-only before forwarding.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-459-20260501-1433)

src/fleche/caches.py
```diff
@@ -384,52 +393,64 @@ def gc(self) -> set[Digest]:
...
 
-    predicate: Callable[[Call | LazyCall], bool]
+    cache: BaseCache
 
-    def load(self, key) -> LazyCall:
+
+class FilteringMixin(CacheWrapper):
+    """Filters ``load`` and ``_query`` results by a predicate."""
+
+    predicate: Callable[[Call | LazyCall], bool]  # declared by the concrete class
```
pmrv: @claude make the mixin's dataclasses the same way we dealt with it for storages. I *do not* want to redeclare fields in the assembled classes.

pmrv: **Claude finished @pmrv's task in 7m 27s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25339136778)

---
### Tasks

- [x] Gather context
- [x] Understand storage mixin dataclass pattern
- [x] Refactor cache mixins to avoid field redeclaration in concrete classes
- [x] Run `ty` type check — only pre-existing unresolved-import errors for optional deps, no attribute errors
- [x] Run tests — 849 passed, 11 skipped
- [x] Commit and push

### Summary

Made `CacheWrapper` and `FilteringMixin` proper `@dataclass(frozen=True)` classes that own their fields, matching the same pattern used by storage mixins (`SerializingMixin`, `SizeLimitedMixin`). Python's dataclass machinery merges fields through MRO, so the assembled classes no longer redeclare anything:

```python
@dataclass(frozen=True)
class CacheWrapper(BaseCache):
    cache: BaseCache       # ← declared here, inherited by all subclasses

@dataclass(frozen=True)
class FilteringMixin(CacheWrapper):
    predicate: Callable    # ← declared here, inherited by FilteredCache

class ReadOnlyMixin(CacheWrapper):  # no fields, stays a plain class
    ...

@dataclass(frozen=True)
class ReadOnlyCache(ReadOnlyMixin): ...     # cache from CacheWrapper via MRO

@dataclass(frozen=True)
class FilteredCache(FilteringMixin, ReadOnlyMixin): ...  # cache + predicate from MRO

@dataclass(frozen=True)
class RefreshingCache(CacheWrapper): ...   # cache from CacheWrapper
```

[Branch](https://github.com/pmrv/fleche/tree/claude/issue-459-20260501-1433) |

PR #467, opened 2026-05-01
feat: replace custom file locking with py-filelock

### LINKED ISSUE #221: 3rd party library for file locking
https://py-filelock.readthedocs.io/en/latest/how-to.html

Apparently they are threadsafe, which we won't be.

pmrv:
Swaps the hand-rolled TOCTOU-racy write lock and polling read lock in storage/file.py for filelock.FileLock (OS-level fcntl/msvcrt, atomic). Retains "read anyway after timeout" semantics: on Timeout the read proceeds with a warning; if it then fails, raises KeyError.

Removes lock_wait_start (was only the initial exponential-backoff interval; filelock has its own configurable poll interval).

Closes #221

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     'numpy',
     'pandas >= 2.2, <4',
     'pyiron-snippets',
+    'filelock',
 ]
 
 [project.optional-dependencies]
--- a/src/fleche/config.py
+++ b/src/fleche/config.py
@@ -22,9 +22,7 @@ following **lowercase** identifiers:
     :class:`~fleche.storage.CallPickleFile`).
     Required: ``root`` (path to storage directory).
     Optional: ``compress`` (bool, default ``False``) — gzip-compress files.
-    Optional: ``lock_timeout`` (float, default ``1.0``) — write-lock wait timeout (s).
-    Optional: ``lock_wait_start`` (float, default ``0.001``) — initial lock-poll
-    interval for exponential backoff (s).
+    Optional: ``lock_timeout`` (float, default ``1.0``) — file-lock acquisition timeout (s).
     Optional: ``secret_key`` (list of hex strings) — HMAC-SHA256 signing keys;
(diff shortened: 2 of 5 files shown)


pmrv:
@claude take the opportunity to critically evaluate all tests related to file locking.  delete unnecessary ones. improve ones that are not useful.  your prime directive is to keep tests small and orthogonal

pmrv:
**Claude finished @pmrv's task in 3m 41s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25219593897)

---
### Evaluating file locking tests

- [x] Read implementation files
- [x] Read and analyze current test file
- [x] Apply changes
- [x] Run tests (6/6 pass)
- [x] Commit and push

**Changes made to `tests/unit/storage/test_file_storage.py`:**

| Test | Action | Reason |
|------|--------|--------|
| `test_file_storage_list_filtering` | Simplified | Removed 5 redundant negative `not in` assertions — `assert len(items) == 2` already proves nothing else leaked in |
| `test_load_waits_for_lock` | Cleaned up | Removed `print` statement; converted to `tmp_path` |
| `test_load_timeouts_and_reads_anyway` | Cleaned up | Removed unused `start`/`end` timing variables and timing assertion — `caplog` check is sufficient proof the timeout path was hit; converted to `tmp_path` |
| `test_load_fails_after_timeout_raises_keyerror` | Cleaned up | Converted to `tmp_path` |
| `test_save_creates_and_removes_lock` → **`test_save_releases_lock`** | Simplified | Dropped the `monkeypatch(Path.write_bytes)` that asserted the lock exists mid-write — that's an implementation detail already guaranteed by `with filelock.FileLock(...):`. Kept only the post-save re-acquire check, which is the observable behaviour |
| `test_list_excludes_locks` | **Deleted** | Redundant subset of `test_file_storage_list_filtering` |
| `test_evict_removes_lock` | Cleaned up | Converted to `tmp_path` |

Also removed the `tempfile` import (no longer used).
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-221-20260501-1452)

PR #478, opened 2026-05-04
fix: eliminate race between Future.result() and done-callback cache save

### LINKED ISSUE #477: 🤖 **CI failure summary** for [`5368017`](https://github.com/pmrv/fleche/actions/runs/25340694846) on `claude/issue-353-20260504-1946`
🤖 **CI failure summary** for [`5368017`](https://github.com/pmrv/fleche/actions/runs/25340694846) on `claude/issue-353-20260504-1946`

**Summary**  
A single test is failing in the futures caching logic: when calling a cached function that returns a `Future`, the second call returns the cached `Future` object instead of its resolved value.

**Root cause**  
In `tests/unit/fleche/test_futures.py::TestThreadPoolFutures::test_multiple_futures_concurrent`, the cache is returning a pending `Future` object on the second call rather than the computed result:

```
assert [0, 1, 4, <Future at 0x... state=pending>, 16] == [0, 1, 4, 9, 16]
  At index 3 diff: <Future at 0x...> != 9
```

The test expects `compute(3)` to return `9` (the cached computed value), but instead it's returning the original `Future` object from the first call.

**Suggested fix**  
Review the futures caching logic in the issue #353 changes. The cache should either:
1. Not cache `Future` objects themselves, or  
2. Automatically resolve cached `Future` objects before returning them

---
_Generated by Claude Haiku 4.5._

_Originally posted by @claude[bot] in https://github.com/pmrv/fleche/issues/476#issuecomment-4374167703_

pmrv:
CPython's Future.set_result() releases the condition lock before calling _invoke_callbacks(), so result() can return in the main thread before the _cache done-callback has run in the worker thread. A second call with the same args would then find a cache miss and return a new pending Future.

Track each in-flight Future in a per-wrapper dict (_in_flight). On a cache miss, if the key is already in-flight, wait on that Future's result and re-check the cache (or return the value directly if the callback lost the race). The cleanup callback pops the key only after the cache save, so the dict always covers the gap window.

Closes #477

### DIFF
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -1,4 +1,5 @@
 import os
+import threading
 from pathlib import Path
 from functools import wraps, partial
 from types import SimpleNamespace
@@ -169,6 +170,14 @@ def make_rerun(func, wrapper):
 
 def make_wrapper(func, policy, meta, isolate, get_call):
     """Build the cached wrapper returned by :func:`fleche`."""
+    # Tracks futures currently being computed so a second call with the same
+    # key can wait for the first result rather than re-executing.  This closes
+    # the race where CPython notifies result() waiters before invoking done
+    # callbacks, allowing a cache miss on the second call even though the first
+    # future has already resolved.
+    _in_flight: dict[str, Future] = {}
+    _in_flight_lock = threading.Lock()
+
     @wraps(func)
     def wrapper(*args: Any, **kwargs: Any) -> _T:
         cache: BaseCache = state._CACHE.get()
@@ -204,6 +213,18 @@ def make_wrapper(func, policy, meta, isolate, get_call):
         except KeyError:
             logger.debug("Cache miss for %s with key %s", call.name, key)
 
+        # If a concurrent call is already computing this key, wait for it and
+        # return the resolved value (or re-check the cache if the callback has
+        # already saved it by the time we get here).
+        with _in_flight_lock:
+            in_flight = _in_flight.get(key)
+        if in_flight is not None:
+            val = in_flight.result()
+            try:
+                return cache.load(key).result
+            except KeyError:
+                return val
+
         def _run_and_cache():
             active_meta = state._METADATA.get() + tuple(meta)
             metadata: Dict[str, Any] = defaultdict(dict)
@@ -235,7 +256,17 @@ def make_wrapper(func, policy, meta, isolate, get_call):
             if not isinstance(result, Future):
                 return _cache()
             else:
-                result.add_done_callback(_cache)
+                with _in_flight_lock:
+                    _in_flight[key] = result
+
+                def _cache_and_cleanup(f):
+                    try:
+                        _cache(f)
+                    finally:
+                        with _in_flight_lock:
+                            _in_flight.pop(key, None)
+
+                result.add_done_callback(_cache_and_cleanup)
                 return result
 
         if isolate:


pmrv:
@claude I don't believe the bug is in the wrapper.  I think the test suite should be more robust before we make changes to the main code.  Check again the offended test case.  Probably it should `f()` then wait on `f.contains()` and only then construct the value for the assert.  If it's a real race condition, draw some ELI5 ascii art for me to understand the problem and timelines

pmrv:
**Claude finished @pmrv's task in 3m 6s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25344412607)

---
### Analysis: race is real, but the test is also wrong

- [x] Read repo instructions and changed files
- [x] Examine the test case for the race condition
- [x] Determine if race is in wrapper or test
- [x] Respond with analysis and ASCII art

---

You're right that the test is fragile. Both can be true at the same time: the CPython race is real, **and** the test asserts too eagerly, which is why the test was the first thing that broke.

### The CPython race (ELI5)

```
Main Thread                         Worker Thread
──────────────────────────────────  ──────────────────────────────────
compute(5)                          
  cache miss
  result = executor.submit(λ)       λ() runs → 25
  result.add_done_callback(_cache)  Future.set_result(25):
  return result                       lock.acquire()
                                      _state = FINISHED
f.result() ← BLOCKS                   condition.notify_all()
                                      lock.release()     ← unblocks f.result()
f.result() RETURNS 25  ◄────────────
                                    ╔══════════════════════════════╗
compute(5) again ← called here      ║  GAP: cache still empty!     ║
  cache miss!                       ╚══════════════════════════════╝
  starts new execution                _invoke_callbacks() runs now
                                      _cache_and_cleanup()
                                      cache.save(...)   ← too late
```

`Future.set_result()` releases the condition lock **before** calling `_invoke_callbacks()`. So `f.result()` can unblock the main thread while the done callback is still pending on the worker thread.

---

### Why the test is wrong

`test_result_is_cached_after_future_completes` and `test_multiple_futures_concurrent` both do:

```python
future = compute(5)
assert future.result() == 10        # ← main thread unblocked here
# comment says "Done callback has run" — this is NOT guaranteed
second = compute(5)                 # ← may still be a cache miss!
assert second == 10
```

The comment in the test ("Done callback has run — result is now cached") is wrong. It only *tends* to be true because of scheduler timing. The race is narrow but reproducible under load, which is exactly what a flaky test looks like.

### The test fix (no wrapper changes needed for this case)

After `f.result()`, spin on `compute.contains(arg)` before re-calling. The callback *will* run, just not synchronously with `result()`:

```python
def test_result_is_cached_after_future_completes(self):
    import time
    call_count = 0
    executor = ThreadPoolExecutor(max_workers=1)
    try:
        @fleche
        def compute(x):
            nonlocal call_count
            call_count += 1
            return executor.submit(lambda: x * 2)

        with cache(_make_cache()):
            future = compute(5)
            assert future.result() == 10
            assert call_count == 1

            # Wait for the done callback to populate the cache.
            # f.result() can return before _invoke_callbacks() fires (CPython
            # releases the condition lock before calling callbacks).
            deadline = time.monotonic() + 2.0
            while not compute.contains(5):
                assert time.monotonic() < deadline, "cache never populated"
                time.sleep(0.001)

            second = compute(5)
            assert second == 10
            assert call_count == 1
    finally:
        executor.shutdown(wait=False)
```

And in `test_multiple_futures_concurrent`, after the `[f.result() for f in futures]` line, wait for all `compute.contains(i)` before the second batch of calls.

---

### Does the wrapper change still add value?

Yes, but for a **different scenario** — truly concurrent callers on separate threads both calling `compute(5)` for the first time simultaneously. The test doesn't exercise that; it's a single caller calling twice sequentially but too fast. For the sequential-reuse case the test fix is sufficient.

My recommendation: revert the wrapper change, fix the two tests with the `contains()` wait pattern, and if the concurrent-caller scenario is a real concern, file a separate issue for it with a proper multi-threaded regression test that demonstrates the failure.

pmrv:
@claude Instead of reverting, just follow your recommendation in a new PR.

pmrv:
**Claude finished @pmrv's task in 2m 38s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25345818023)

---
### Test fixes implemented in a new PR

- [x] Read repo instructions (AGENTS.md)
- [x] Find and read the relevant test files
- [x] Create new branch from main (`fix/future-cache-race-test-robustness`)
- [x] Fix `test_result_is_cached_after_future_completes` with `contains()` wait pattern
- [x] Fix `test_different_args_are_cached_independently` with `contains()` wait pattern
- [x] Fix `test_multiple_futures_concurrent` with `contains()` wait pattern
- [x] Commit and push changes
- [x] Create new PR → **[#484](https://github.com/pmrv/fleche/pull/484)**

---

All three racy tests now call `_wait_for_cache(compute, arg)` after `future.result()` before re-invoking the cached function. The helper spins on `fn.contains()` until the done callback has had a chance to save to the cache (2 s deadline). No changes to `wrapper.py`. All 7 tests in `test_futures.py` pass.
 | [New PR #484](https://github.com/pmrv/fleche/pull/484)

pmrv:
@claude In a new issue tagged 'backburner', clearly document the issue and the proposed fix.  A few things should be changed in the design: the future should only be in the in flight dict during the `_cache`, not during the actual run of the process behind the future. That is we only take responsibility of proper future handling if the result is in our hands and not while it's still running.  I want to minimize our exposure to dealing with ~state~ stale executors and futures, since fleche's inability to cope with stampedes is clearly documented already.

pmrv:
**Claude finished @pmrv's task in 2m 4s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25345964131)

---
- [x] Read repo instructions (AGENTS.md)
- [x] Read current wrapper.py implementation
- [x] Create GitHub issue tagged 'backburner' with the refined design

---

Created **[issue #485](https://github.com/pmrv/fleche/issues/485)**: *fix: narrow the `_in_flight` window to cover only the cache-write gap, not the full computation*

The issue documents:
- The CPython race with ASCII timeline
- Why the current PR #478 design is too broad (tracks futures for the full computation lifetime)
- The proposed design: enter `_in_flight` only at the start of `_cache_and_cleanup`, exit after `cache.save` completes
- The explicit trade-off table (stampedes remain out of scope, consistent with documented limitations)
- Suggested regression test using a `threading.Barrier` to reliably trigger the gap window
 | [Issue #485](https://github.com/pmrv/fleche/issues/485)

PR #483, opened 2026-05-04
fix(digest): handle generic Mapping types via type-aware case

### LINKED ISSUE #479: digest should handle generic iterables and mappings
iterables work, but mappings not, which are instead handling implicitly via the iterables case, which is not at all intended.

pmrv:
Add a `Mapping` match case before `Iterable` so non-dict Mapping subclasses (ChainMap, custom ABCs, etc.) are hashed over their key-value pairs instead of just their keys.  The existing `m.update(t.__name__.encode())` prefix means different mapping types with identical content produce distinct digests.

Add tests verifying:
- value change in a non-dict Mapping changes its digest
- key change in a non-dict Mapping changes its digest
- different mapping types with same content hash differently
- same type + same content always hashes identically (order-invariant)

Closes #479

### DIFF
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -1,4 +1,5 @@
 import cmath
+import datetime
 import hashlib
 import logging
 import dataclasses
@@ -134,6 +135,18 @@ def load_entry_points():
             logger.error("Failed to load entry point %s: %s", ep.name, e)
 
 
+def _digest_mapping(type_salt: str, contents: dict) -> Digest:
+    m = hashlib.sha256()
+    m.update(type_salt.encode())
+    sorted_items = sorted(
+        ((digest(k), k, v) for k, v in contents.items()), key=lambda item: item[0]
+    )
+    for k_digest, k, v in sorted_items:
+        m.update(k_digest.encode())
+        m.update(digest(v).encode())
+    return Digest(m.hexdigest())
+
+
 def digest(value: Any) -> Digest:
     try:
         return _digest(value)
@@ -248,6 +261,16 @@ def _digest(value: Any) -> Digest:
             if hasattr(value, "co_exceptiontable"):
                 props.append(value.co_exceptiontable)
             m.update(digest(tuple(props)).encode())
+        case datetime.timezone():
+            m.update(digest(value.utcoffset(None)).encode())
+        case datetime.timedelta():
+            m.update(digest(value.total_seconds()).encode())
+        case datetime.datetime():  # datetime subclasses date; must precede date case
+            m.update(value.isoformat().encode())
+        case datetime.date():
+            m.update(value.isoformat().encode())
+        case datetime.time():
+            m.update(value.isoformat().encode())
         case _ if dataclasses.is_dataclass(value):
             # cannot use asdict because it recursively converts values which destroys digests
             # instead (flat-) convert to dictionaries, salt with type name, then fallback to dictionary case.
@@ -260,12 +283,7 @@ def _digest(value: Any) -> Digest:
             # with the same name + field layout hash identically.
             m.update(digest(dict(_attrs.field_items(value))).encode())
         case Mapping():
-            sorted_items = sorted(
-                ((digest(k), k, v) for k, v in value.items()), key=lambda item: item[0]
-            )
-            for k_digest, k, v in sorted_items:
-                m.update(k_digest.encode())
-                m.update(digest(v).encode())
+            return _digest_mapping(str(type(value)), dict(value))
         case Iterable():
             for v in value:
                 m.update(digest(v).encode())
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -135,9 +135,7 @@ def load_entry_points():
             logger.error("Failed to load entry point %s: %s", ep.name, e)
 
 
-def _digest_mapping(type_salt: str, contents: dict) -> Digest:
-    m = hashlib.sha256()
-    m.update(type_salt.encode())
+def _digest_mapping(m, contents: dict) -> Digest:
     sorted_items = sorted(
         ((digest(k), k, v) for k, v in contents.items()), key=lambda item: item[0]
     )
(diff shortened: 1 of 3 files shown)


src/fleche/digest.py
```diff
@@ -267,6 +267,13 @@ def _digest(value: Any) -> Digest:
             # mirror the dataclass digest format so an attrs class and a dataclass
             # with the same name + field layout hash identically.
             m.update(digest(dict(_attrs.field_items(value))).encode())
+        case Mapping():
```
pmrv, line 283: This also means we no longer need the explicit dict case

pmrv (commented):
@claude revise

pmrv:
**Claude finished @pmrv's task in 2m 11s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25346112688)

---
### Revision complete

- [x] Gather context
- [x] Remove explicit `dict` case from digest.py (covered by `Mapping` case)
- [x] Run tests to verify (33/33 passed)
- [x] Commit and push

Removed the `case dict():` block (lines 222–229 of `digest.py`). Since `dict` is a `Mapping` subclass, the new `case Mapping():` case handles it identically — same sort-by-key-digest logic. Plain dict digests are unaffected because the type-name prefix (`m.update(t.__name__.encode())`) is applied before the match, so `dict` and other `Mapping` subclasses with identical content still produce distinct digests.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-479-20260504-2121)

pmrv:
@claude resolve conflict and add a internal helper `_digest_mapping(typestr, contents)`, where `typestr` (use a different name) is used as the salt for the digest.  the mapping case then becomes `_digest_mapping(str(type(mapping)), dict(mapping))`.

pmrv:
**Claude finished @pmrv's task in 4m 55s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25346521157)

---
### Revision complete

- [x] Gather context
- [x] Resolve merge conflict (incorporate `import datetime` + datetime match cases from main)
- [x] Extract `_digest_mapping` helper with `type_salt` parameter
- [x] Update mapping case to call `_digest_mapping(str(type(mapping)), dict(mapping))`
- [x] Run tests (47/47 passed)
- [x] Commit and push

**What changed:**

- Added `_digest_mapping(type_salt: str, contents: dict) -> Digest` helper before `digest()` in `digest.py`. The parameter is named `type_salt` (rather than `typestr`) to make its role as a hash prefix clear.
- The `Mapping()` match case is now a one-liner: `return _digest_mapping(str(type(value)), dict(value))`
- Merged the datetime support from main (`#480`): added `import datetime` and the five datetime match cases (`timezone`, `timedelta`, `datetime`, `date`, `time`) to resolve the branch conflict, plus the corresponding tests

Note: `str(type(value))` (e.g. `"<class 'dict'>"`) replaces `t.__name__` (e.g. `"dict"`) as the salt for mapping digests, so dict/Mapping hash values change from the previous implementation.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-479-20260504-2121)

pmrv:
@claude Use t.__name__ to not break things.  fix more conflicts.

pmrv:
**Claude finished @pmrv's task in 6m 56s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25346790577)

---
### Revision complete

- [x] Gather context
- [x] Fix `digest.py`: `str(type(value))` → `type(value).__name__` so dict digests stay compatible with main (`SHA256("dict" + sorted_kv)` in both)
- [x] Fix `test_futures.py`: sync with main's PR #484 — add `_wait_for_cache` helper + usages to avoid the CPython race where `future.result()` returns before `_invoke_callbacks()` fires
- [x] Run tests (47 digest + 7 futures passed)
- [x] Commit and push

**What changed:**

- `src/fleche/digest.py` line 285: `_digest_mapping(str(type(value)), ...)` → `_digest_mapping(type(value).__name__, ...)`. For a plain `dict` this produces salt `"dict"`, which matches the outer `m.update(t.__name__.encode())` path in main's `case dict()` block — digests are unchanged.
- `tests/unit/fleche/test_futures.py`: brought in line with main (PR #484). Added `import time`, `_wait_for_cache` polling helper, and three call sites before re-invoking cached futures. The original `# Done callback has run` comment was optimistic — CPython can return from `future.result()` before the done callback actually runs.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-479-20260504-2121)

src/fleche/digest.py
```diff
@@ -135,6 +135,18 @@ def load_entry_points():
             logger.error("Failed to load entry point %s: %s", ep.name, e)
 
 
+def _digest_mapping(type_salt: str, contents: dict) -> Digest:
```
pmrv: @claude Actually, make the helper take the hash context from outside, so that do not duplicate the type salting logic.

pmrv: **Claude finished @pmrv's task in 2m 26s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25442451143)

---
### Revision complete

- [x] Gather context
- [x] Refactor `_digest_mapping` to accept `m` (hash context) instead of `type_salt`
- [x] Update call site in `_digest()` to pass `m` directly
- [x] Run tests (47/47 passed)
- [x] Commit and push

**What changed:**

`_digest_mapping` no longer creates its own `hashlib.sha256()` context or applies a type-salt — it now takes the already-prepared `m` from `_digest()`. Since `m.update(t.__name__.encode())` is applied at line 195 before the `match` statement, the type salt is in place when `_digest_mapping(m, dict(value))` is called. Digest output is unchanged.

Before:
```python
def _digest_mapping(type_salt: str, contents: dict) -> Digest:
    m = hashlib.sha256()
    m.update(type_salt.encode())   # duplicate of line 195 in _digest()
    ...
```

After:
```python
def _digest_mapping(m, contents: dict) -> Digest:
    # m already has t.__name__ salted in by the caller
    ...
```

[Branch](https://github.com/pmrv/fleche/tree/claude/issue-479-20260504-2121) |

PR #484, opened 2026-05-04
test(futures): wait for cache population before re-calling after future.result()

### LINKED ISSUE #477: 🤖 **CI failure summary** for [`5368017`](https://github.com/pmrv/fleche/actions/runs/25340694846) on `claude/issue-353-20260504-1946`

pmrv:
## Summary

- CPython's `Future.set_result()` releases the condition lock **before** calling `_invoke_callbacks()`, so `future.result()` can unblock the caller while the done callback that saves to the cache is still pending on the worker thread.
- The previous tests assumed the callback had already run by the time `result()` returned, making them racy under load (narrow window, but real — see #477).
- Adds a `_wait_for_cache()` helper that spins on `fn.contains()` with a 2 s deadline, and applies it in the three affected tests before re-calling the decorated function.
- No changes to `src/fleche/wrapper.py`; the wrapper itself is correct.

Closes #477

## Test plan

- [ ] All 7 tests in `tests/unit/fleche/test_futures.py` pass locally
- [ ] No other tests regressed (`pytest tests/`)

Related: #478 (the wrapper-level fix approach — superseded by this simpler test fix)

### DIFF
--- a/tests/unit/fleche/test_futures.py
+++ b/tests/unit/fleche/test_futures.py
@@ -5,6 +5,7 @@ When a fleche-decorated function returns a concurrent.futures.Future,
 fleche passes the future back to the caller and attaches a done callback
 that caches the result once the future completes.
 """
+import time
 from concurrent.futures import Future, ThreadPoolExecutor, ProcessPoolExecutor
 from unittest.mock import Mock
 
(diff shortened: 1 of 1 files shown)


tests/unit/fleche/test_futures.py
```diff
@@ -14,6 +15,19 @@
 from fleche.storage import ValueMemory, CallMemory
 
 
+def _wait_for_cache(fn, *args, timeout=2.0):
```
pmrv:
```suggestion
def _wait_for_cache(fn, *args, timeout=0.1):
```

PR #489, opened 2026-05-06
bench: add locked memory variants to benchmarks (issue #449)

### LINKED ISSUE #449: Cache race: ValueMemory/CallMemory have no thread-safety

## Problem

`ValueMemory` and `CallMemory` (the default in-memory backends, used by the `Cache` created at import time in `state.py`) carry **no lock mixin**.  All disk-backed classes include `PerKeyLockMixin`, but the memory classes do not:

```python
class ValueMemory(DestructuringMixin, ValueMixin, MemoryBackend): ...
class CallMemory(CallMixin, MemoryBackend): ...  # no SerializingMixin/PerKeyLockMixin
```

`CallMixin.save()` in `base.py:239–245` has a non-atomic read-modify-write:

```python
def save(self, call: DigestedCall) -> Digest:
    key = call.to_lookup_key()
    with self._operation_context(key):   # no-op for MemoryBackend
        if self.contains(key):
            self.evict(key)              # evict may run while another thread is mid-put
        return self.put(call, key)
```

Concurrent saves on the same key can interleave: thread A checks `contains → False`, thread B checks `contains → False`, thread A calls `put`, thread B calls `put` — silent double-write.  Worse, thread A `evict` then `put` can interleave with thread B's `put`, losing a save.

Additionally, `MemoryBackend.list()` returns a snapshot (`tuple(self.storage.keys())`), but `CallMixin.query()` iterates that snapshot and calls `self.load(key)` for each; a concurrent evict between `list()` and `load()` will raise a spurious `KeyError`.  `CallMixin.query()` at `base.py:253–269` has no `try/except` around `self.load(key)` to absorb this.

## Impact

Corrupt or lost writes under concurrent access to the default in-memory cache.  Spurious `KeyError` raised from `query()`.  Because the default cache is created at module-import time, any multi-threaded program that uses `@fleche` without an explicit cache is affected.

## Possible fix

Add `SerializingMixin` (or `PerKeyLockMixin`) to `ValueMemory` and `CallMemory`, mirroring the disk-backed classes.  Alternatively, add a `try/except KeyError: continue` guard in `CallMixin.query()` to at minimum make concurrent iteration safe.

pmrv:
Add `SerializingMixin`-locked variants of `ValueMemory` and `CallMemory` to both benchmark scripts to measure the performance cost of adding thread-safety to the default in-memory backends.

Key finding: the lock acquisition adds ~1.5 µs per raw storage operation but is negligible at the `@fleche` integration level where SHA256 hashing dominates.

Also documents that `PerKeyLockMixin` cannot be used with `MemoryBackend` due to the unhashable `dict` field.

Closes #449


pmrv:
@claude Make it work with PerKeyLockMixin, probably needs some fiddling with `__hash__`.  Use the fact that each storage is a world onto itself, it's not a problem if the hash is constant if the underlying storage dict changes nor that two storages with same content (but different instances) hash differently.  `__hash__` only needs to map cleanly onto the object identify of the storage.

pmrv:
**Claude finished @pmrv's task in 7m 45s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25445728241)

---
### Making `MemoryBackend` work with `PerKeyLockMixin`

- [x] Read relevant source files
- [x] Add `__hash__ = object.__hash__` to `MemoryBackend`, `ValueMemory`, and `CallMemory`
- [x] Update `benchmark_storage.py` with `PerKeyLockMixin` variant
- [x] Update `benchmark_integration.py` with `PerKeyLockMixin` variant
- [x] Update `run_benchmarks.py` color mapping
- [x] Update test comment in `test_thread_safe.py`
- [x] Run tests (246 storage tests pass)
- [x] Commit and push

### What was done

**Root fix (`src/fleche/storage/memory.py`):** Added `__hash__ = object.__hash__` to `MemoryBackend`, `ValueMemory`, and `CallMemory`. The subtlety: `@dataclass(frozen=True)` checks `'__hash__' in cls.__dict__` (own dict only, not inherited) to decide whether to generate one — so without an explicit declaration in each class body, the decorator would generate a `__hash__` based on `storage: dict` that raises `TypeError` when called. Declaring it explicitly at each level gives every instance a stable, identity-based hash. This matches the contract the user described: each storage is its own world, hash is constant regardless of dict mutations, and two instances with the same contents are still distinct.

**Benchmarks:** Added `Memory+Locked(PerKey)` variants to both `benchmark_storage.py` and `benchmark_integration.py` (and the color map in `run_benchmarks.py`). The per-key benchmark classes also declare `__hash__ = object.__hash__` for the same dataclass reason.

pmrv:
Closes #171.

pmrv:
@claude Use per key locking by default in the memory storages, but keep a 'raw' version in the benchmarks.

pmrv:
**Claude finished @pmrv's task in 4m 40s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25446825500)

---
### Baking `PerKeyLockMixin` into `ValueMemory` / `CallMemory` by default

- [x] Read relevant source files
- [x] Update `memory.py` — add `PerKeyLockMixin` to `ValueMemory` and `CallMemory`
- [x] Update `benchmark_storage.py` — add raw unlocked class, remove redundant `ValueMemoryPerKey`
- [x] Update `benchmark_integration.py` — add raw unlocked classes, remove redundant `PerKey` classes
- [x] Update `run_benchmarks.py` — adjust color map
- [x] Update `AGENTS.md` — reflect new default MROs
- [x] Run tests (229 passed, 9 pre-existing optional-dep errors)
- [x] Commit and push

### What was done

**`src/fleche/storage/memory.py`:** Added `PerKeyLockMixin` to the MRO of both `ValueMemory` and `CallMemory`. They now match the pattern of every other concrete backend (`ValuePickleFile`, `CallPickleFile`, `Sql`).

```python
class ValueMemory(PerKeyLockMixin, DestructuringMixin, ValueMixin, MemoryBackend)
class CallMemory(PerKeyLockMixin, CallMixin, MemoryBackend)
```

**Benchmarks:** Added unlocked `ValueMemoryRaw` / `CallMemoryRaw` classes (building directly from `MemoryBackend`) so the locking overhead remains measurable. The old `ValueMemoryPerKey` / `CallMemoryPerKey` benchmark classes are removed since `ValueMemory` itself is now that. Labels: `"Memory(Raw)"` for unlocked, `"Memory"` for the default (now locked), `"Memory+Locked(Serializing)"` for the serializing variant.

**`AGENTS.md`:** Storage composition table updated to show the new default MROs.

pmrv:
@claude remove the fully serializing variant, call raw (no locking) instead.  Make the shapes for the new entries also circles not squares.

pmrv:
**Claude finished @pmrv's task in 1m 54s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25447730829)

---
### Removing serializing variant and updating markers

- [x] Read benchmark files to understand current structure
- [x] Remove `Memory+Locked(Serializing)` from `benchmark_storage.py` (dropped `ValueMemorySerializing` class and `SerializingMixin` import)
- [x] Remove `Memory+Locked(Serializing)` from `benchmark_integration.py` (dropped `ValueMemorySerializing`, `CallMemorySerializing` classes and `SerializingMixin` import)
- [x] Update `run_benchmarks.py` — removed serializing color entry, changed `Memory` marker from `🟪` (square) to `🟣` (circle)
- [x] Commit and push

Done. The `Memory+Locked(Serializing)` variant is gone from both benchmarks. The remaining new entries — `Memory(Raw)` (unlocked) and `Memory` (PerKeyLock default) — both now use `🟣`.

Issue #440, opened 2026-04-29
## Issue 4 — Iterable / dict digest hex-encodes every child (~12 % of `digest()` for big iterables)


pmrv:
### Where

`src/fleche/digest.py:243` (Iterable case) and `:204-205` (dict case):

```python
case dict():
    sorted_items = sorted(...)
    for k_digest, k, v in sorted_items:
        m.update(k_digest.encode())          # ← 64 hex chars → 64 utf-8 bytes
        m.update(digest(v).encode())         # ← same again

case Iterable():
    for v in value:
        m.update(digest(v).encode())         # ← every recursion eats this
```

Each child digest is computed as 32 raw bytes, formatted to 64 hex
chars, returned as a `Digest` (str), then re-encoded to 64 utf-8 bytes
to feed back into the parent SHA-256.  The hex round-trip is pure waste.

### Evidence

cProfile, `digest(generate_examples(st.lists(st.integers(), min_size=100)))`
(50 outer calls; 946k recursive `_digest` calls):

```
   ncalls  tottime  cumtime  filename:lineno(function)
946900/50    2.869    4.796  src/fleche/digest.py:133(_digest)
  2835600    0.394    0.394  {method 'update' of '_hashlib.HASH' objects}
   946900    0.386    0.386  {method 'hexdigest' of '_hashlib.HASH' objects}
  1893750    0.205    0.205  {method 'encode' of 'str' objects}
   946900    0.203    0.203  {built-in method _hashlib.openssl_sha256}
```

`hexdigest` (0.39 s) + `str.encode` of those hex strings (~0.10 s of the
0.21 s total) ≈ **0.49 s of 4.80 s = 10 %** spent translating between hex
strings and the bytes the parent hash actually consumes.

### Proposed fix

Introduce an internal `_digest_bytes(value) -> bytes` returning the raw
32-byte SHA-256, and rewrite the recursive cases to use it:

```python
def _digest_bytes(value):
    m = hashlib.sha256()
    ...
    case Iterable():
        for v in value:
            m.update(_digest_bytes(v))       # 32 raw bytes, no hex
    ...
    return m.digest()                        # raw bytes

def digest(value):
    return Digest(_digest_bytes(value).hex())
```

Saves the hex format + encode on every recursive call and halves the
bytes fed into each parent hash.

### Caveat

**This changes the wire format of every digest.**  All persisted caches
become unreadable.  Has to be coordinated with a `hash_version` bump
[...]

PR #490, opened 2026-05-06
refactor(digest): introduce _digest_bytes for isolated wire-format boundary

pmrv:
Renames `_digest` → `_digest_bytes(value) -> bytes` and adds a thin `digest()` wrapper that calls `.decode()` on the result.  All recursive call sites inside `_digest_bytes` (iterables, dicts, numpy arrays, dataclasses, attrs, CodeType, datetime types) now call `_digest_bytes` directly and pass the raw bytes to `m.update()`, eliminating every intermediate `digest(v).encode()` call.

For now `_digest_bytes` returns `m.hexdigest().encode()` (64 UTF-8 hex bytes) so the wire format is unchanged and no cache migration is needed. The single-spot change to gain the raw-bytes speedup (Issue #440) is:
  - `_digest_bytes`: `m.hexdigest().encode()` → `m.digest()`
  - `digest`: `.decode()` → `.hex()`
  - early-return paths: `.encode()` → `bytes.fromhex(...)`

### DIFF
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -142,8 +142,6 @@ def _digest_mapping(m, contents: Mapping) -> bytes:
     for k_bytes, k, v in sorted_items:
         m.update(k_bytes)
         m.update(_digest_bytes(v))
-    # To gain the raw-bytes speedup (Issue #440), change this to m.digest() in
-    # sync with the return at the end of _digest_bytes.
     return m.hexdigest().encode()
 
 
@@ -188,6 +186,7 @@ def _digest_bytes(value: Any) -> bytes:
     t = type(value)
     if t not in _TYPES_WITHOUT_DIGEST:
         if hasattr(t, "__digest__"):
+            # For raw-bytes speedup (Issue #440): bytes.fromhex(value.__digest__())
             return value.__digest__().encode()
         _TYPES_WITHOUT_DIGEST.add(t)
 
@@ -202,7 +201,7 @@ def _digest_bytes(value: Any) -> bytes:
             )
         case Digest():
             # Must precede str (Digest ⊂ str)
-            return value.encode()
+            m.update(value.encode())
         case str():
             m.update(value.encode())
         case None:
@@ -289,9 +288,9 @@ def _digest_bytes(value: Any) -> bytes:
             names = getattr(value, "__all__", None)
             if names is None:
                 names = dir(value)
-            return _digest_mapping(m, {name: getattr(value, name) for name in names})
+            _digest_mapping(m, {name: getattr(value, name) for name in names})
         case Mapping():
-            return _digest_mapping(m, dict(value))
+            _digest_mapping(m, dict(value))
         case Iterable():
             for v in value:
                 m.update(_digest_bytes(v))
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -201,7 +201,7 @@ def _digest_bytes(value: Any) -> bytes:
             )
         case Digest():
             # Must precede str (Digest ⊂ str)
-            m.update(value.encode())
+            return value.encode()
         case str():
             m.update(value.encode())
         case None:
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -71,9 +71,9 @@ _EP_HOOKS = []
 # Types confirmed *not* to define ``__digest__``.  ``__digest__`` is the opt-in
 # protocol that lets user types (including subclasses of dict/list/...) take
 # over their own digest, so the check has to run before the built-in ``match``
-# cases — which means it executes on every recursive ``_digest`` call for plain
-# ints, strs, dicts, lists, etc., where the answer is always False.  After the
-# first sighting of each built-in type, a set-membership check skips the
+# cases — which means it executes on every recursive ``_digest_bytes`` call for
+# plain ints, strs, dicts, lists, etc., where the answer is always False.  After
+# the first sighting of each built-in type, a set-membership check skips the
 # ``hasattr`` MRO walk entirely.  No semantic change for class-defined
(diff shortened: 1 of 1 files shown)


pmrv:
@claude resolve by rebase

pmrv:
**Claude finished @pmrv's task in 6m 13s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25522603648)

---
### Resolved via rebase

- [x] Fetch latest `origin/main`
- [x] Understand divergence (no shared ancestor — main was squash-reset to a single commit `5b836f4`)
- [x] Apply `_digest_bytes` refactoring onto `origin/main`'s `digest.py` (preserving `int`-first arm order, `staticmethod`/`classmethod`/`property`/`ModuleType` cases from `#497` base)
- [x] Committed as `9287434` and force-pushed

The branch now has a single commit on top of `origin/main` (`5b836f4`). The rebased `digest.py` carries over all structural choices from `#497` (arm ordering, `staticmethod`/`classmethod`/`property`/`ModuleType` arms) while applying the full `_digest_bytes` refactoring.
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-440-20260506-1533)

src/fleche/digest.py
```diff
@@ -135,49 +135,51 @@ def load_entry_points():
...
     )
-    for k_digest, k, v in sorted_items:
-        m.update(k_digest.encode())
-        m.update(digest(v).encode())
+    for k_bytes, k, v in sorted_items:
+        m.update(k_bytes)
+        m.update(_digest_bytes(v))
+    # To gain the raw-bytes speedup (Issue #440), change this to m.digest() in
+    # sync with the return at the end of _digest_bytes.
+    return m.hexdigest().encode()
```
pmrv: not needed anymore
```suggestion
```

src/fleche/digest.py
```diff
@@ -186,7 +188,7 @@ def _digest(value: Any) -> Digest:
     t = type(value)
     if t not in _TYPES_WITHOUT_DIGEST:
         if hasattr(t, "__digest__"):
-            return Digest(value.__digest__())
+            return value.__digest__().encode()
```
pmrv, line 190: make a note that this will require bytes.fromhex

src/fleche/digest.py
```diff
@@ -255,45 +257,48 @@ def _digest(value: Any) -> Digest:
...
+            m.update(_digest_bytes(dict(_attrs.field_items(value))))
         case _ if isinstance(value, types.ModuleType):
             names = getattr(value, "__all__", None)
             if names is None:
                 names = dir(value)
-            _digest_mapping(m, {name: getattr(value, name) for name in names})
+            return _digest_mapping(m, {name: getattr(value, name) for name in names})
         case Mapping():
-            _digest_mapping(m, value)
+            return _digest_mapping(m, dict(value))
```
pmrv: do not return from cases, keep original fall through behavior.  Audit the entire function again to ensure the final return is the only exit from this function, except delegating recursive calls.

pmrv (commented):
revise

pmrv:
> revise

@claude

pmrv:
**Claude finished @pmrv's task in 4m 0s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25567061017)

---
### Revised digest.py — all review comments addressed

- [x] Read `src/fleche/digest.py` and understand current state
- [x] Remove unneeded comment at lines 145–146 in `_digest_mapping` — migration note was already captured in `_digest_bytes` docstring
- [x] Add `bytes.fromhex` note at the `__digest__` early-return path — `# For raw-bytes speedup (Issue #440): bytes.fromhex(value.__digest__())`
- [x] Audit `match` block — remove early returns that aren't delegating recursive `_digest_bytes` calls:
  - **`Digest` case**: `return value.encode()` → `m.update(value.encode())` (fall through to final return)
  - **`ModuleType` case**: dropped `return` before `_digest_mapping(m, ...)` — `_digest_mapping` mutates `m` in place, so the final `return m.hexdigest().encode()` produces the same result
  - **`Mapping` case**: same — dropped `return` before `_digest_mapping(m, dict(value))`
  - Kept `return _digest_bytes(...)` in `Number` (non-NaN), `np.bool_`, and `FunctionType` cases — these are true delegating recursive calls
- [x] Committed and pushed (`92cc1aa`)
 | [Branch](https://github.com/pmrv/fleche/tree/claude/issue-440-20260506-1533)

src/fleche/digest.py
```diff
@@ -202,7 +201,7 @@ def _digest_bytes(value: Any) -> bytes:
             )
         case Digest():
             # Must precede str (Digest ⊂ str)
-            return value.encode()
+            m.update(value.encode())
```
pmrv: @claude Revert, pretty sure this breaks

pmrv: **Claude finished @pmrv's task in 1m 3s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25567978436)

---
### Reverted `Digest` case early return

- [x] Read `src/fleche/digest.py` and identify the change at line 204
- [x] Understand what needs to be reverted
- [x] Make the revert
- [x] Commit and push

The `Digest` case was changed in the last revision from `return value.encode()` to `m.update(value.encode())` (fall-through). You were right that this breaks things:

- A `Digest` is a terminal hash value — it should be passed through as-is.
- With the fall-through, it gets re-hashed as `SHA256("Digest" + value_bytes)` (because `m` already has the type name in it from line 193), producing wrong/inconsistent digests for any value that contains an already-computed `Digest`.

Reverted to `return value.encode()` in commit `4d012cf`. This is a legitimate non-recursive early return (it delegates to the value's own bytes directly), not a delegating recursive call, so it's correct to keep it as an early return.

[Branch](https://github.com/pmrv/fleche/tree/claude/issue-440-20260506-1533) |

PR #496, opened 2026-05-06
docs(configuration): fix SQL URL tilde expansion and simplify config lookup rules

pmrv:
### 1. `~` in SQL URLs is not expanded

The `url` key description and the HDF5+SQL full-config example both used:
```toml
calls.url = "sqlite:///~/.cache/fleche/calls.db"
```

Unlike the `root` key for file backends — which calls `Path.expanduser()` in `FileStorage.__post_init__` — SQL URLs are passed verbatim to SQLAlchemy via `_coerce_sqlite_url`. The `~` is treated as a **literal directory name** relative to the current working directory, not the home directory.

Verified locally:
```python
from fleche.storage.sql import _coerce_sqlite_url
_coerce_sqlite_url("sqlite:///~/.fleche/calls.db")
# → "sqlite:///~/.fleche/calls.db"  (unchanged; ~ not expanded)
```
And `Path("~/.fleche").mkdir(parents=True, exist_ok=True)` would create `./~/.fleche/` in the CWD, not `~/.fleche/`.

### 2. Redundant config-path rules

The global config lookup was listed as three rules:
- `$XDG_CONFIG_HOME` set → `$XDG_CONFIG_HOME/fleche/cache.toml`
- `$HOME` set → `$HOME/.fleche.toml`
- Otherwise → `~/.fleche.toml`

Cases 2 and 3 resolve to the same path on any normal system (`$HOME/.fleche.toml == ~/.fleche.toml`). The only difference is the Python fallback when `$HOME` is unset, which is an implementation detail not worth surfacing to users.

## Changes

- **`url` key**: Document that `~` is not expanded; recommend absolute paths; update the example to a placeholder absolute path (`sqlite:////absolute/path/to/calls.db`).
- **Global config lookup**: Collapse the three rules to two (XDG / otherwise).

---
_Generated by [Claude Code](https://claude.ai/code/session_01NNKmavMCzD63AWAJCyCNQr)_

### DIFF
--- a/docs/storage/configuration.rst
+++ b/docs/storage/configuration.rst
@@ -163,10 +163,8 @@ Key descriptions
     ``FLECHE_SECRET_KEY`` environment variable.
 
 ``url``
-    SQLAlchemy connection URL, e.g. ``"sqlite:////absolute/path/to/calls.db"``.
-    Unlike ``root``, ``~`` is **not** expanded in SQL URLs — always use an
-    absolute path for SQLite, or configure the cache programmatically when the
-    path must be derived from the home directory at runtime.
+    SQLAlchemy connection URL, e.g. ``"sqlite:///~/.cache/fleche/calls.db"``.
+    Leading ``~`` is expanded to the home directory in ``sqlite:///`` URLs.
 
 ``echo``
     (bool, default ``false``) — log all SQL statements to stderr (useful for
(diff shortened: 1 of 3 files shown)


pmrv:
@claude instead implement ~ expansion.

pmrv:
**Claude finished @pmrv's task in 2m 44s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25450585612)

---
### Implemented `~` expansion in SQL URLs

- [x] Gather context
- [x] Implement `~` expansion in `_coerce_sqlite_url`
- [x] Add tests for `~` expansion
- [x] Revert doc warning about `~` not being expanded, restore example URL
- [x] Commit and push changes

**What changed:**

`_coerce_sqlite_url` in `src/fleche/storage/sql.py` now calls `Path.expanduser()` on both bare paths and the path portion of `sqlite:///` URLs. Three new tests cover the expansion behaviour.

The docs now restore the original `sqlite:///~/.cache/fleche/calls.db` example and note that leading `~` is expanded, instead of warning users to avoid it.

The config path rule simplification (collapsing `$HOME`/otherwise to just "otherwise") from the original PR is retained.
 |

PR #500, opened 2026-05-07
fix(query): make QueryIterator re-iterable

pmrv:
## Summary

- `QueryIterator.calls` was a live generator object; consuming it once left nothing for a second traversal, silently returning empty results
- Change `calls` to a `Callable[[], Iterable[LazyCall]]` factory; `__iter__` now calls `self.calls()` to get a fresh generator on every iteration
- `caches.py` passes `_safe_iter` (the function) instead of `_safe_iter()` (the generator object)
- All lazy methods (`filter`, `take`, `skip`, `unique`, `sorted`, `groupby`) updated to wrap their generator expressions in lambdas that close over `self`, which is itself now re-iterable
- New test `test_query_iterator_is_re_iterable` verifies that iterating a `QueryIterator` twice yields the same results

## Test plan

- [ ] `pytest tests/unit/fleche/test_query_iterator.py` — all 58 tests pass
- [ ] `pytest tests/unit/ tests/integration/` — passes (pre-existing failure in `test_parallel_execution.py::test_process_executor_bound_wrapper` is unrelated, reproduces on `main` without these changes)

---
_Generated by [Claude Code](https://claude.ai/code/session_01WW1STsLJh7Amwi4CjVEfx8)_

### DIFF
--- a/src/fleche/query.py
+++ b/src/fleche/query.py
@@ -24,11 +24,11 @@ def _resolve_key(key: "str | Callable[[call.LazyCall], Any]") -> "Callable[[call
 class QueryIterator(Iterable[call.LazyCall]):
     """Re-iterable view over a lazy query result.
 
-    ``calls`` must be a zero-argument callable that returns a fresh iterable
-    each time it is invoked.  This makes ``QueryIterator`` itself re-iterable:
-    every ``for`` loop, ``list()``, or consuming method starts a new traversal
-    of the underlying source rather than silently seeing an empty sequence on
-    the second call.
+    ``calls`` is a zero-argument callable that returns a fresh iterable each
+    time it is invoked.  Every ``for`` loop, ``list()``, or consuming method
+    therefore starts a new traversal of the underlying source, so the same
+    ``QueryIterator`` can be used multiple times and will reflect the current
+    state of the cache on each pass.
 
     Args:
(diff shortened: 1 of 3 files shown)


src/fleche/query.py
```diff
@@ -22,15 +22,22 @@ def _resolve_key(key: "str | Callable[[call.LazyCall], Any]") -> "Callable[[call
...
 @dataclass(frozen=True)
 class QueryIterator(Iterable[call.LazyCall]):
-    """Iterator that adds some convenience to plain iterators over calls of query result.
+    """Re-iterable view over a lazy query result.
+
+    ``calls`` must be a zero-argument callable that returns a fresh iterable
+    each time it is invoked.  This makes ``QueryIterator`` itself re-iterable:
+    every ``for`` loop, ``list()``, or consuming method starts a new traversal
+    of the underlying source rather than silently seeing an empty sequence on
+    the second call.
```
pmrv: Do not use ahistoric formulations.  Write to describe the current behavior/functionality without reference to previous code.

tests/unit/fleche/test_query_iterator.py
pmrv, line 1: Add also tests where changes to the underlying cache are reflecting in the yielded calls, e.g. query(); add a new call to cache then traverse first query().

pmrv (commented):
revise

PR #475, opened 2026-05-04
fix: support digest() of types (classes themselves)

pmrv:
Fixes #469

Types (plain classes, dataclass classes, built-in types like int/str) previously raised Unhashable or AttributeError when passed to digest(). Adds an isinstance(value, type) match case before the dataclass case so all types are digested by their __module__ + __qualname__, giving stable unique digests without touching instance behaviour.

Also adds 8 regression tests.

Generated with [Claude Code](https://claude.ai/code)

PR #509, opened 2026-05-08
feat(digest): make dataclasses._FIELD_BASE instances digestible

pmrv:
Splits off the `dataclasses._FIELD_BASE` digest support from #475 into its own PR so it can be reviewed and merged in separately.

## Changes

- Adds an `isinstance(value, _DATACLASSES_FIELD_BASE)` match case in `_digest_bytes` that uses the singleton's `.name` attribute (`'_FIELD'`, `'_FIELD_CLASSVAR'`, `'_FIELD_INITVAR'`) as the stable digest key
- Accesses `_FIELD_BASE` via `getattr(dataclasses, "_FIELD_BASE", None)` at module level to avoid the `ty` type-checker `unresolved-attribute` error (the `None` fallback disables the match arm gracefully on any future Python that removes the private class)
- Unblocks digesting `dataclasses.Field` instances: their `_field_type` slot holds one of these singletons and was the last blocker for `digest(type) == digest(dict(type.__dict__))` on dataclass *classes*

## Tests

5 new tests:
- `test_field_base_singletons_are_digestible` — all three singletons return a `Digest`
- `test_field_base_singletons_have_distinct_digests` — they differ from each other
- `test_field_base_digest_is_stable` — same singleton → same digest
- `test_field_instance_is_digestible` — a full `Field` object can be digested
- `test_field_instance_digest_changes_with_value` — two fields for different names differ

Merges into #475.

### DIFF
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -307,6 +307,10 @@ def _digest_bytes(value: Any) -> bytes:
             m.update(_digest_bytes(dict(_attrs.field_items(value))))
         case _ if value is dataclasses.MISSING:
             m.update(b"__MISSING__")
+        case _ if isinstance(value, dataclasses._FIELD_BASE):
+            # _FIELD_BASE singletons (_FIELD, _FIELD_CLASSVAR, _FIELD_INITVAR) carry
+            # a stable .name attribute; use it as the digest key.
+            m.update(value.name.encode())
         case _ if isinstance(value, enum.Enum):
             return _digest_bytes((type(value).__qualname__, value.value))
         case _ if (
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -90,6 +90,11 @@ _C_DESCRIPTOR_TYPES = (
     types.MemberDescriptorType,
 )
 
+# dataclasses._FIELD_BASE is a private sentinel class (Python 3.10+).  Access via
+# getattr to avoid type-checker errors on an unresolved private attribute; None
+# disables the match arm gracefully if a future Python removes it.
+_DATACLASSES_FIELD_BASE: type | None = getattr(dataclasses, "_FIELD_BASE", None)
+
 
 def get_hooks():
     return list(reversed(_HOOKS)) + _EP_HOOKS
@@ -307,7 +312,7 @@ def _digest_bytes(value: Any) -> bytes:
             m.update(_digest_bytes(dict(_attrs.field_items(value))))
         case _ if value is dataclasses.MISSING:
             m.update(b"__MISSING__")
-        case _ if isinstance(value, dataclasses._FIELD_BASE):
+        case _ if _DATACLASSES_FIELD_BASE is not None and isinstance(value, _DATACLASSES_FIELD_BASE):
             # _FIELD_BASE singletons (_FIELD, _FIELD_CLASSVAR, _FIELD_INITVAR) carry
             # a stable .name attribute; use it as the digest key.
             m.update(value.name.encode())
--- a/tests/unit/digest/test_digest.py
+++ b/tests/unit/digest/test_digest.py
@@ -1016,3 +1016,54 @@ def test_type_digest_stable_across_cloudpickle_roundtrip():
     original = digest(cls)
     reloaded = cloudpickle.loads(cloudpickle.dumps(cls))
     assert digest(reloaded) == original
+
+
+# --- Tests for dataclasses._FIELD_BASE (Python 3.12: _FIELD / _FIELD_CLASSVAR / _FIELD_INITVAR) ---
+
+
+def test_field_base_singletons_are_digestible():
+    """The three _FIELD_BASE singletons must be digestible."""
+    import dataclasses as dc
+    assert isinstance(digest(dc._FIELD), Digest)
+    assert isinstance(digest(dc._FIELD_CLASSVAR), Digest)
+    assert isinstance(digest(dc._FIELD_INITVAR), Digest)
+
+
+def test_field_base_singletons_have_distinct_digests():
+    """The three singletons must produce different digests."""
+    import dataclasses as dc
+    assert digest(dc._FIELD) != digest(dc._FIELD_CLASSVAR)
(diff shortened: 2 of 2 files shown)


pmrv:
@claude `_FIELD_BASE` sounds scary.  Is there a public facing class, too?

pmrv:
**Claude finished @pmrv's task in 1m 43s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25575956390)

---
### Answer: `_FIELD_BASE` public equivalent

**No** — there's no public-facing class for `_FIELD_BASE`. It's an entirely private implementation detail of CPython's `dataclasses` module with no public API counterpart.

Here's the full picture from the Python 3.12 source:

```python
# Private sentinel class — not in __all__, not documented
class _FIELD_BASE:
    def __init__(self, name):
        self.name = name
    def __repr__(self):
        return self.name

_FIELD         = _FIELD_BASE('_FIELD')          # regular field
_FIELD_CLASSVAR = _FIELD_BASE('_FIELD_CLASSVAR')  # ClassVar annotation
_FIELD_INITVAR  = _FIELD_BASE('_FIELD_INITVAR')   # InitVar pseudo-field
```

The module's `__all__` exports (`dataclass`, `field`, `Field`, `fields`, `MISSING`, `InitVar`, `KW_ONLY`, …) contain nothing related to field-type categorisation. The `_field_type` attribute on `Field` objects is explicitly commented "Private: not to be used by user code."

**Why we still need to handle it:** `dataclasses.Field` uses `__slots__`, so the slots-based fallback in `_digest_bytes` tries to read all slots — including `_field_type`, which holds one of those three singletons. Without the new `isinstance(value, _DATACLASSES_FIELD_BASE)` arm, digesting a bare `Field` object raises `Unhashable`. The existing dataclass-class arm (`dataclasses.fields()`) sidesteps this for *class* digests, but a user can still pass a `Field` directly to `digest()`.

The `getattr(dataclasses, "_FIELD_BASE", None)` approach in this PR is the correct strategy — there's simply no public hook to use instead.

pmrv:
@claude In my testing `fl.D(F.__dict__)` runs fine if `F` is a dataclass.  What's the issue we're solving here?

pmrv:
**Claude finished @pmrv's task in 4m 35s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25612052772)

---
### Answer: what `fl.D(F.__dict__)` actually exercises

- [x] Read digest.py to understand the dataclass digest path
- [x] Check what `fl.D(F.__dict__)` actually does vs what this PR adds
- [x] Formulate answer

---

**`fl.D(F.__dict__)` succeeds or fails depending on whether `F` has any fields.**

Here's the traversal for a dataclass with fields:

```
fl.D(F.__dict__)
  → mappingproxy → Mapping() arm → _digest_mapping(dict(F.__dict__))
    → key '__dataclass_fields__', value {'x': Field(...)}
      → Field instance → slots arm (Field has __slots__, no __dict__)
        → builds {'default': MISSING, ..., '_field_type': _FIELD, 'name': 'x', ...}
          → _FIELD (_FIELD_BASE instance)
            → _FIELD_BASE HAS a __dict__ → skips the slots arm
            → not Mapping, not Iterable, ...
            → raise Unhashable  ← fails here
```

If you tested with a **fieldless** dataclass (`@dataclass class F: pass`), `F.__dataclass_fields__` is `{}`, no `Field` objects are traversed, and it runs fine. That's likely the discrepancy.

The existing comment at `digest.py:296` even documents this explicitly:
```python
case _ if isinstance(value, type) and dataclasses.is_dataclass(value):
    # digest dataclass classes by their field schema; dict(cls.__dict__)
    # contains Python-internal _FieldType sentinels that are not
    # independently digestible.
    return _digest_bytes([(f.name, f.type) for f in dataclasses.fields(value)])
```

The `isinstance(value, type)` guard ensures `fl.D(F)` never goes through `dict(F.__dict__)` — that special case exists precisely *because* `dict(F.__dict__)` was broken. This PR is what makes it work.

**So the PR solves two related things:**
1. `fl.D(dc.fields(F)[0])` — digesting a `Field` instance directly — was always broken for any field.
2. `fl.D(F.__dict__)` — was broken for any dataclass with at least one field (the `mappingproxy` path hits the same `Field`→`_field_type` chain).

pmrv:
> @claude In my testing `fl.D(F.__dict__)` runs fine if `F` is a dataclass. What's the issue we're solving here?

This is actually working because of the Mapping bug fixed in #483, `type.__dict__` is not a `dict` but a `mappingproxy`.  So not helpful

pmrv:
**Claude finished @pmrv's task in 1m 21s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25637995571)

---
### Clarification: dependency on #483

- [x] Read digest.py to understand the full picture
- [x] Formulate answer

---

That's the right diagnosis. The two PRs are coupled:

| State | `fl.D(F.__dict__)` behaviour |
|---|---|
| Neither #483 nor #509 merged | `mappingproxy` falls through to `Unhashable` (or is silently wrong) — never reaches `Field._field_type` |
| #483 merged, #509 not | `mappingproxy` is now traversed as a `Mapping` → recurses into `__dataclass_fields__` → hits `Field._field_type` → `_FIELD_BASE` instance → `Unhashable` |
| Both merged | `mappingproxy` traversed → `_FIELD_BASE` handled → works correctly |

So your earlier test was a false negative: `fl.D(F.__dict__)` appeared to succeed only because the `mappingproxy` path was silently broken first, and execution never reached the `_FIELD_BASE` problem that this PR fixes.

**What this PR actually unblocks** (given #483): digesting a dataclass's `__dict__` (or a `Field` object directly) without hitting `Unhashable` on the `_field_type` slot. The guard at `digest.py:295` (`isinstance(value, type) and dataclasses.is_dataclass(value)`) already short-circuits `fl.D(F)` via the field-schema path, so that route was never broken — only the raw `F.__dict__` / `Field` routes.

pmrv:
Too complicated for now.

PR #515, opened 2026-05-11
refactor: replace _intern_rec match ladder with dispatch table

### LINKED ISSUE #332: Refactor: turn `DestructuringMixin._intern_rec` type ladder into a dispatch table
## Location

`src/fleche/storage/destructuring.py:100-158`.

## Problem

`_intern_rec` uses a `match value:` ladder to decide how to destructure a value into a `Digested*` wrapper. Cases: `Number | str | bytes`, empty containers, namedtuple escape hatch (`_is_trojan_tuple`), `list | tuple`, `dict`. Adding a new container type (e.g. `frozenset`, `set`, third-party `pandas.Series`, `dataclass` — see #3) means editing the `match` body directly, which is an invasive change and rules out plugin-style extension.

The same concern applies to `put()` just above, which has a similar 3-case `match`.

## Suggested refactor

Mirror the `digest.py` hook mechanism: a list of `(type_predicate, destructure_fn)` pairs with a default "treat as opaque scalar" handler. Concrete example:

```python
DESTRUCTURERS: list[tuple[type, Callable]] = [
    ((list, tuple), DigestedIterable.sunder),
    ((dict,),       DigestedDict.sunder),
]
```

`_intern_rec` iterates these, returning on first match. Scalar / empty-container / namedtuple guards stay inline as they need to short-circuit *before* the dispatch.

A `register_destructurer(pred, fn)` hook at module scope lets #3 add dataclass support without modifying this file, and leaves room for user-supplied container types.

## Pros

- Aligns destructuring with the digest hook pattern, reducing cognitive load.
- Makes #3 (structural dedup for dataclasses) a pure additive change.
- Opens the door for library users to plug in custom container types.

## Cons / risks

- Moderate refactor; depth bookkeeping (`depth = max(...) + 1`) has to be preserved exactly — hypothesis tests on round-trip (`put(x); get(key) == x`) at varying `remaining_depth` should catch regressions.
- One more indirection when debugging storage layout.

pmrv:
Replaces the `match value:` type ladders in `_intern_rec` and `save` with a module-level `_DESTRUCTURERS` list of `(predicate, sunder_fn)` pairs. Adds `register_destructurer(pred, fn)` as the public extension point, mirroring `digest.add_hook`.

Closes #332

### DIFF
--- a/docs/dev/developer.rst
+++ b/docs/dev/developer.rst
@@ -32,6 +32,55 @@ they are handled correctly without special-casing at call sites.
 code then reads it from the profile instead of calling introspection APIs
 directly.
 
+.. _extending-destructurer:
+
+Extending the destructurer
+--------------------------
+
+The ``DestructuringMixin`` type-dispatch table, ``_DESTRUCTURERS``, is a
+module-level list of ``(predicate, sunder_fn)`` pairs.  It ships with four
+built-in entries — lists/tuples, dicts, dataclasses, and attrs instances — and
+can be extended at import time via
+:func:`~fleche.storage.destructuring.register_destructurer`.
+
+.. warning::
+
+   ``_DESTRUCTURERS`` is a global, mutable list.  Every
+   :class:`~fleche.storage.destructuring.DestructuringMixin` instance shares it.
+   Registering a destructurer after storage instances have already been used
+   produces undefined behaviour because ``_intern_rec`` closes over the list
+   state at call time.
+
+Before reaching for this function, consider whether implementing
+``__digest__`` (or ``add_hook``) on the type in question is sufficient.
+Destructurer registration is only necessary when a container type must have
+its *children* stored as independent, reusable keys rather than being pickled
+as a single opaque blob.
+
+**Contract for** ``sunder_fn``
+
+The function must have the signature ``(intern, value) -> (result, depth)``
+where:
+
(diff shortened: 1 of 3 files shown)


pmrv:
@claude document in the developer section, but not to invitingly.  This is not something to be customized lightly.

pmrv:
**Claude finished @pmrv's task in 1m 25s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25645894386)

---
### Adding Developer Documentation for `register_destructurer`

- [x] Read changed files to understand the API
- [x] Find the developer documentation section (`docs/dev/developer.rst`)
- [x] Add appropriately cautious documentation
- [x] Commit and push changes

Added a new **"Extending the destructurer"** section to `docs/dev/developer.rst`, placed after the `FunctionProfile` section. The section:

- Opens with a factual description of `_DESTRUCTURERS` and `register_destructurer`
- Leads immediately with a `.. warning::` block about the global mutable state and call-time ordering constraints
- Redirects readers toward `__digest__` / `add_hook` first before they reach for this
- Documents the exact `sunder_fn` contract (`(intern, value) -> (result, depth)` semantics, `Digested` subclass requirements) for the rare case it genuinely is needed
- Points to `DigestedIterable` / `DigestedDict` as canonical reference implementations

No examples, no "try this!" framing — just a technical specification with a clear warning at the top.
 |

PR #519, opened 2026-05-11
docs(configuration): fix remaining_depth description and destructuring backend list

pmrv:
## Summary

Two factual inaccuracies in `docs/storage/configuration.rst`, in the **Destructuring** section:

### 1. "All value backends except sql" — technically wrong

`ValueVoid` does **not** include `DestructuringMixin` and performs no element splitting (it discards all data anyway). The old claim:

> All value backends except `"sql"` … store collections … by *destructuring* them

was incorrect for `void`. The fix lists the backends that actually do destructuring explicitly, and notes that `void` is an exception.

### 2. `remaining_depth` description was misleading for nested collections

**Old text:**
> `remaining_depth = N` (positive) — elements at nesting levels shallower than *N* are stored inline…
> For example, `remaining_depth = 1` inlines scalars within their parent list or dict so each top-level collection is stored as a single entry.

**Problems:**
- "nesting levels shallower than N" is ambiguous. The code measures **depth from the deepest leaf** (scalars = 0; a container = `1 + max(children depths)`), not depth from the root.
- "each top-level collection is stored as a single entry" is **wrong** for nested collections. Verified:

```python
from fleche.storage.memory import ValueMemory

vm = ValueMemory(storage={}, remaining_depth=1)
vm.save([[1, 2], [3, 4]])
# → 3 storage entries:
#   one for [1, 2], one for [3, 4], one DigestedIterable referencing them
# NOT one entry for the outer list
```

Only a **flat** list/dict of pure scalars becomes a single entry with `remaining_depth=1`.

The TOML example comment `# inline scalars; one entry per top-level collection` repeated this incorrect claim and was also fixed.

## Test plan

- [x] Verified `remaining_depth=1` behavior for flat (`[1, 2, 3]`) and nested (`[[1, 2], [3, 4]]`) inputs using `ValueMemory`
- [x] Confirmed `ValueVoid` does not have `DestructuringMixin` (`isinstance(ValueVoid(), DestructuringMixin)` → `False`)

---
_Generated by [Claude Code](https://claude.ai/code/session_01TaPYuCmW5oMjWCznKRZvEn)_

### DIFF
--- a/docs/storage/configuration.rst
+++ b/docs/storage/configuration.rst
@@ -182,20 +182,25 @@ Key descriptions
 Destructuring
 ^^^^^^^^^^^^^
 
-All value backends except ``"sql"`` (which is call-only) store collections
-(:class:`list`, :class:`tuple`, :class:`dict`) by *destructuring* them: each element is
-stored independently under its own cache key, and on load the original structure is
-reassembled.  This avoids redundant storage of shared sub-structures across different
-cached calls.
+All persistent value backends (``"memory"``, ``"pickle"``, ``"cloudpickle"``,
+``"dill"``, ``"bagofholding_hdf"``) store collections (:class:`list`,
+:class:`tuple`, :class:`dict`) by *destructuring* them: each element is stored
+independently under its own cache key, and on load the original structure is
+reassembled.  This avoids redundant storage of shared sub-structures across
+different cached calls.  The ``"void"`` backend discards everything and performs
+no splitting.
 
 The optional ``remaining_depth`` key (integer, default ``0``) controls the granularity:
 
 * ``remaining_depth = 0`` — maximum splitting: every element at every nesting level is
   stored as a separate entry.
-* ``remaining_depth = N`` (positive) — elements at nesting levels shallower than *N* are
-  stored inline within their parent entry rather than as separate entries.
-  For example, ``remaining_depth = 1`` inlines scalars within their parent list or dict
-  so each top-level collection is stored as a single entry.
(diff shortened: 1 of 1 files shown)


pmrv:
> ### 1. "All value backends except sql" — technically wrong
> 
> `ValueVoid` does **not** include `DestructuringMixin` and performs no element splitting (it discards all data anyway). The old claim:
> 
> > All value backends except `"sql"` … store collections … by _destructuring_ them
> 
> was incorrect for `void`. The fix lists the backends that actually do destructuring explicitly, and notes that `void` is an exception.

Void is for special circumstances and testing, no need to burden average user with knowledge of its existance.

pmrv:
> > ### 1. "All value backends except sql" — technically wrong
> > `ValueVoid` does **not** include `DestructuringMixin` and performs no element splitting (it discards all data anyway). The old claim:
> > > All value backends except `"sql"` … store collections … by _destructuring_ them
> > 
> > 
> > was incorrect for `void`. The fix lists the backends that actually do destructuring explicitly, and notes that `void` is an exception.
> 
> Void is for special circumstances and testing, no need to burden average user with knowledge of its existance.

@claude

pmrv:
**Claude finished @pmrv's task in 1m 0s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25764025475)

---
### Tasks

- [x] Gather context
- [x] Understand request and read current file state
- [x] Simplify the backends description to avoid mentioning `void`
- [x] Commit and push changes

Removed the sentence about `void` from the destructuring section. The description now just lists the persistent backends that do destructuring without any mention of `void`.
 |

PR #537, opened 2026-05-18
docs(configuration): replace 'persistent' with 'non-void' in Destructuring section

pmrv:
## Summary

- `docs/storage/configuration.rst` line 185 called the `memory` backend a "persistent value backend" in the Destructuring sub-section.
- The same page (line 20, Reserved Cache Names) explicitly describes `memory` as a **transient** in-memory store that is lost when the process exits.
- Using "persistent" in one place and "transient" in another for the same backend is contradictory and misleading.

**Root cause:** The word "persistent" was intended to mean "data-retaining" (as opposed to the void backend which discards everything), but the established meaning in caching/storage contexts is "survives process restart" — which memory does not.

## What changed

Single word change in `docs/storage/configuration.rst`:

```diff
-All persistent value backends ("memory", "pickle", "cloudpickle",
+All non-void value backends ("memory", "pickle", "cloudpickle",
```

"Non-void" accurately describes the grouping (every value backend except `void`) without implying disk-persistence for `memory`.

## Test plan

- [ ] Read `docs/storage/configuration.rst` — "non-void" appears in the Destructuring section; "transient" (line 20) and "non-void" (line 185) are now consistent
- [ ] No code changes; no tests required

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kh7ygKRdjeS1rrG9L3hJe8)_

### DIFF
--- a/docs/storage/configuration.rst
+++ b/docs/storage/configuration.rst
@@ -182,7 +182,7 @@ Key descriptions
 Destructuring
 ^^^^^^^^^^^^^
 
-All persistent value backends (``"memory"``, ``"pickle"``, ``"cloudpickle"``,
+All non-void value backends (``"memory"``, ``"pickle"``, ``"cloudpickle"``,
 ``"dill"``, ``"bagofholding_hdf"``) store collections (:class:`list`,
 :class:`tuple`, :class:`dict`) by *destructuring* them: each element is stored
 independently under its own cache key, and on load the original structure is
--- a/docs/storage/configuration.rst
+++ b/docs/storage/configuration.rst
@@ -182,7 +182,7 @@ Key descriptions
 Destructuring
 ^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


pmrv:
Just make it 'most value backends',  users won't need to hear about void in 99% percent of cases.

pmrv:
> Just make it 'most value backends', users won't need to hear about void in 99% percent of cases.

@claude

pmrv:
**Claude finished @pmrv's task in 31s** —— [View job](https://github.com/pmrv/fleche/actions/runs/26053747383)

---
### Updating terminology

- [x] Understand the request
- [x] Read the file and make the change
- [x] Commit and push

Changed `non-void value backends` → `Most value backends` on line 185 of `docs/storage/configuration.rst`.

pmrv:
expand the sentence you're already touching to mentions attrs and dataclasses are supported too.

PR #542, opened 2026-05-18
fix: guard PerKeyLockMixin.expand() against concurrent put()/evict() races

### LINKED ISSUE #453: Cache race: PerKeyLockMixin.expand() lock scope too narrow

## Problem

`expand()` in `storage/base.py:108–117`, as called from `thread_safe.py:138–142`, acquires a per-key lock on the *prefix string* and then scans **all** keys:

```python
def expand(self, key: Digest | str) -> Digest:
    with self._operation_context(key):        # per-key lock on the PREFIX only
        ...
        candidates = sorted(k for k in self.list() if k.startswith(key))
        return _resolve_prefix(str(key), candidates[:2])
```

A concurrent `put()` on a *different* key (which acquires a different per-key lock and is therefore **not blocked**) can insert a new key sharing the same prefix between `self.list()` and `_resolve_prefix`.

Possible outcomes:
- `expand()` returns a stale (pre-insert) view — resolves the wrong key.
- Two concurrent expansions of the same prefix return different results.
- A spurious `AmbiguousDigestError` is raised if the newly-inserted key also matches the prefix.

## Affected backends

All `PerKeyLockMixin`-backed storages: `ValuePickleFile`, `CallPickleFile`, `ValueBagOfHoldingH5File`, `CallBagOfHoldingH5File`, `Sql`.  The issue does **not** affect `SerializingMixin` (global lock).

## Impact

Incorrect or non-deterministic prefix expansion under concurrent writes; spurious `AmbiguousDigestError`.

## Possible fix

Either:
- Use a global lock (or a read-write lock) for `expand()`, matching the behaviour of `SerializingMixin`.
- Or narrow the scope: after collecting `candidates` under the prefix lock, re-validate that no new matching key was inserted (e.g. via a second `list()` inside the lock and comparing lengths).

pmrv:
Fixes #453.

Introduces `_MutationScanLock`, an inverted readers-writer lock where mutations (`put`/`_evict`) are the concurrent side and a scan (`expand`) is the exclusive side. This avoids the performance cost of a global lock or a double scan: concurrent puts remain parallel; they only wait when an `expand()` scan is actively running.

### DIFF
--- a/src/fleche/storage/sql.py
+++ b/src/fleche/storage/sql.py
@@ -218,11 +218,11 @@ class Sql(PerKeyLockMixin, CallStorage):
             with super()._operation_context(key):
                 yield
 
-    def put(self, call: DigestedCall, key: Digest) -> Digest:
+    def put(self, value: DigestedCall, key: Digest) -> Digest:
         session = self._local.session
         existing = session.get(CallModel, str(key))
         if existing is not None:
-            if self.get(key) == call:
+            if self.get(key) == value:
                 return key
 
             session.delete(existing)
@@ -230,26 +230,26 @@ class Sql(PerKeyLockMixin, CallStorage):
 
         call_model = CallModel(
             key=str(key),
-            name=call.name,
-            module=call.module,
-            version=json.dumps(call.version) if call.version is not None else None,
-            code_digest=call.code_digest,
-            result=call.result if call.result is None else str(call.result),
+            name=value.name,
+            module=value.module,
+            version=json.dumps(value.version) if value.version is not None else None,
+            code_digest=value.code_digest,
+            result=value.result if value.result is None else str(value.result),
         )
         session.add(call_model)
 
-        for i, (k, v) in enumerate(call.arguments.items()):
+        for i, (k, v) in enumerate(value.arguments.items()):
             session.add(
                 ArgumentModel(
                     call_key=str(key), position=i, name=str(k), value=str(v)
                 )
             )
 
-        if call.metadata:
+        if value.metadata:
             session.add_all(
                 [
                     MetaModel(call_key=str(key), name=name, data=data)
-                    for name, data in call.metadata.items()
+                    for name, data in value.metadata.items()
                 ]
             )
         session.commit()
--- a/src/fleche/storage/thread_safe.py
+++ b/src/fleche/storage/thread_safe.py
@@ -23,12 +23,23 @@ Choose the mixin based on the access pattern you need:
 
 Both mixins use reentrant locks so nested acquisitions (e.g. ``expand`` being
 called inside ``load``) do not deadlock.
+
+``PerKeyLockMixin`` additionally guards ``expand()`` against concurrent
+``put()`` / ``_evict()`` calls via a :class:`_MutationScanLock`.  This
+prevents a concurrent insert from creating a spurious
+:exc:`~fleche.storage.base.AmbiguousDigestError` inside the full-key-set scan
+that ``expand()`` performs.  The lock is an inverted readers-writer lock:
+mutations are the *concurrent* side (multiple puts may proceed in parallel)
+while a scan is the *exclusive* side (blocks new mutations and waits for
+in-flight ones to drain).  ``get`` / ``load`` / ``contains`` never touch this
+lock, so read throughput is unaffected.
 """
 
 import contextlib
 import threading
 import weakref
 from dataclasses import dataclass, field
+from typing import Any
 
 from .base import KeyManagement
 from ..digest import Digest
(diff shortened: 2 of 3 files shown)


pmrv:
@claude fix ty

pmrv:
**Claude finished @pmrv's task in 58s** —— [View job](https://github.com/pmrv/fleche/actions/runs/26053864786)

---
### Fix applied

- [x] Understand the request
- [x] Read `sql.py` to locate the parameter mismatch
- [x] Apply fix: renamed `call` → `value` in `SQLStorage.put()` signature and all usages within the method body (`src/fleche/storage/sql.py:221`)
- [x] Commit and push

Renamed the parameter throughout the method body as well (lines 225, 233–237, 241, 248, 252) so nothing was left referencing the old name.

pmrv:
@claude Curious, it should never check a `PerKeyLockMixin` for a put method.  Research affected classed its MRO.

pmrv:
**Claude finished @pmrv's task in 1m 37s** —— [View job](https://github.com/pmrv/fleche/actions/runs/26167758185)

---
### MRO Analysis for `PerKeyLockMixin.put()` Type Error

**Tasks**
- [x] Read `thread_safe.py` to understand `PerKeyLockMixin` hierarchy
- [x] Read `base.py` to map the full class tree
- [x] Trace the MRO and explain why the type checker fails

---

### Root cause

`PerKeyLockMixin` inherits from `KeyManagement` (`thread_safe.py:180`):

```python
class PerKeyLockMixin(KeyManagement): ...
```

Its static MRO, seen in isolation, is:

```
PerKeyLockMixin → KeyManagement → ABC → object
```

`KeyManagement` does **not** define `put`. `put` lives one level down, in `StorageBackend`:

```
KeyManagement          (expand, evict, contains, shrink — no put)
    └── StorageBackend (abstract put, abstract get)
            ├── ValueMixin
            └── CallMixin
```

When the type checker evaluates `super().put(value, key)` inside `PerKeyLockMixin.put()` (`thread_safe.py:250`), it resolves `super()` against the **static** parent — `KeyManagement` — not against the full MRO of whatever concrete class will eventually mix `PerKeyLockMixin` in. `KeyManagement` has no `put`, so the checker raises `unresolved-attribute`.

At **runtime** this works fine because concrete users look like:

```python
class PerKeyValuePickle(PerKeyLockMixin, ValuePickleFile): ...
# MRO: PerKeyValuePickle → PerKeyLockMixin → ValuePickleFile → ValueMixin → StorageBackend → KeyManagement
```

Here `super()` inside `PerKeyLockMixin.put()` does reach `StorageBackend.put()` through the full MRO. But the type checker never sees that concrete class when it checks `PerKeyLockMixin` — it only sees `KeyManagement` as the next in line.

### Why it didn't fail before this PR

`SerializingMixin` (same parent, `KeyManagement`) never overrides `put`, so the checker never needs to resolve `super().put()` inside it. This PR introduced a `put` override in `PerKeyLockMixin` for the first time, exposing the gap.

### Fix

Change `PerKeyLockMixin`'s declared parent from `KeyManagement` to `StorageBackend` (`thread_safe.py:180`):

```python
# before
class PerKeyLockMixin(KeyManagement):

# after
class PerKeyLockMixin(StorageBackend):
```

With this change the static MRO is:

```
PerKeyLockMixin → StorageBackend → KeyManagement → ABC → object
```

`super().put()` now resolves to `StorageBackend.put()` (abstract), which satisfies the type checker. The class remains abstract (both `put`/`get` are still abstract), which is correct for a mixin. Runtime behaviour is unchanged because the concrete subclass MRO is identical.

Note: `_evict` (`thread_safe.py:252`) calls `super()._evict()` — `_evict` *is* abstract on `KeyManagement`, so that call is already type-safe. Only `put` is the problem. [Fix this →](https://claude.ai/code?q=In%20src%2Ffleche%2Fstorage%2Fthread_safe.py%2C%20on%20branch%20claude%2Fissue-453-20260513-1618%2C%20change%20line%20180%20from%20%60class%20PerKeyLockMixin(KeyManagement)%3A%60%20to%20%60class%20PerKeyLockMixin(StorageBackend)%3A%60%20and%20add%20StorageBackend%20to%20the%20imports%20from%20.base%20if%20needed.)

---

pmrv:
@claude Cherry pick 25185c14cbcdbb4ccf6558fd62cb9cb5dd4dd1b4 onto a new branch off main and create new PR.

pmrv:
**Claude finished @pmrv's task in 1m 0s** —— [View job](https://github.com/pmrv/fleche/actions/runs/26167860230)

---
### Cherry-pick commit to new branch

- [x] Understand the request
- [x] Create new branch off `main` → `fix/sql-put-parameter-rename`
- [x] Cherry-pick `25185c1` onto new branch
- [x] Push new branch
- [x] Create PR

Commit `25185c1` (rename `call` → `value` in `SQLStorage.put()`) has been cherry-picked onto a new branch off `main` and a PR created.

**New PR:** https://github.com/pmrv/fleche/pull/546

pmrv:
Adding storage backend methods to the mixins is definitely the wrong track.

PR #551, opened 2026-05-22
feat(remote): SshCache for sharing fleche caches across machines

pmrv:
## Summary

Adds SshCache(BaseCache) — a cache that forwards every operation to a remote python -m fleche.remote --serve process over a single persistent SSH subprocess. The remote loads its own fleche.toml and proxies into whichever cache it has configured, so the remote side keeps full freedom of backend (file / SQL / HDF5 / stack). Compose it with a local layer via the existing CacheStack machinery for read-through sharing across machines.

## Usage

```
[default]
cache = "shared"

[[shared]]                          # local layer (saves go here)
values.type = "cloudpickle"
values.root = "~/.fleche/values"
calls.type = "sql"
calls.url = "sqlite:///~/.fleche/calls.db"

[[shared]]                          # remote layer (read-through)
type = "ssh"
host = "user@bigpc.example.com"
cache_name = "shared"               # optional: named cache on remote
python = "python3"                  # optional
ssh_options = ["-o", "ControlMaster=auto",
               "-o", "ControlPath=~/.ssh/cm-%r@%h:%p",
               "-o", "ControlPersist=10m"]
setup_commands = ["module load python/3.11",
                  "source ~/.venv/bin/activate"]   # optional
```

Reads check the local layer first, fall through to the SSH layer, and back-fill hits into local — all through CacheStack, no special-casing. Requires pip install fleche[ssh] (cloudpickle is the wire format, not optional for this feature).
## How it works

    Wire protocol. 4-byte big-endian length prefix + cloudpickle payload. Requests are (method, args, kwargs); responses are ("ok", value) / ("err", exception), so server-side exceptions (KeyError, Rejected, …) re-raise on the client with their real type. An exception that can't be cloudpickled is downgraded to a RuntimeError carrying the traceback.
    One method = one round-trip, serialised behind a per-connection lock. query materialises the server-side generator into a tuple before sending. Each returned DigestedCall is re-bound to the client SshCache as a LazyCall, so values round-trip lazily on first .result access. shrink follows the variadic BaseCache contract (single key → Digest, multiple → tuple).
    Same module is client and server. python -m fleche.remote --serve [--cache NAME] runs the request loop against the active cache; --cache NAME is installed as the active cache (ContextVar), not just loaded, so metadata hooks / nested @fleche() calls on the remote see the same cache.

## Lifecycle & operability

    Subprocess spawned lazily on first op, reused for the process lifetime, closed at atexit. Never auto-reconnects (would silently re-trigger interactive 2FA); reconnect() forces a fresh spawn. ControlMaster + ControlPersist share one authentication across runs.
    setup_commands run pre-launch in the remote shell, &&-joined and exec-ed into the server so the pipe stays unwrapped; a setup failure surfaces via captured stderr.
    Remote stderr is drained by a daemon thread into the per-host fleche.remote.host.<host> logger and a 50-line ring buffer; on connection drop the buffer + exit code are surfaced in RemoteConnectionError, which also names reconnect() directly.

## Security model

The wire format is cloudpickle in both directions, so a hostile endpoint can execute code on the other — the trust boundary is exactly ssh user@host. Don't point SshCache at a host you wouldn't ssh into, and don't expose --serve on any transport other than the spawned ssh stdin/stdout (the RPC stream itself is unauthenticated). The info RPC redacts credentials (secret_key, URL passwords) before they go on the wire, since the client's DEBUG tracing logs full payloads.
Diagnostics

    info() reports the served cache config (credentials redacted), cache_name, read_only, fleche_version, cloudpickle_version, cwd, hostname, python, pid — one round-trip to answer "the remote isn't doing what I expected."
    First op runs a version handshake that warns on fleche / cloudpickle skew.
    read_only short-circuits save/evict locally to avoid a pointless round-trip.
    Every RPC is logged at DEBUG (rpc → … / rpc ← …) on the fleche.remote logger.

## Config integration

Single type = "ssh" clause in cache_from_config / cache_to_config; the array-of-tables form already gives stack composition. New [ssh] extra pulls in cloudpickle.
Testing

    Unit ([tests/unit/test_remote.py](https://github.com/pmrv/fleche/blob/claude/affectionate-pasteur-fYOy1/tests/unit/test_remote.py)): in-process os.pipe() transport runs serve() in a daemon thread — full wire protocol, dispatch, exception propagation, lifecycle, info redaction, version handshake, read-only short-circuit, variadic shrink, connection-drop diagnostics. No SSH or subprocess.
    Integration ([tests/integration/test_remote.py](https://github.com/pmrv/fleche/blob/claude/affectionate-pasteur-fYOy1/tests/integration/test_remote.py)): real python -m fleche.remote --serve subprocess (still no SSH) — cross-process round-trips, --cache NAME selection.

## Commits

    fix(types): resolve latent ty 0.0.40 errors surfaced by this branch — two pre-existing main type errors that the path-gated ty job let through on dependabot bumps (see #587); this branch is the first to touch src/ under ty 0.0.40.
    feat(remote): SshCache for sharing fleche caches across machines
    docs(remote): SshCache usage and internals

## Known limitations / follow-ups

    query() materialises the full result set on both ends — no streaming/cursor; large remote caches pay it all up front.
    Dropped SshCache instances aren't closed on GC (atexit holds a ref); call close() explicitly for caches you won't reuse.
    info() / read_only cost one RPC on first access (cached after; reconnect() invalidates).
    The per-connection lock serialises a slow query() ahead of other threads' RPCs.
    _strip_cache reaches into LazyCall internals → tracked in #588 (promote to a public LazyCall.to_digested_call()).

### DIFF
--- a/src/fleche/config.py
+++ b/src/fleche/config.py
@@ -113,6 +113,28 @@ Example fleche.toml
     calls.type = "cloudpickle"
     calls.root = "~/.fleche/calls"
 
+    # SshCache — share results with another machine over SSH.  The remote
+    # runs `python -m fleche.remote --serve` and proxies into its own
+    # configured cache.  Compose with a local cache by stacking two
+    # entries (saves go to the first entry; reads fall back to the SSH
+    # remote and back-fill hits into the local layer).
+    [[shared]]
+    values.type = "cloudpickle"
+    values.root = "~/.fleche/values"
+    calls.type = "cloudpickle"
+    calls.root = "~/.fleche/calls"
+
+    [[shared]]
+    type = "ssh"
+    host = "user@bigpc.example.com"
+    cache_name = "shared"               # optional: named cache on remote
+    python = "python3"                  # optional: remote python interpreter
+    ssh_options = ["-o", "ControlMaster=auto",
+                   "-o", "ControlPath=~/.ssh/cm-%r@%h:%p",
+                   "-o", "ControlPersist=10m"]
+    setup_commands = ["module load python/3.11",  # optional: shell snippets
+                      "source ~/.venv/bin/activate"]  # run before the server
+
 Config file discovery
 ---------------------
 
@@ -139,6 +161,7 @@ import os
 from typing import Any
 
 from . import storage, metadata, caches
+from .remote import SshCache
 
 logger = logging.getLogger("fleche.config")
 
@@ -398,6 +421,9 @@ def cache_from_config(d: "dict[str, Any] | list[dict[str, Any]]") -> caches.Base
         return caches.CacheStack(tuple(cache_from_config(c) for c in d))
 
     d = dict(d)
+    if d.get("type") == "ssh":
+        d.pop("type")
+        return SshCache(**d)
     read_only = d.pop("read_only", False)
     max_size = d.pop("max_size", None)
 
@@ -456,6 +482,17 @@ def cache_to_config(c: caches.BaseCache) -> "dict[str, Any] | list[dict[str, Any
         case caches.CacheStack():
             return cast("list[dict[str, Any]]", [cache_to_config(s) for s in c.stack])
         case _:
+            if isinstance(c, SshCache):
+                d: dict[str, Any] = {"type": "ssh", "host": c.host}
+                if c.cache_name is not None:
+                    d["cache_name"] = c.cache_name
+                if c.python != "python3":
+                    d["python"] = c.python
+                if c.ssh_options:
+                    d["ssh_options"] = list(c.ssh_options)
+                if c.setup_commands:
+                    d["setup_commands"] = list(c.setup_commands)
+                return d
             raise ValueError(f"Cannot convert cache of type {type(c).__name__!r} to config")
 
 
--- a/src/fleche/remote.py
+++ b/src/fleche/remote.py
@@ -1,0 +1,868 @@
+"""SSH-connected cache for sharing fleche results across machines.
+
+:class:`SshCache` is a :class:`~fleche.caches.BaseCache` that forwards every
+operation to a remote ``python -m fleche.remote`` process over a single
+persistent SSH subprocess.  The remote process loads its own ``fleche.toml``
+and proxies operations into whichever cache it has configured, so the remote
+side keeps full freedom to use any backend (file / SQL / HDF5 / stack).
+
+Typical configuration in ``fleche.toml`` — local cache first, remote cache
+second, composed automatically into a :class:`~fleche.caches.CacheStack`::
+
+    [default]
+    cache = "shared"
+
+    [[shared]]                          # local layer (saves go here)
+    values.type = "cloudpickle"
+    values.root = "~/.fleche/values"
+    calls.type = "sql"
+    calls.url = "sqlite:///~/.fleche/calls.db"
+
+    [[shared]]                          # remote layer (read-through)
+    type = "ssh"
+    host = "marvin@bigpc.example.com"
+    cache_name = "shared"               # optional: named cache on remote
+    python = "python3"                  # optional
+    ssh_options = ["-o", "ControlMaster=auto",
+                   "-o", "ControlPath=~/.ssh/cm-%r@%h:%p",
+                   "-o", "ControlPersist=10m"]
+    setup_commands = ["module load python/3.11",
+                      "source ~/.venv/bin/activate"]    # optional
+
+ControlMaster + ControlPersist in ``~/.ssh/config`` (or via ``ssh_options``)
+mean 2FA is prompted once and then re-used across multiple fleche sessions
+within the persist window.  The SSH subprocess is spawned lazily on the first
+cache operation and reused for the lifetime of the Python process, so within
+a single run only one authentication round-trip is required.
+"""
+
+import abc
+import argparse
+import atexit
+import collections
+import logging
+import os
+import shlex
+import socket
+import struct
+import subprocess
+import sys
+import threading
+import traceback
+from dataclasses import dataclass, field
+from typing import Any, Iterable, overload
+
+from pyiron_snippets.import_alarm import ImportAlarm
+
+from . import call as _call
+from .caches import BaseCache, Rejected
+from .call import Call, DigestedCall, LazyCall, QueryCall
+from .digest import Digest
+
+logger = logging.getLogger("fleche.remote")
+
+with ImportAlarm(
+    "SshCache requires 'cloudpickle' (used for the client/server wire "
+    "protocol). Install it with `pip install fleche[ssh]`.",
+    raise_exception=True,
+) as cloudpickle_alarm:
+    import cloudpickle
+
+
+# ---------------------------------------------------------------------------
+# Wire protocol
+# ---------------------------------------------------------------------------
+
+_HEADER = struct.Struct(">I")
+
+
+def _write_frame(stream, obj: Any) -> None:
+    """Write a length-prefixed cloudpickle frame to *stream*."""
+    data = cloudpickle.dumps(obj)
+    stream.write(_HEADER.pack(len(data)))
+    stream.write(data)
+    stream.flush()
+
+
+def _read_frame(stream) -> Any:
+    """Read a length-prefixed cloudpickle frame from *stream*.
+
+    Raises:
+        EOFError: stream closed before a full frame was available.
+    """
+    header = stream.read(_HEADER.size)
+    if len(header) < _HEADER.size:
+        raise EOFError("connection closed before frame header")
+    (n,) = _HEADER.unpack(header)
+    chunks = []
+    remaining = n
+    while remaining > 0:
+        chunk = stream.read(remaining)
+        if not chunk:
+            raise EOFError("connection closed mid-frame")
+        chunks.append(chunk)
+        remaining -= len(chunk)
+    return cloudpickle.loads(b"".join(chunks))
+
+
+# ---------------------------------------------------------------------------
+# Server side
+# ---------------------------------------------------------------------------
+
+
+def _strip_cache(lc: LazyCall) -> DigestedCall:
+    """Return the cache-free :class:`DigestedCall` shadow of a :class:`LazyCall`.
+
+    The remote-bound ``_cache`` reference inside :class:`LazyCall` cannot
+    travel across the wire — the client wraps the returned ``DigestedCall``
+    back into a :class:`LazyCall` bound to itself via ``DigestedCall.fetch``.
+    """
+    return DigestedCall(
+        name=lc.name,
+        arguments=dict(lc._arguments),
+        result=lc._result,
+        metadata=dict(lc.metadata),
+        module=lc.module,
+        version=lc.version,
+        code_digest=lc.code_digest,
+    )
+
+
+def _dispatch(cache: BaseCache, method: str, args: tuple, kwargs: dict) -> Any:
+    if method == "save":
+        (c,) = args
+        return cache.save(c)
+    if method == "load":
+        (key,) = args
+        return _strip_cache(cache.load(key))
+    if method == "load_value":
+        (key,) = args
+        return cache.load_value(key)
+    if method == "evict":
+        (key,) = args
+        cache.evict(key)
+        return None
+    if method == "contains":
+        (key,) = args
+        return cache.contains(key)
+    if method == "expand":
+        (key,) = args
+        return cache.expand(key)
+    if method == "shrink":
+        # Variadic: single key → Digest, multiple keys → tuple[Digest, ...].
+        return cache.shrink(*args)
+    if method == "query":
+        (template,) = args
+        return tuple(_strip_cache(lc) for lc in cache.query(template))
+    raise ValueError(f"Unknown remote cache method: {method!r}")
+
+
+def _is_read_only(cache: BaseCache) -> bool:
+    """Best-effort check of whether *cache* will reject every ``save``.
+
+    Recognises :class:`~fleche.caches.ReadOnlyMixin` directly (covers
+    :class:`~fleche.caches.ReadOnlyCache` and
+    :class:`~fleche.caches.FilteredCache`); for a
+    :class:`~fleche.caches.CacheStack` checks ``stack[0]`` since that is
+    where ``CacheStack.save`` writes.  Returns ``False`` for anything
+    else — at worst the client falls back to the round-trip path.
+    """
+    from .caches import CacheStack, ReadOnlyMixin
+
+    if isinstance(cache, ReadOnlyMixin):
+        return True
+    if isinstance(cache, CacheStack):
+        return bool(cache.stack) and _is_read_only(cache.stack[0])
+    return False
+
+
+# Config keys whose values are credentials we must not put on the wire.
+# ``secret_key`` carries HMAC signing keys; ``url`` may carry a database
+# password in its userinfo component (postgres / mysql).
+_SENSITIVE_CONFIG_KEYS = frozenset({"secret_key"})
+
+
+def _redact_url_password(url: str) -> str:
+    """Replace the password component of a SQLAlchemy URL with ``"***"``.
+
+    Plain sqlite URLs (no userinfo) round-trip unchanged.  Anything that
+    doesn't parse as a URL with a userinfo component is returned as-is.
+    """
+    try:
+        import urllib.parse as _up
+
+        parsed = _up.urlsplit(url)
+    except Exception:
+        return url
+    if not parsed.password:
+        return url
+    user = parsed.username or ""
+    host_and_port = parsed.hostname or ""
+    if parsed.port is not None:
+        host_and_port = f"{host_and_port}:{parsed.port}"
+    netloc = f"{user}:***@{host_and_port}" if user else f":***@{host_and_port}"
+    return _up.urlunsplit(
+        (parsed.scheme, netloc, parsed.path, parsed.query, parsed.fragment)
+    )
+
+
+def _redact_config(value: Any) -> Any:
+    """Recursively redact credential-ish fields from a cache_to_config dict.
+
+    Walks dicts and lists; for any dict whose key is in
+    :data:`_SENSITIVE_CONFIG_KEYS`, replaces the value with
+    ``"<redacted>"``.  For ``url`` keys whose value is a string with a
+    password component, the password component alone is masked so the
+    rest of the URL still appears in info output.
+    """
+    if isinstance(value, dict):
+        out: dict[str, Any] = {}
+        for k, v in value.items():
+            if k in _SENSITIVE_CONFIG_KEYS:
+                out[k] = "<redacted>"
+            elif k == "url" and isinstance(v, str):
+                out[k] = _redact_url_password(v)
+            else:
+                out[k] = _redact_config(v)
+        return out
+    if isinstance(value, list):
+        return [_redact_config(item) for item in value]
+    return value
+
+
+def _server_info(cache: BaseCache, cache_name: str | None) -> dict[str, Any]:
+    """Diagnostic snapshot of the server's view of itself.
+
+    Surfaces the same things a user would `ssh host` to check by hand: which
+    Python is running, in which directory, with what cache loaded.  The
+    served cache config plus the CWD is usually enough to spot the common
+    "remote loaded a different fleche.toml than I expected" failure mode.
+
+    Credential fields (``secret_key``, URL passwords) are redacted by
+    :func:`_redact_config` before going on the wire — the info dict is
+    visible to anyone the client logs against, including the DEBUG
+    ``rpc ←`` lines, and signing keys must not leak there.
+
+    ``fleche_version`` and ``cloudpickle_version`` are exposed so the
+    client can fail fast (or warn) on version skew with the server.
+    The ``read_only`` flag lets the client short-circuit ``save`` and
+    ``evict`` locally instead of paying a round-trip just to receive
+    :class:`~fleche.caches.Rejected`.
+    """
+    from .config import cache_to_config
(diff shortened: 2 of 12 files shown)


pmrv:
We also need an option to run some setup commands on the remote side before attempting to run the remote process.

pmrv:
manual testing trigger this error

```
>>> fl.cache('gl')
<fleche.state._StickyContext object at 0x7f2cd99ffee0>
>>> fl.cache().table()
Traceback (most recent call last):
  File "/home/ponder/projects/fleche/src/fleche/remote.py", line 245, in call
    tag, payload = _read_frame(self._stdout)
                   ~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/ponder/projects/fleche/src/fleche/remote.py", line 92, in _read_frame
    raise EOFError("connection closed before frame header")
EOFError: connection closed before frame header

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    fl.cache().table()
    ~~~~~~~~~~~~~~~~^^
  File "/home/ponder/projects/fleche/src/fleche/caches.py", line 197, in table
    return self.query(tpl).table(arguments=arguments, results=results)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ponder/projects/fleche/src/fleche/query.py", line 232, in table
    for c in self:
             ^^^^
  File "/home/ponder/projects/fleche/src/fleche/query.py", line 43, in __iter__
    yield from self.calls()
  File "/home/ponder/projects/fleche/src/fleche/caches.py", line 156, in _safe_iter
    yield from self._query(template)
  File "/home/ponder/projects/fleche/src/fleche/remote.py", line 465, in _query
    results: tuple[DigestedCall, ...] = self._conn.call("query", call)
                                        ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/ponder/projects/fleche/src/fleche/remote.py", line 248, in call
    raise RemoteConnectionError(
        f"Remote cache connection lost during {method!r}: {e}"
    ) from e
fleche.remote.RemoteConnectionError: Remote cache connection lost during 'query': connection closed before frame header
```

pmrv:
cloudpickle becomes required now, this needs to be more prominent

pmrv:
doesn't crash anymore, but table comes back empty, need debug tools

pmrv:
cache_name is not actually used on the remote end.  It should be used to select the active cache by with `fleche.cache(cache_name)`.

tests/unit/test_remote.py
```diff
@@ -294,6 +294,9 @@ def test_info_round_trip(remote, server_cache):
     assert info["pid"] == os.getpid()
     # cache_name is None in the test rig (the in-process serve() takes no arg).
     assert info["cache_name"] is None
```
pmrv, line 365: sounds like this is not a great test then

tests/unit/test_remote.py
```diff
@@ -294,6 +294,9 @@ def test_info_round_trip(remote, server_cache):
...
+    # The in-process pipe rig doesn't go through `_main`, so the test
+    # process's own active cache is unrelated to server_cache.
+    assert info["active_is_served"] is False
```
pmrv: confirmed that it works, can be removed again

src/fleche/config.py
```diff
@@ -393,6 +415,10 @@ def cache_from_config(d: "dict[str, Any] | list[dict[str, Any]]") -> caches.Base
         return caches.CacheStack(tuple(cache_from_config(c) for c in d))
 
     d = dict(d)
+    if d.get("type") == "ssh":
+        from .remote import SshCache
```
pmrv: top level

src/fleche/remote.py
```diff
@@ -0,0 +1,631 @@
...
+    Surfaces the same things a user would `ssh host` to check by hand: which
+    Python is running, in which directory, with what cache loaded.  The
+    repr of the active cache plus the CWD is usually enough to spot the
+    common "remote loaded a different fleche.toml than I expected"
+    failure mode.
+    """
+    from . import cache as get_active_cache
+
+    return {
+        "repr": repr(cache),
```
pmrv: cache def roundtrips to dict so lets do that here to avoid sending endlessly long string

src/fleche/remote.py
```diff
@@ -0,0 +1,631 @@
...
+# ---------------------------------------------------------------------------
+# Client side
+# ---------------------------------------------------------------------------
+
+
+class RemoteConnectionError(RuntimeError):
+    """Raised when the SSH subprocess cannot be reached or has died."""
+
+
+class _Connection:
```
pmrv: make abc

src/fleche/remote.py
```diff
@@ -0,0 +1,631 @@
...
+    def _close(self) -> None:
+        """Release any transport-level resources."""
+        return None
+
+    def _diagnose(self) -> str:
+        """Return diagnostic detail to append to a connection-lost error.
+
+        Subclasses with access to subprocess state (exit code, captured
+        stderr) override this so the user sees the *cause* — typically a
+        failed setup command — rather than just a bare EOF.
```
pmrv: sounds like you have a wealth experience in running this, which you don't

src/fleche/remote.py
```diff
@@ -0,0 +1,631 @@
...
+
+    # ------------------------------------------------------------------
+    # public API
+    # ------------------------------------------------------------------
+
+    def call(self, method: str, *args: Any, **kwargs: Any) -> Any:
+        with self._lock:
+            self._ensure_open()
+            if logger.isEnabledFor(logging.DEBUG):
+                logger.debug("rpc → %s args=%r kwargs=%r", method, args, kwargs)
```
pmrv: explain what's the advantage over just calling debug

src/fleche/remote.py
```diff
@@ -0,0 +1,631 @@
...
+        self._opened = False
+        try:
+            self._close()
+        except Exception:
+            logger.debug("error while closing remote connection", exc_info=True)
+        self._stdin = None
+        self._stdout = None
+
+
+class _SshConnection(_Connection):
```
pmrv, line 471: precarious little documentation on architecture, needs longer docstring

tests/unit/test_remote.py
pmrv: Agreed — dropped `active_is_served` and both tests in 654c735. Activation is verified end-to-end and the redundant signal isn't worth carrying.

---

tests/unit/test_remote.py
pmrv: Removed in 654c735 — the `active_is_served` field is gone, along with the unit assertion and the dedicated integration test that referenced it.

---

src/fleche/config.py
pmrv: Moved both `from .remote import SshCache` imports in `config.py` to module level in 654c735. No cycle: `remote.py` only imports caches/call/digest at module load time.

---

src/fleche/remote.py
pmrv: Done in 654c735 — `_server_info` now serialises through `cache_to_config(cache)` and exposes the dict under the `cache` key, so the payload is structured and round-trippable through `cache_from_config` instead of a long opaque repr.

---

pmrv (commented):
we should introspect whether the remote cache is read_only and expose this on the local end to avoid unnecessary roundtrips

src/fleche/remote.py
pmrv: Done in 654c735 — `_Connection` now inherits `abc.ABC` with `_open` as the single `@abstractmethod`. `_close` and `_diagnose` stay as overridable defaults with sensible no-ops.

---

src/fleche/remote.py
pmrv: Fair. Softened the `_diagnose` docstring in 654c735 — it now describes what the hook does (enrich the bare EOF with whatever the transport can see) instead of speculating about typical causes.

---

src/fleche/remote.py
pmrv: None — `logger.debug` already short-circuits on level internally, and the formatted args are small. Removed the `isEnabledFor` guards in 654c735.

---

src/fleche/remote.py
pmrv: Expanded `_SshConnection`'s docstring in 654c735 with the process model, the stream layout (stdin/stdout for RPC frames, stderr drained to logger + ring buffer for diagnostics), and the `&&` setup_commands chain that gets `exec`-ed into the server.

---

pmrv:
Re: introspecting remote `read_only` to skip round-trips — done in eab9df5.

- `_server_info` now exposes a `read_only` flag computed via a new `_is_read_only` helper: `isinstance(cache, ReadOnlyMixin)` (covers `ReadOnlyCache` and `FilteredCache`) and recurses into `CacheStack.stack[0]` since that's where saves go.
- `SshCache` lazily caches the full info dict on first access; `save`/`evict` consult the cached `read_only` flag and raise `Rejected` directly with no RPC.
- `SshCache.info(refresh=True)` (default) re-fetches and updates the cache; `info(refresh=False)` reuses it. `reconnect()` invalidates the cache so the next info reflects the new subprocess.
- New unit tests spy on `_conn.call` to assert zero RPCs after the first info fetch when saving against a read-only remote.

---

pmrv (commented):
will need a deep dive explanation in the developer section of docs

pmrv:
Done in 96b6657. New page `docs/dev/ssh_cache.rst` registered under the Development toctree, covering:

- Design goals + constraints (no daemon, single auth, backend-agnostic, cheap round-trips)
- Process / stream layout (RPC frames over stdin/stdout, stderr → daemon thread → logger + ring buffer)
- Wire protocol (length-prefixed cloudpickle, `(method, args, kwargs)` / `("ok"|"err", payload)`)
- Server dispatch — the explicit method table and why `_strip_cache` exists
- Lifecycle: lazy spawn, no auto-reconnect (the 2FA rationale), `setup_commands` `&&`-chain ending in `exec`
- The `info` RPC and what each key surfaces; active-cache vs served-cache wiring through `fleche.state._CACHE`
- Read-only short-circuit + lazy info cache
- Two-layer testing strategy (in-process pipes for unit, real subprocess for integration)
- "Adding a new RPC" recipe

Sphinx builds clean with no new warnings for the new page.

---

pmrv:
Critically evaluate the full change set and flag most important design decisions, potential foot guns, etc.

pmrv:
Here's a critical evaluation of the SshCache change set. Grouped by importance; not sugar-coating.

## Decisions where the alternative was a real choice

**D1. Cloudpickle as the wire format.** Pragmatic: we need to ship arbitrary Python objects (Call/DigestedCall/value payloads, user types, numpy arrays). Alternatives — JSON, msgpack, protobuf — would force a typed schema that fights against fleche's whole "your function takes whatever, we hash it" ethos. **Locks us into trusting the wire** (see F1 below) but the trust boundary is the same as `ssh user@host`, so it's the right v1 call. If we ever want untrusted-multi-tenant, this is the load-bearing decision to revisit.

**D2. One mutex, blocking request/response, no streaming or pipelining.** Cheap to implement, easy to reason about, trivially thread-safe. Throughput is round-trip-limited — a batch `contains()` over 10k digests is 10k serial RPCs. For fleche's actual usage pattern (sporadic cache ops, occasional batch queries) this is fine. Going async with request IDs would buy throughput at significant complexity cost; postpone unless real users feel the pain.

**D3. No auto-reconnect on dropped connection.** Hard-fail with `RemoteConnectionError` until `reconnect()` is called explicitly. The motivation is real: an auto-reconnect would silently re-trigger interactive 2FA. The friction is also real: a long-running notebook or batch job loses progress on a transient network blip. Net: the right default given the 2FA constraint, but the docs should mention the wrapper-script pattern for batch jobs.

**D4. `query()` materializes the whole iterator into one tuple on the server.** Simple within the request/response model. Foot-gun on large caches — see F3.

**D5. Explicit `_dispatch` table instead of reflective `getattr(cache, method)`.** Costs an `if` per method on add; gains: documents the API surface, prevents remote callers from poking arbitrary cache attributes, gives translation (e.g., `_strip_cache`) a clear place to live. Worth the boilerplate.

**D6. `LazyCall` responses ship digests only, not inline values.** Values round-trip lazily through `load_value` on first `.result` access. Right call — query results are usually inspected for metadata, not all results loaded — but it means a "query then materialise all" pattern is N+1 RPCs.

**D7. Lifetime-of-process subprocess, lazy spawn, atexit cleanup.** Sensible default for the dominant use case (interactive notebook, one persistent remote). See F7 for the GC corner case.

## Foot guns and rough edges

**F1. Trust model: cloudpickle in both directions = RCE in both directions.** The server `cloudpickle.loads`-es whatever the client sends; the client `cloudpickle.loads`-es whatever the server returns. Trust boundary == SSH trust. This is the same trust as `ssh user@host` so it's defensible, but the threat model deserves a callout in the docs: a hostile remote can run code on your client, full stop. **We do not document this.**

**F2. `info()` leaks `secret_key` via `cache_to_config`.** Confirmed in `config.py:367` — `cache_to_config` on a pickle storage emits the HMAC signing keys as hex. The new `info` RPC ships that across the wire, the client caches it, and DEBUG-level RPC tracing on `fleche.remote` logs the full response payload verbatim. Anyone with debug logging configured will write the remote's signing keys to disk. Same risk for sqlite `sqlite:///` URLs with embedded credentials, postgres URLs, etc. **Should redact known-sensitive fields before sending and/or before logging.**

**F3. `query()` over a large remote cache.** No streaming, no cursor, no early-exit. Server materialises every `DigestedCall` into a tuple, ships the whole tuple in one frame, client iterates the tuple. A user writing `for c in sc.query(): if c.x > 0: break` pays for all N results regardless. For caches with >>10k entries this becomes painful. Page through with a server-side iterator cursor when this bites.

**F4. `setup_commands` are unquoted shell code.** Each string is passed verbatim to the remote shell, joined with `&&`. If `fleche.toml` is shared or checked in, anyone who can edit it gets remote code execution at session start. The user provides the strings, so this isn't an injection vuln in the classical sense — but documenting "treat these as code, not data" is worth doing.

**F5. No protocol version negotiation.** If client fleche and server fleche disagree on the `Call`/`DigestedCall`/`LazyCall` schema (e.g. a new field gets added), cloudpickle deserialisation will mostly succeed and produce silently-wrong records. Same risk between cloudpickle major versions. A handshake that exchanges fleche+cloudpickle versions at connect time would fail fast with a meaningful error.

**F6. `connect_timeout` is a documented public dataclass field that does nothing.** "Reserved for future use; currently informational." Users will set it and be surprised it has no effect. Either implement it (wrap the connect in a watchdog), or remove it from the public API for v1.

**F7. `SshCache` instances aren't auto-closed on GC.** `atexit.register(self.close)` holds a reference to the connection forever. If the user creates and drops many SshCache instances in a long-running session, the subprocesses pile up (alive until process exit, not eligible for GC because atexit holds them). Probably not a real workload, but worth a `__del__` or a weakref-based atexit. Document at least.

**F8. `read_only` is a `@property` that silently spends one RPC on first access.** A user poking `sc.read_only` in a REPL gets a network call they didn't expect. The cost is bounded (cached afterwards), but documenting "this lazy-fetches an info() the first time" would prevent surprises.

**F9. Single mutex = a slow `query()` blocks every other RPC on the cache.** Threaded callers serialise behind a thread doing a large query. Probably fine for fleche's typical workloads (single-threaded scientific code), but worth knowing.

**F10. `_strip_cache` reaches into `LazyCall._arguments` / `._result`.** If LazyCall's internal layout ever changes, this breaks silently (no test asserts on the internals). Worth promoting to a `LazyCall.to_digested_call()` public method.

**F11. `_is_read_only` only recognises `ReadOnlyMixin`.** A custom subclass that overrides `save` to always raise `Rejected` won't be detected; we'll round-trip and catch. Acceptable fallback, but the detection logic is necessarily best-effort, not a guarantee.

## What I'm confident is right

- Stack composition for read-through (`[[shared]]` array-of-tables → CacheStack with local first, ssh second).
- Persistent SSH + ControlMaster/ControlPersist guidance for 2FA-protected hosts.
- Daemon stderr drain + 50-line ring buffer surfacing through `_diagnose` — turned bare-EOF failures into actionable error messages.
- `_Connection` as an `abc.ABC` with one abstract `_open`, swappable transports for testing (in-process pipes for unit, real subprocess for integration).
- Activating the served cache via `fleche.cache(name)` (ContextVar), not just `load_cache_config(name)`, so nested fleche operations on the remote see the same cache.
- Read-only short-circuit: at most one extra RPC per session, zero in the common case, big win for write-heavy workflows against an RO mirror.

## Most worth fixing before merge, in order

1. **F2** (secret_key leak via info) — real security issue, easy to fix with a redactor in `_server_info` and/or in the DEBUG log line.
2. **F1** (trust model docs) — add a "Security model" section to `docs/dev/ssh_cache.rst` so users understand what they're signing up for.
3. **F6** (`connect_timeout` no-op) — remove the field or implement it; don't keep a misleading public API.
4. **F5** (no version negotiation) — small handshake at session open prevents a class of confusing-failure bugs.

The rest (F3, F4, F7, F8, F9, F10, F11) are real but not blocking — document them in the dev page and address as users hit them.

---

pmrv:
Responding to the five commit comments on 96b6657 (commit comments aren't reachable via the MCP tooling, so I'm answering here). All addressed in 1e5f66b.

- **L47 (ASCII diagram says `save()` specifically)** — generalised to `SshCache.<op>(...)` and listed the actual ops dispatched (save / load / load_value / contains / expand / shrink / query / evict / info).
- **L69 (where is the per-host logger defined? per-host sub-loggers would be cool)** — the name was constructed in `_SshConnection._open` as `fleche.remote[<host>]`, but bracket syntax is **not** a Python logger hierarchy separator, so `fleche.remote[somehost]` was a sibling of `fleche.remote`, not a child — setting level on the parent didn't propagate. Now uses dotted naming `fleche.remote.host.<sanitised_host>` (dots and `@` become `_`) so it's a real child of `fleche.remote` and standard config inheritance works. Documented this and the rationale in the stderr bullet of the "Process and stream layout" section.
- **L151 (`reconnect()` not in the error message but should be)** — fixed in `_Connection.call`: the `RemoteConnectionError` message now reads ``Remote cache connection lost during 'contains': ...\nCall SshCache.reconnect() to spawn a fresh subprocess.`` Existing tests updated to assert on the new hint.
- **L217 ("empty `.table()`" anecdote too specific)** — generalised the `info()` section to "any unexpected remote state — wrong cache, wrong working directory, surprising read_only, surprising Python, surprising host" instead of a single failure-mode story.
- **L224 (`load_cache_config` framing is a historical artifact)** — rewrote the section as "Active cache on the remote": just states the invariant directly (the served cache *is* the active cache, the process is single-threaded, that's the only sensible behaviour). No more comparison to the older code path.

All 33 unit + integration tests pass; `ty` clean.

---

pmrv:
> ## Most worth fixing before merge, in order
> 
>     1. **F2** (secret_key leak via info) — real security issue, easy to fix with a redactor in `_server_info` and/or in the DEBUG log line.
> 
>     2. **F1** (trust model docs) — add a "Security model" section to `docs/dev/ssh_cache.rst` so users understand what they're signing up for.
> 
>     3. **F6** (`connect_timeout` no-op) — remove the field or implement it; don't keep a misleading public API.
> 
>     4. **F5** (no version negotiation) — small handshake at session open prevents a class of confusing-failure bugs.

pick these up

pmrv:
Picked up the four must-fix items in 69103d6.

**F2 — `info()` credential redaction**
`_server_info` now walks the `cache_to_config` output through a new `_redact_config` helper before putting it on the wire: `secret_key` becomes `"<redacted>"`, URL passwords are masked via `_redact_url_password` (e.g. `postgresql://alice:hunter2@db:5432/x` → `postgresql://alice:***@db:5432/x`), and the rest of the URL stays visible for diagnostics. Sensitive field names live in `_SENSITIVE_CONFIG_KEYS`; extend that frozenset when adding new credential-bearing storage types.

**F5 — version handshake**
`_server_info` now also exposes `fleche_version` and `cloudpickle_version`. Every BaseCache method on `SshCache` calls a new `_ensure_handshake()` at the top, which triggers the lazy info fetch and runs `_warn_on_version_skew` against the response. Skew logs a `WARNING` on `fleche.remote` rather than raising — forwards-compatible patch releases are common and a hard failure mid-session leaves users no recourse. Net cost: exactly one extra RPC at session open.

**F6 — `connect_timeout` removed**
Was documented as "reserved for future use; currently informational" — a misleading public API field that did nothing. Removed from both the `SshCache` dataclass and the `cache_to_config` round-trip.

**F1 — security-model docs**
New "Security model" section at the top of `docs/dev/ssh_cache.rst` covers:
- The cloudpickle = bidirectional-RCE trust model and the boundary == SSH trust.
- Don't expose the server entry point on any transport other than the spawned ssh stdin/stdout.
- The credential-redaction behaviour and how to extend `_SENSITIVE_CONFIG_KEYS`.
- The version handshake and why it warns rather than raises.

Tests: 5 new unit tests covering each piece (redactor on `secret_key`, URL-password masking on postgres/sqlite/no-password URLs, info exposes versions, version skew logs warning, handshake runs on the first op). 38 unit + integration tests pass; `ty` clean.

---

pmrv:
document f7, f8, remove f6, open new issue for f10.

rewrite the OP to reflect the final stage of the code, squash commits (more than one is ok, if you feel there's natural units of change here)

pmrv:
Oh and your ascii graphic has the ' stdin  (cloudpickle frame)' written such that it cuts through the list of ops on the right.  move it to the left of the arrow its annotating.  you ssh command block also doesn't have any arrows (for data) coming out, though presumably we do get data from it.

PR #566, opened 2026-05-25
feat(query): add QueryIterator.transfer for replaying matches into a target cache

pmrv:
## Summary

`QueryIterator.transfer(target, pop=False, overwrite=False)` replays the iterator's matching calls into a target cache. Mirrors `BaseCache.transfer` semantics but operates on the already-filtered subset, so callers can chain `.filter() / .skip() / .take() / .unique() / ...` before transferring.

```python
source.query(name="my_func").filter(lambda c: c.arguments["x"] > 10).transfer(target)
```

- `overwrite=False` (default) skips keys that already exist in the target.
- `pop=True` evicts transferred calls from their source cache (skipped on conflict, with the same rule as `BaseCache.transfer`).

## Test plan

- [x] `pytest tests/unit/fleche/test_query_iterator.py` — 63 passed (6 new)
- [x] `pytest tests/unit/caches/ tests/unit/fleche/` — 243 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01QhnBfrUJCM8DPLw6EjVVjm)_

### DIFF
--- a/src/fleche/query.py
+++ b/src/fleche/query.py
@@ -1,12 +1,15 @@
 import builtins
 import datetime
 import itertools
+import logging
 from dataclasses import dataclass
 from typing import Iterable, Iterator, Any, Literal, Callable
 
 import pandas as pd
 
 from . import call
+
+logger = logging.getLogger("fleche.query")
 
 
 def _resolve_key(key: "str | Callable[[call.LazyCall], Any]") -> "Callable[[call.LazyCall], Any]":
@@ -179,9 +182,14 @@ class QueryIterator(Iterable[call.LazyCall]):
         for c in self:
             key = c.to_lookup_key()
             conflict = not overwrite and target.contains(key)
-            if not conflict:
-                target.save(c.fetch())
-            if pop and not conflict:
+            if conflict:
+                logger.warning(
+                    "Not transferring %s: already exists in target and overwrite=False",
+                    key,
+                )
+                continue
+            target.save(c.fetch())
+            if pop:
                 c._cache.evict(key)
 
     def table(self, arguments: Iterable[str] | str | Literal[True] = (), results=False) -> pd.DataFrame:
--- a/src/fleche/query.py
+++ b/src/fleche/query.py
@@ -183,9 +183,13 @@ class QueryIterator(Iterable[call.LazyCall]):
             key = c.to_lookup_key()
             conflict = not overwrite and target.contains(key)
             if conflict:
+                try:
+                    short = target.shrink(key)
+                except KeyError:
+                    short = key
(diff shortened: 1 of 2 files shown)


src/fleche/query.py
```diff
@@ -167,6 +167,23 @@ def evict(self) -> None:
...
+                If False (default), conflicts are skipped.
+        """
+        for c in self:
+            key = c.to_lookup_key()
+            conflict = not overwrite and target.contains(key)
+            if not conflict:
+                target.save(c.fetch())
+            if pop and not conflict:
+                c._cache.evict(key)
+
```
pmrv, line 194: needs a warning on conflict

src/fleche/query.py
```diff
@@ -167,6 +170,28 @@ def evict(self) -> None:
...
+            overwrite: if True, overwrite entries already present in the target.
+                If False (default), conflicts are skipped.
+        """
+        for c in self:
+            key = c.to_lookup_key()
+            conflict = not overwrite and target.contains(key)
+            if conflict:
+                logger.warning(
+                    "Not transferring %s: already exists in target and overwrite=False",
+                    key,
```
pmrv: use shrunk version

tests/unit/fleche/test_query_iterator.py
```diff
@@ -621,3 +621,92 @@ def test_evict_only_removes_matched(test_cache):
...
+
+
+def test_transfer_skips_existing_by_default(test_cache, other_cache, caplog):
+    c = Call(name="f", arguments={"x": 1}, result=10)
+    test_cache.save(c)
+    # Pre-populate target with the same key but a different stored result
+    other_cache.calls.save(Call(name="f", arguments={"x": 1}, result=999).stash(other_cache.values))
+
+    tpl = QueryCall(name="f", arguments=None, metadata=None, module=None, version=None, result=None)
+    with caplog.at_level("WARNING", logger="fleche.query"):
```
pmrv, line 670: more specific

src/fleche/query.py
```diff
@@ -167,6 +170,32 @@ def evict(self) -> None:
...
+                If False (default), conflicts are skipped.
+        """
+        for c in self:
+            key = c.to_lookup_key()
+            conflict = not overwrite and target.contains(key)
+            if conflict:
+                try:
+                    short = target.shrink(key)
+                except KeyError:
+                    short = key
```
pmrv: This never be case, iirc.  shrink raises KeyError iff key does not exists in cache, but by virtue having queried it the key should exist.  There's a race condition here, log it into one of open issues we have to tracking cache parallel features

PR #574, opened 2026-05-27
feat(query): shrink lookup keys in QueryIterator.table by default

pmrv:
## Summary

- Add a `shrink_keys` argument (default `True`) to `QueryIterator.table` and `Cache.table`.
- When set, each entry in the DataFrame index is replaced with the shortest unambiguous prefix via the owning cache's `shrink` (`AmbiguousDigestError` falls back silently to the full digest, which shouldn't happen in practice since the key was just retrieved from the cache).
- Pass `shrink_keys=False` to keep raw 64-char digests — cheaper on large caches, and what you want when the index needs to be a literal storage key (e.g. the existing issue #319 regression test).

## Test plan

- [x] `pytest tests/unit/fleche/test_query_iterator.py` (new tests for `shrink_keys=True/False`)
- [x] `pytest tests/regression/` (issue #319 updated to opt out)
- [x] Full unit + integration suite green

A follow-up comment will post a benchmark of `.table()` for a cache with 10 entries called 100 times, with and without shrinking.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VM8QGrdsiMAmLdY5Zrrw8s)_

### DIFF
--- a/tests/unit/fleche/test_query_iterator.py
+++ b/tests/unit/fleche/test_query_iterator.py
@@ -4,6 +4,7 @@ import pandas as pd
 from fleche import fleche, cache
 from fleche.call import Call, LazyCall, QueryCall
 from fleche.caches import Cache
+from fleche.digest import DIGEST_LENGTH
 from fleche.query import QueryIterator
 from fleche.storage import ValueMemory, CallMemory
 
@@ -156,8 +157,7 @@ def test_query_iterator_table_index_shrunk_by_default(test_cache):
         str(test_cache.shrink(call2.to_lookup_key())),
     }
     assert set(df.index) == expected_keys
-    # Both should be strictly shorter than the full 64-char digest.
-    assert all(len(k) < 64 for k in df.index)
+    assert all(len(k) < DIGEST_LENGTH for k in df.index)
 
 
 def test_query_iterator_table_shrink_keys_false_keeps_full_digest(test_cache):
@@ -165,7 +165,7 @@ def test_query_iterator_table_shrink_keys_false_keeps_full_digest(test_cache):
     test_cache.save(Call(name="f", arguments={"x": 1}, result=10))
     tpl = QueryCall(name="f", arguments=None, metadata=None, module=None, version=None, result=None)
     df = test_cache.query(tpl).table(shrink_keys=False)
-    assert all(len(k) == 64 for k in df.index)
+    assert all(len(k) == DIGEST_LENGTH for k in df.index)
 
 
 def test_query_iterator_table_no_result_by_default(test_cache):
--- a/tests/unit/fleche/test_query_iterator.py
+++ b/tests/unit/fleche/test_query_iterator.py
@@ -135,10 +135,37 @@ def test_query_iterator_table_index_is_lookup_key(test_cache):
     test_cache.save(call2)
 
     tpl = QueryCall(name="f", arguments=None, metadata=None, module=None, version=None, result=None)
-    df = test_cache.query(tpl).table()
+    df = test_cache.query(tpl).table(shrink_keys=False)
 
     expected_keys = {str(call1.to_lookup_key()), str(call2.to_lookup_key())}
     assert set(df.index) == expected_keys
+
+
+def test_query_iterator_table_index_shrunk_by_default(test_cache):
+    """By default, the DataFrame index uses shortest unambiguous prefixes."""
+    call1 = Call(name="f", arguments={"x": 1}, result=10)
+    call2 = Call(name="f", arguments={"x": 2}, result=20)
+    test_cache.save(call1)
+    test_cache.save(call2)
+
+    tpl = QueryCall(name="f", arguments=None, metadata=None, module=None, version=None, result=None)
+    df = test_cache.query(tpl).table()
+
+    expected_keys = {
+        str(test_cache.shrink(call1.to_lookup_key())),
+        str(test_cache.shrink(call2.to_lookup_key())),
+    }
+    assert set(df.index) == expected_keys
+    # Both should be strictly shorter than the full 64-char digest.
+    assert all(len(k) < 64 for k in df.index)
+
+
+def test_query_iterator_table_shrink_keys_false_keeps_full_digest(test_cache):
+    """shrink_keys=False leaves the full-length digest as the index."""
+    test_cache.save(Call(name="f", arguments={"x": 1}, result=10))
+    tpl = QueryCall(name="f", arguments=None, metadata=None, module=None, version=None, result=None)
+    df = test_cache.query(tpl).table(shrink_keys=False)
+    assert all(len(k) == 64 for k in df.index)
 
 
 def test_query_iterator_table_no_result_by_default(test_cache):
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -3,7 +3,7 @@ import logging
 import random
 import threading
 from dataclasses import dataclass, replace, field
-from typing import Iterable, Any, Callable, Literal
+from typing import Iterable, Any, Callable, Literal, overload
 
 import pandas as pd
 
@@ -103,6 +103,10 @@ class BaseCache(ABC):
         """
         ...
 
+    @overload
+    def shrink(self, key: Digest | str, /) -> Digest: ...
+    @overload
+    def shrink(self, key: Digest | str, /, *keys: Digest | str) -> "tuple[Digest, ...]": ...
     @abstractmethod
     def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
         """
@@ -113,6 +117,11 @@ class BaseCache(ABC):
         the batched form lets sub-storages list their keys once instead of
         per-key, which matters on backends where listing is expensive (e.g.
         SQL, filesystem).
+
+        Each input key must belong to *one* of the sub-storages (call or
+        value).  Mixing call keys and value keys in a single call is
+        undefined behaviour — the result depends on internal partitioning
+        order and may change without notice.
 
         .. warning::
 
@@ -162,7 +171,7 @@ class BaseCache(ABC):
                 yield from self._query(template)
             except _digest.Indigestible as e:
                 logger.warning("No hash for query argument: %s", e.args[0])
-        return query.QueryIterator(_safe_iter)
+        return query.QueryIterator(_safe_iter, cache=self)
 
     def table(
         self,
@@ -287,6 +296,10 @@ class Cache(BaseCache):
                 pass
         return _combine_expand(key, results)
 
+    @overload
+    def shrink(self, key: Digest | str, /) -> Digest: ...
+    @overload
+    def shrink(self, key: Digest | str, /, *keys: Digest | str) -> "tuple[Digest, ...]": ...
     def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
         if not keys:
             raise TypeError("shrink() requires at least one key")
@@ -441,6 +454,10 @@ class CacheWrapper(BaseCache):
     def expand(self, key: Digest | str) -> Digest:
         return self.cache.expand(key)
 
+    @overload
+    def shrink(self, key: Digest | str, /) -> Digest: ...
+    @overload
+    def shrink(self, key: Digest | str, /, *keys: Digest | str) -> "tuple[Digest, ...]": ...
     def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
         return self.cache.shrink(*keys)
 
@@ -560,6 +577,10 @@ class CacheStack(BaseCache):
     def expand(self, key: Digest | str) -> Digest:
         return _combine_expand(key, self._collect(lambda c: c.expand(key)))
 
+    @overload
+    def shrink(self, key: Digest | str, /) -> Digest: ...
+    @overload
+    def shrink(self, key: Digest | str, /, *keys: Digest | str) -> "tuple[Digest, ...]": ...
     def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
         if not keys:
             raise TypeError("shrink() requires at least one key")
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -104,9 +104,15 @@ class BaseCache(ABC):
         ...
 
     @abstractmethod
-    def shrink(self, key: Digest | str) -> Digest:
+    def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
         """
-        Find the shortest substring that is still an unambigious reference to the same call.
+        Find the shortest substring(s) that unambiguously reference each call.
+
+        With a single key, returns one :class:`Digest`.  With multiple keys,
+        returns a tuple of :class:`Digest` in the same order as the inputs;
+        the batched form lets sub-storages list their keys once instead of
+        per-key, which matters on backends where listing is expensive (e.g.
+        SQL, filesystem).
 
         .. warning::
 
@@ -115,13 +121,13 @@ class BaseCache(ABC):
             Do not rely on this function in your programs, it is provided as a convenience for users only!
 
         Args:
-            key (str or :class:`Digest`): the key to shorten
+            *keys (str or :class:`Digest`): one or more keys to shorten
 
         Returns:
-            :class:`Digest`: shortest key possible
+            :class:`Digest` (single key) or tuple of :class:`Digest` (multiple)
 
         Raises:
-            :class:`AmbiguousDigestError`: if no shorter key is possible
+            :class:`AmbiguousDigestError`: if no shorter key is possible for any input
         """
         ...
 
@@ -281,12 +287,29 @@ class Cache(BaseCache):
                 pass
         return _combine_expand(key, results)
 
-    def shrink(self, key: Digest | str) -> Digest:
-        if self.calls.contains(key):
-            return self.calls.shrink(key)
-        if self.values.contains(key):
-            return self.values.shrink(key)
-        raise KeyError(key)
+    def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
+        if not keys:
+            raise TypeError("shrink() requires at least one key")
+        call_keys: list = []
+        value_keys: list = []
+        for k in keys:
+            if self.calls.contains(k):
+                call_keys.append(k)
+            elif self.values.contains(k):
+                value_keys.append(k)
+            else:
+                raise KeyError(k)
+        results: dict = {}
+        for sub, ks in ((self.calls, call_keys), (self.values, value_keys)):
+            if not ks:
+                continue
+            r = sub.shrink(*ks)
+            if len(ks) == 1:
+                r = (r,)
+            for k, s in zip(ks, r):
+                results[k] = s
+        out = tuple(results[k] for k in keys)
+        return out[0] if len(keys) == 1 else out
 
     def _query(self, call: call.QueryCall) -> Iterable[LazyCall]:
         """Query for cached calls that match a template and return decoded results.
@@ -418,8 +441,8 @@ class CacheWrapper(BaseCache):
     def expand(self, key: Digest | str) -> Digest:
         return self.cache.expand(key)
 
-    def shrink(self, key: Digest | str) -> Digest:
-        return self.cache.shrink(key)
+    def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
+        return self.cache.shrink(*keys)
 
     def _query(self, call: call.QueryCall) -> Iterable[LazyCall]:
         return self.cache.query(call)
@@ -537,8 +560,26 @@ class CacheStack(BaseCache):
     def expand(self, key: Digest | str) -> Digest:
         return _combine_expand(key, self._collect(lambda c: c.expand(key)))
 
-    def shrink(self, key: Digest | str) -> Digest:
-        return _combine_shrink(key, self._collect(lambda c: c.shrink(key)))
+    def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
+        if not keys:
+            raise TypeError("shrink() requires at least one key")
+        per_key: dict = {k: [] for k in keys}
+        for cache in self.stack:
+            present = [k for k in keys if cache.contains(k)]
+            if not present:
+                continue
+            r = cache.shrink(*present)
+            if len(present) == 1:
+                r = (r,)
+            for k, s in zip(present, r):
+                per_key[k].append(s)
+        out_list = []
+        for k in keys:
+            if not per_key[k]:
+                raise KeyError(k)
+            out_list.append(_combine_shrink(k, per_key[k]))
+        out = tuple(out_list)
+        return out[0] if len(keys) == 1 else out
 
     def _query(self, call: call.QueryCall) -> Iterable[LazyCall]:
         """Aggregate query results across the stack, avoiding duplicates.
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -158,7 +158,12 @@ class BaseCache(ABC):
                 logger.warning("No hash for query argument: %s", e.args[0])
         return query.QueryIterator(_safe_iter)
 
-    def table(self, arguments: Iterable[str] | str | Literal[True] = (), results=False) -> pd.DataFrame:
+    def table(
+        self,
+        arguments: Iterable[str] | str | Literal[True] = (),
+        results=False,
+        shrink_keys: bool = True,
+    ) -> pd.DataFrame:
         """Return a pandas DataFrame summarizing cached calls via query().
 
         This implementation uses a fully-wildcard Call template to retrieve
@@ -182,6 +187,9 @@ class BaseCache(ABC):
                 Pass ``True`` to add all arguments, or a single string as a shortcut for a
                 one-element tuple.
             results (bool): if True, add results of queried calls to table
+            shrink_keys (bool): if True (default), shrink each index entry to
+                its shortest unambiguous prefix.  Set to ``False`` to keep
+                full-length digests.
 
         Returns:
             :class:`pandas.DataFrame`: table of all calls on cache
@@ -194,7 +202,7 @@ class BaseCache(ABC):
             version=None,
             result=None,
         )
-        return self.query(tpl).table(arguments=arguments, results=results)
+        return self.query(tpl).table(arguments=arguments, results=results, shrink_keys=shrink_keys)
 
     def filter(self, predicate: Callable[[Call | LazyCall], bool] | QueryCall) -> 'FilteredCache':
         """Create a read-only view of this cache that only exposes calls matching the predicate.
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -50,9 +50,7 @@ class Digest(str):
         if cache is None:
             from .state import cache as get_cache
             cache = get_cache()
-        result = cache.shrink(self)
-        assert isinstance(result, Digest)  # single-arg shrink returns a single Digest
-        return result
+        return cache.shrink(self)
 
 
 DIGEST_LENGTH = 64
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -50,7 +50,9 @@ class Digest(str):
         if cache is None:
             from .state import cache as get_cache
             cache = get_cache()
-        return cache.shrink(self)
+        result = cache.shrink(self)
+        assert isinstance(result, Digest)  # single-arg shrink returns a single Digest
+        return result
 
 
 DIGEST_LENGTH = 64
--- a/src/fleche/query.py
+++ b/src/fleche/query.py
@@ -2,13 +2,15 @@ import builtins
 import datetime
 import itertools
 import logging
-from dataclasses import dataclass
-from typing import Iterable, Iterator, Any, Literal, Callable
+from dataclasses import dataclass, field
+from typing import Iterable, Iterator, Any, Literal, Callable, TYPE_CHECKING
 
 import pandas as pd
 
 from . import call
-from .storage.base import AmbiguousDigestError
+
+if TYPE_CHECKING:
+    from .caches import BaseCache
 
 logger = logging.getLogger("fleche.query")
 
@@ -36,9 +38,14 @@ class QueryIterator(Iterable[call.LazyCall]):
 
     Args:
         calls: factory returning an iterable of :class:`~fleche.call.LazyCall`
+        cache: the cache that produced these calls; set by
+            :meth:`~fleche.caches.BaseCache.query` and propagated through
+            chainable methods.  Used by :meth:`table` to call ``shrink``
+            without reaching into ``LazyCall._cache``.
     """
 
     calls: Callable[[], Iterable[call.LazyCall]]
+    cache: "BaseCache | None" = field(default=None, repr=False)
 
     def __iter__(self) -> Iterator[call.LazyCall]:
         yield from self.calls()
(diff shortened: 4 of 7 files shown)


pmrv:
### Benchmark: `shrink_keys=True` overhead

Setup: in-memory `Cache(ValueMemory, CallMemory)` with **10 saved entries**; time `cache.table(shrink_keys=…)` over batches of N calls; 5 repeats, report best/mean.

Python 3.11.15 · Linux x86_64 (CI runner-class, single thread).

| N calls | `shrink_keys=True` (best / mean) | `shrink_keys=False` (best / mean) | overhead (best) | per `.table()` |
|---:|---|---|---:|---:|
| 100  | 174.6 ms / 184.1 ms | 131.1 ms / 133.2 ms | **+33.2%** | ~435 µs |
| 500  | 862.3 ms / 877.0 ms | 649.5 ms / 662.4 ms | **+32.8%** | ~426 µs |
| 1000 | 1743.4 ms / 1762.5 ms | 1347.8 ms / 1362.8 ms | **+29.3%** | ~396 µs |

So for a 10-entry cache, turning shrinking on adds **~0.4 ms per `.table()` call**, or **~40 µs per row** (each row triggers one `shrink` ⇒ one `expand` ⇒ one full `list()` scan over the storage). The relative overhead is ~30% of `.table()`'s own cost, but in absolute terms it's negligible at this size and constant per call — over 100 calls the total cost difference is only ~43 ms.

Scaling note: `shrink` is O(N_entries) per row, so the cost grows as O(N_entries · N_rows_in_table). For 10 entries this is cheap; for very large caches users can pass `shrink_keys=False` to bypass it.

Bench script:

```python
import time, statistics
from fleche.call import Call
from fleche.storage import ValueMemory, CallMemory

c = Cache(values=ValueMemory({}), calls=CallMemory({}))
for i in range(10):
    c.save(Call(name="f", arguments={"x": i}, result=i * 2))

for n_calls in (100, 500, 1000):
    for sk in (True, False):
        samples = []
        for _ in range(5):
            t0 = time.perf_counter()
            for _ in range(n_calls):
                c.table(shrink_keys=sk)
            samples.append(time.perf_counter() - t0)
        print(n_calls, sk, min(samples), statistics.mean(samples))
```

---

pmrv:
do the same table but N controls total size of cache, and your query should subset ~10 entries to go into the table

pmrv:
### Benchmark v2: cache size N, query subsets ~10 calls

Setup: in-memory `Cache(ValueMemory, CallMemory)` populated with **10 calls named `"target"` + (N − 10) filler calls**; query template selects only the 10 targets; time `cache.query(tpl).table(shrink_keys=…)` over **100 calls**, 5 repeats, best + mean.

Python 3.11.15 · Linux x86_64.

| N (cache size) | `shrink_keys=True` (best / mean) | `shrink_keys=False` (best / mean) | overhead | per `.table()` | per row |
|---:|---|---|---:|---:|---:|
| 10     | 208.4 ms / 221.6 ms     | 159.0 ms / 163.3 ms     | **+31.1%** | ~494 µs   | ~49 µs |
| 100    | 507.6 ms / 518.6 ms     | 434.9 ms / 450.5 ms     | **+16.7%** | ~727 µs   | ~73 µs |
| 1 000  | 3 508 ms / 3 625 ms     | 3 185 ms / 3 216 ms     | **+10.1%** | ~3.2 ms   | ~323 µs |
| 10 000 | 34 848 ms / 35 008 ms   | 30 985 ms / 31 089 ms   | **+12.5%** | ~38.6 ms  | ~3.86 ms |

Shrink cost per row scales roughly linearly with cache size (each row's `shrink` does an `expand` ⇒ full storage `list()`), so the per-`.table()` overhead grows ~O(N_cache × N_rows). For the prompt's "10 entries in a cache with 100s of calls" the absolute overhead is tiny (~50 ms over 100 calls); past ~1 k cache entries with 10-row tables the cost climbs into the few-ms-per-call range and `shrink_keys=False` becomes worth setting.

Note: the `.table()` baseline itself scales with N too, because in-memory `_query` iterates every stored call to filter by name — that's why the *relative* overhead drops even as absolute shrink cost rises.

Bench script:

```python
import time, statistics
from fleche.call import Call, QueryCall
from fleche.storage import ValueMemory, CallMemory

def build(n_total, n_target=10):
    c = Cache(values=ValueMemory({}), calls=CallMemory({}))
    for i in range(n_target):
        c.save(Call(name="target", arguments={"x": i}, result=i*2))
    for i in range(n_total - n_target):
        c.save(Call(name="filler", arguments={"x": i}, result=i*2))
    return c

tpl = QueryCall(name="target", arguments=None, metadata=None,
                module=None, version=None, result=None)
for n_total in (10, 100, 1000, 10_000):
    c = build(n_total)
    for sk in (True, False):
        s = []
        for _ in range(5):
            t = time.perf_counter()
            for _ in range(100):
                c.query(tpl).table(shrink_keys=sk)
            s.append(time.perf_counter() - t)
        print(n_total, sk, min(s), statistics.mean(s))
```

---

pmrv:
do one more benchmark with a sql call storage and a cloudpickle value storage.

let's also casually think whether shrinking multiple keys together at once would help

pmrv:
### Benchmark v3: SQL calls + cloudpickle values

Same setup as v2 (10 `"target"` + (N−10) `"filler"`, query selects the 10, 100 `.table()` calls, 3 repeats), but the cache now uses `Sql` for `calls` and `ValuePickleFile.with_cloudpickle(...)` for `values`.

Python 3.11.15 · Linux x86_64 (tmpfs-backed).

| N (cache) | `shrink_keys=True` (best / mean) | `shrink_keys=False` (best / mean) | overhead | per `.table()` | per row |
|---:|---|---|---:|---:|---:|
| 10   | 1 632.5 ms / 1 674.3 ms | 1 065.0 ms / 1 074.7 ms | **+53.3%**  | ~5.7 ms | ~568 µs |
| 100  | 2 180.6 ms / 2 271.8 ms | 1 012.4 ms / 1 029.4 ms | **+115.4%** | ~11.7 ms | ~1.17 ms |
| 1000 | 8 996.3 ms / 9 111.9 ms | 978.8 ms / 1 019.3 ms   | **+819.1%** | ~80.2 ms | ~8.02 ms |

This is much harsher than the pure-memory case:

- `shrink_keys=False` is *flat* in N — the underlying `Sql._query` pushes the `name="target"` filter down to SQL, so the query cost barely grows with N.
- `shrink_keys=True` grows roughly **linearly with N** (and the overhead crowds out the rest of `.table()` at 1k entries).

Reason: `Cache.shrink` calls `self.calls.shrink(key)` and `self.values.shrink(key)`. Each of those loops 4…64 over `expand(prefix)`, and each `expand` does a *full* `self.list()` — on SQL that's a `SELECT key FROM calls`, on the cloudpickle backend an `os.listdir` over every digest dir. So each row in the table triggers `~2 × prefix_len × (one SQL query + one disk listing)`. With 10 rows × ~5 prefix lengths × 2 backends, that's roughly 100 backend roundtrips per `.table()` call.

Strong argument for **`shrink_keys=False` as the SQL default** (or at least for documenting the cliff). I'm tempted to flip the default to `False` actually — opinions?

---

### Casual notes on batched shrinking

The current `shrink(key)` does `O(prefix_len)` calls to `expand`, each doing a full `list()` over the storage. For a table with R rows that's `R × O(prefix_len) × O(N_storage_list)`. There's a lot of redundant work.

**Sketch:** add `KeyManagement.shrink_many(keys: Iterable[str]) -> dict[str, str]` with a one-shot algorithm:

```python
def shrink_many(self, keys):
    sorted_all = sorted(self.list())          # one backend hit, O(N log N)
    out = {}
    for k in keys:
        i = bisect.bisect_left(sorted_all, k) # O(log N)
        # longest common prefix with the two neighbours in the sorted list
        lcp_left  = common_prefix_len(k, sorted_all[i-1]) if i > 0 else 0
        lcp_right = common_prefix_len(k, sorted_all[i+1]) if i+1 < len(sorted_all) else 0
        n = max(4, max(lcp_left, lcp_right) + 1)
        out[k] = k[:n]
    return out
```

Correctness intuition: a prefix is unambiguous iff it doesn't appear at the start of any *other* full key.  In a sorted key list, the only candidates for ambiguity with `k` are its immediate sorted neighbours — they share the longest prefix with `k` of any key in the list. So the answer is `max(LCP_left, LCP_right) + 1` (floored at 4 for the 4-char minimum).

Then `Cache.shrink_many` would call `shrink_many` on each sub-storage and combine per-key the way `_combine_shrink` does today, and `QueryIterator.table` would call `cache.shrink_many(all_keys)` once instead of `cache.shrink(k)` per row.

**Expected effect on the benchmarks above:**

- Pure-memory: `list()` is already cheap. The win would be ~`R−1` saved `list()` scans per `.table()` call. At R=10 rows and N=10 000, that's ~9 × 50 µs ≈ 0.5 ms shaved per call, modest. At R=100 it'd matter more.
- SQL + cloudpickle: this is where it really pays off. One SQL `SELECT key` per `.table()` call instead of `~50` (`10 rows × ~5 prefix lengths`). The per-row 8 ms at N=1000 should drop to ~well under 1 ms because the per-row cost stops involving any backend I/O at all.

**Caveats:**

- `Sql` doesn't implement `list()` directly today — it inherits from `CallStorage`. Need to check; if missing, batched shrink would have to push a single `SELECT key FROM calls` itself, or we add `list()` to `Sql`. Worth checking before promising the win.
- `CacheStack.shrink` (and `_combine_shrink`) need a batched counterpart, but it's a straightforward aggregation.
- Backwards compat: keep `shrink(key)` as `next(iter(self.shrink_many([key]).values()))`.
- Mild risk: caching the sorted list across the `.table()` call lifetime means a concurrent save during the table iteration would not be reflected — but that's already the case row-by-row today (and the existing docstring warns shrunk keys may become ambiguous in the future anyway).

I'd say it's worth a follow-up PR once this lands. Happy to take it on.

---

pmrv:
Why does BaseCache call *both* value and call shrink? That seems wasteful generic case even disregarding that in this specific case we know only the call storage matters.

pmrv:
Good catch — I think the answer is "defensiveness that no caller actually needs":

The current `Cache.shrink` (and `expand`) treats the *union* of `calls.list() ∪ values.list()` as the keyspace a short prefix must be unambiguous in.  That sounds principled, but:

1. **Call keys and value keys are disjoint by construction.**  Call keys digest `(name, args, module, version, code_digest)`; value keys digest the values themselves.  They can only collide via SHA-256 collision, and at that point we have bigger problems.
2. **Every legitimate use of a short key resolves into one side or the other.**  `cache.load(short)` → goes to `calls`.  `cache.load_value(short)` → goes to `values`.  No code path takes a short prefix and interprets it as "could be either" — the API the user picked already disambiguates which keyspace they're in.

So the "I want this prefix to be unambiguous in *both* storages" property buys correctness only for an interactive eyeballing scenario where a human glances at a printed digest and wants to be safe regardless of which `load*` method they happen to type next.  Real callers (`QueryIterator.table`, `QueryIterator.transfer`'s log line, any future `.gc` reporting) all know they're dealing with call keys.

So: in this PR I'll switch `QueryIterator.table` to call `c._cache.calls.shrink(key)` directly (or expose a small `_shrink_call_key` helper on `Cache` to keep `CacheStack` working).  That alone halves the per-row backend work and, on the SQL+cloudpickle bench, should cut the v3 overhead roughly in half before we even consider batched shrinking.

Separately, I think it's worth a follow-up to drop the `values` branch from `Cache.shrink`/`Cache.expand` entirely — but that's a behaviour change with a test (`test_cache_shrink_returns_longest_of_two_storages`) that explicitly pins down the current "longest of both" semantics, so it deserves its own PR + an AGENTS note about the rationale.

I'll push the targeted fix here and re-run the SQL benchmark to quantify.

---

pmrv:
### Re-bench after the call-storage-only fix (4d06edd)

Same SQL + cloudpickle setup as v3, after switching `QueryIterator.table` to `shrink_call_key`:

| N (cache) | `shrink_keys=True` (best / mean) | `shrink_keys=False` (best / mean) | overhead | per `.table()` | per row |
|---:|---|---|---:|---:|---:|
| 10   | 1 369.2 ms / 1 375.2 ms | 947.4 ms / 950.1 ms | **+44.5%** | ~4.2 ms | ~422 µs |
| 100  | 1 352.0 ms / 1 372.9 ms | 949.5 ms / 955.5 ms | **+42.4%** | ~4.0 ms | ~403 µs |
| 1000 | 1 423.3 ms / 1 467.0 ms | 960.4 ms / 969.4 ms | **+48.2%** | ~4.6 ms | ~463 µs |

Before vs. after, per-row shrink cost on SQL+cloudpickle:

| N | before | after | speedup |
|---:|---:|---:|---:|
| 10   | 568 µs | 422 µs | 1.3× |
| 100  | 1 168 µs | 403 µs | 2.9× |
| 1000 | 8 018 µs | 463 µs | **17.3×** |

The N=1000 case is the headline: the per-row cost stops scaling with cache size, because we no longer walk the cloudpickle directory tree on every prefix length.  The remaining ~400 µs/row is the `Sql.shrink` ⇒ `Sql.list()` ⇒ `SELECT key FROM calls` chain, which is the next thing batched shrinking would collapse to one query per `.table()`.

Memory-only cache numbers will be barely different (value-storage `list()` was cheap there); I can re-run if you want a complete picture.

---

pmrv:
Remove c04c6b471488bf87a3c065002095a1df79472875 since its useless now, then repeat benchmarks

pmrv:
Dropped c04c6b4 (`shrink_call_key`) — pushed e77f6e4 as the new head.  Re-ran the SQL+cloudpickle bench (same setup as v3/v4: 10 `"target"` + (N−10) `"filler"`, 100 `.table()` calls, 3 repeats).

| N (cache) | `shrink_keys=True` (best / mean) | `shrink_keys=False` (best / mean) | overhead | per `.table()` | per row |
|---:|---|---|---:|---:|---:|
| 10   | 1 605.6 ms / 1 623.5 ms | 891.6 ms / 904.6 ms | **+80.1%** | ~7.1 ms | ~714 µs |
| 100  | 1 508.8 ms / 1 561.6 ms | 857.2 ms / 863.8 ms | **+76.0%** | ~6.5 ms | ~652 µs |
| 1000 | 1 612.8 ms / 1 630.3 ms | 874.8 ms / 907.2 ms | **+84.4%** | ~7.4 ms | ~738 µs |

Per-row shrink cost across the three configurations:

| N | pre-#577 `shrink(both)` | post-#577 + `shrink_call_key` | post-#577 plain `shrink()` (HEAD) |
|---:|---:|---:|---:|
| 10   |   568 µs | 422 µs | 714 µs |
| 100  | 1 168 µs | 403 µs | 652 µs |
| 1000 | 8 018 µs | 463 µs | 738 µs |

So #577 alone gives the big win (no more scaling with N), and `shrink_call_key` was a further ~1.6× shave on top by skipping the extra `calls.contains(key)` lookup that `Cache.shrink` now does for dispatch.  Not worth the API surface — happy to live with the ~700 µs/row.

---

pmrv:
let's trial now the multi key shrinking.  I want it to be `shrink(*keys)` to stay backwards compat without new api.  benchmark afterwards

pmrv:
### Batched `shrink(*keys)` — bench

`Cache.shrink` now accepts `*keys` (single-arg form unchanged; multi-arg returns a tuple).  `QueryIterator.table` calls `cache.shrink(*all_keys)` once per `.table()` instead of per row.  Backing algorithm: sort the storage's key list once, then resolve each key by binary-searching its sorted position and taking `max(LCP_left, LCP_right) + 1` as the shrink length.

Same SQL+cloudpickle setup (10 `"target"` + (N−10) `"filler"`, 100 `.table()` calls, 3 repeats):

| N (cache) | `shrink_keys=True` (best / mean) | `shrink_keys=False` (best / mean) | overhead | per `.table()` | per row |
|---:|---|---|---:|---:|---:|
| 10   | 1 171.2 ms / 1 193.0 ms | 937.9 ms / 939.3 ms | **+24.9%** | ~2.3 ms | ~233 µs |
| 100  | 1 240.3 ms / 1 260.6 ms | 913.1 ms / 915.5 ms | **+35.8%** | ~3.3 ms | ~327 µs |
| 1000 | 1 464.2 ms / 1 482.1 ms | 948.7 ms / 984.9 ms | **+54.3%** | ~5.2 ms | ~516 µs |

Per-row shrink cost across the whole story:

| N | pre-#577 (combine both) | post-#577 (HEAD) | with `shrink_call_key` (dropped) | **batched `shrink(*keys)`** |
|---:|---:|---:|---:|---:|
| 10   |   568 µs | 714 µs | 422 µs | **233 µs** |
| 100  | 1 168 µs | 652 µs | 403 µs | **327 µs** |
| 1000 | 8 018 µs | 738 µs | 463 µs | **516 µs** |

The variadic form beats even the dedicated-method `shrink_call_key` route, and produces a single `SELECT key FROM calls` per `.table()` call instead of one per row.  The remaining ~500 µs at N=1000 is dominated by the O(N log N) sort of the storage key list inside the batched algorithm — that's amortised across all rows in the table, so the per-row figure is mostly an artefact of the 10-row table size.

Compared to the original pre-#577 baseline at N=1000, total per-row shrink cost dropped from **~8 ms → ~0.5 ms** (15.5×), with no special API surface — just a variadic positional signature.

API notes:
- Backwards compat: `cache.shrink(k)` returns a `Digest` (unchanged).  `cache.shrink(k1, k2, ...)` returns `tuple[Digest, ...]`.
- `shrink()` with no args raises `TypeError`.
- The semantic for "key cannot be shrunk shorter than itself" is now `n >= len(key)` (rather than the old loop-exits-empty behaviour) — same outcome, raises `AmbiguousDigestError` (covered by the existing `test_shrink_ambiguous_raises`).

---

tests/unit/fleche/test_query_iterator.py
```diff
@@ -135,12 +135,39 @@ def test_query_iterator_table_index_is_lookup_key(test_cache):
...
+    tpl = QueryCall(name="f", arguments=None, metadata=None, module=None, version=None, result=None)
+    df = test_cache.query(tpl).table()
+
+    expected_keys = {
+        str(test_cache.shrink(call1.to_lookup_key())),
+        str(test_cache.shrink(call2.to_lookup_key())),
+    }
+    assert set(df.index) == expected_keys
+    # Both should be strictly shorter than the full 64-char digest.
+    assert all(len(k) < 64 for k in df.index)
```
pmrv: respect the global var in digest module rather than hardcoding

pmrv (commented):
Critically evaluate the full change set and flag most important design decisions, potential foot guns, etc.

pmrv:
## Self-review of the full change set

### Design decisions worth flagging

**Default `shrink_keys=True`** is a silent behaviour change.  Anyone calling `cache.table().index` will now see ~6–8 character prefixes where they used to see 64-character digests.  `cache.load(short)` still works (the storage's `_normalize_key` re-expands), but any downstream code that *compares* index entries to keys retrieved from another source (e.g. `set(table.index) == set(c.calls.list())`) silently breaks.  Issue #319's regression test was the canary — it now passes `shrink_keys=False`.  Worth a CHANGELOG entry; AGENTS.md `QueryIterator` row may want an update too.

**Variadic `shrink(*keys)` with arity-dispatched return type** (`Digest` for 1, `tuple[Digest, ...]` for N) keeps the public API backwards compatible but is a known Python foot-gun: `cache.shrink(*some_list)` where `some_list` happens to be length 1 returns a `Digest`, not a `tuple[Digest]`.  Type checkers can't help — see the `Digest.shrink` site, which now needs an `isinstance` assert to narrow the union.  Alternative would be `shrink_many([k1, k2, ...])` but the user explicitly asked for varargs.  The pattern matches `min`/`max`/`os.path.join`, so it's idiomatic Python, but callers in library code should always handle both branches.

**Batched LCP algorithm correctness** rests on a fact that's worth stating in a comment: in a sorted key list, the longest prefix any *other* key shares with a target is the LCP with one of its two immediate sorted neighbours.  This is because lexicographic adjacency ⇒ prefix adjacency.  Without this comment, the line `n = max(4, max(lcp_left, lcp_right) + 1)` reads like a heuristic rather than a tight bound.  Will add the comment in a follow-up if you agree.

**`Cache.shrink` dispatch is "first sub-storage wins"**: calls before values.  Pre-#577 it returned the longer (safer) of the two; now it returns whatever `calls.shrink` produces for keys owned by calls.  Since call keys and value keys are disjoint by construction (different digest domains), this is correct except under a SHA-256 collision — which we treat the same way we treat every other SHA-256 invariant.  Worth being aware of.

**`KeyManagement.shrink` no longer enters `_operation_context`**.  The batched version reads `self.list()` outside any per-key lock.  For pickle / HDF backends with file locks this means we read the directory listing without holding the lock.  Acceptable because `shrink` is purely informational and the warning in its docstring already says results can be invalidated by concurrent writes, but it's a deliberate divergence from the old code path.

### Foot-guns and pitfalls

**One ambiguous key kills the entire batch in `QueryIterator.table`**.  The `except AmbiguousDigestError: pass` falls back to *full digests for every row*, not just the offending one.  In practice this needs two calls whose digests agree on the first 63 chars, which is implausible, but a per-key fallback (catching at the `_shrink_one` granularity) would be more graceful.  Would require restructuring the batched path.

**`shrink_keys=False` is only discoverable via the docstring.**  With the default flipped, users who need stable full digests have to know to opt out.  A logged DeprecationWarning would be heavy-handed; I'd rather leave it and rely on the CHANGELOG.

**Reaching into `LazyCall._cache`**.  `QueryIterator.table` calls `items[0]._cache.shrink(*keys)` — a private attribute access that the original code already did.  If `LazyCall` ever changes its internal cache reference (e.g. lazy-resolves via `weakref`), table breaks silently.  No fix here, but worth a note for future refactors.

**The variadic API's amortisation benefit only kicks in when keys are batched.**  Calling `cache.shrink(k1); cache.shrink(k2); ...` separately pays one `list()` per call.  The `c._cache.shrink(*keys)` call in `table` is what unlocks the speedup; other call sites (e.g. `transfer`'s log line in `QueryIterator.transfer`, `Digest.shrink`) still shrink one key at a time and don't benefit.  Probably fine — those aren't hot paths — but worth knowing if someone re-uses the API in a loop.

**Empty cache shrink semantics**: `cache.shrink()` (no args) raises `TypeError`; `cache.shrink(k)` on an empty cache raises `KeyError`.  Both fine, but inconsistent in error class — the no-args case could plausibly return `()` for zero-arg convenience.  Picked the strict route.

**Algorithmic change to `KeyManagement.shrink`** affects every storage backend, not just the table path.  Reads `list()` instead of iterating `expand(prefix)`.  For backends where `list()` is more expensive than `expand` (none I can think of, but `Sql.list` does fetch every key as a single column), this is a regression on the *single-key* path.  Mitigated by the fact that the bisect is O(log N) per key after a single sort; the only worse case is "one key, huge cache", which is rare.

**`CacheStack.shrink` calls `cache.contains(k)` per layer × per key** to decide which keys to batch into each layer.  For an N-key query against an M-layer stack of Sql backends, that's N×M `SELECT EXISTS(...)` round-trips, *plus* one batched shrink per layer.  Could be improved with a `contains(*keys)` API, but out of scope here.

### Tests

- The "longest of two storages" coverage went away with #577 — replaced by `dispatches_to_call_storage` / `dispatches_to_value_storage` / `missing_key_raises`.  The substantive correctness property survives (each sub-storage's shrink is exercised).
- I did not add a batched-shrink-of-mixed-keyspaces test (e.g. one call key + one value key).  Would be a clean unit test to write before merge.
- No test for `shrink(*keys)` against a `CacheStack` directly — the existing `CacheStack.shrink` test still uses single-arg.  Would be worth adding.

### Where I'd push back on myself

The arity-dispatched return type is the design decision I'm least happy with.  It's idiomatic Python and you asked for `shrink(*keys)` to stay backwards compatible, but in a strongly-typed codebase the union return type leaks into every caller.  An alternative is to keep `shrink(key) -> Digest` as-is and add a separate batched method (`shrink_many(keys) -> tuple[Digest, ...]`).  Two methods, two return types, no surprise.  Happy to refactor if you want — the call sites are localised.

---

pmrv:
> ## Self-review of the full change set
> ### Design decisions worth flagging
> 
> **Default `shrink_keys=True`** is a silent behaviour change. Anyone calling `cache.table().index` will now see ~6–8 character prefixes where they used to see 64-character digests. `cache.load(short)` still works (the storage's `_normalize_key` re-expands), but any downstream code that _compares_ index entries to keys retrieved from another source (e.g. `set(table.index) == set(c.calls.list())`) silently breaks. Issue #319's regression test was the canary — it now passes `shrink_keys=False`. Worth a CHANGELOG entry; AGENTS.md `QueryIterator` row may want an update too.

that's ok

> **Variadic `shrink(*keys)` with arity-dispatched return type** (`Digest` for 1, `tuple[Digest, ...]` for N) keeps the public API backwards compatible but is a known Python foot-gun: `cache.shrink(*some_list)` where `some_list` happens to be length 1 returns a `Digest`, not a `tuple[Digest]`. Type checkers can't help — see the `Digest.shrink` site, which now needs an `isinstance` assert to narrow the union. Alternative would be `shrink_many([k1, k2, ...])` but the user explicitly asked for varargs. The pattern matches `min`/`max`/`os.path.join`, so it's idiomatic Python, but callers in library code should always handle both branches.

type dispatch in the same of the same name using typing overload to keep the type nazis quiet.

> **Batched LCP algorithm correctness** rests on a fact that's worth stating in a comment: in a sorted key list, the longest prefix any _other_ key shares with a target is the LCP with one of its two immediate sorted neighbours. This is because lexicographic adjacency ⇒ prefix adjacency. Without this comment, the line `n = max(4, max(lcp_left, lcp_right) + 1)` reads like a heuristic rather than a tight bound. Will add the comment in a follow-up if you agree.

do so
 
> **`Cache.shrink` dispatch is "first sub-storage wins"**: calls before values. Pre-#577 it returned the longer (safer) of the two; now it returns whatever `calls.shrink` produces for keys owned by calls. Since call keys and value keys are disjoint by construction (different digest domains), this is correct except under a SHA-256 collision — which we treat the same way we treat every other SHA-256 invariant. Worth being aware of.

if you find a unobtrusive place in the docs put it there, but not very worried about it
 
> **`KeyManagement.shrink` no longer enters `_operation_context`**. The batched version reads `self.list()` outside any per-key lock. For pickle / HDF backends with file locks this means we read the directory listing without holding the lock. Acceptable because `shrink` is purely informational and the warning in its docstring already says results can be invalidated by concurrent writes, but it's a deliberate divergence from the old code path.

re-enter the context for now

> ### Foot-guns and pitfalls

> **One ambiguous key kills the entire batch in `QueryIterator.table`**. The `except AmbiguousDigestError: pass` falls back to _full digests for every row_, not just the offending one. In practice this needs two calls whose digests agree on the first 63 chars, which is implausible, but a per-key fallback (catching at the `_shrink_one` granularity) would be more graceful. Would require restructuring the batched path.

we go in with full length 64 char digests, ambiguity is impossible module sha collision, no?
 
> **`shrink_keys=False` is only discoverable via the docstring.** With the default flipped, users who need stable full digests have to know to opt out. A logged DeprecationWarning would be heavy-handed; I'd rather leave it and rely on the CHANGELOG.

I (the only user) am fine with this.

> **Reaching into `LazyCall._cache`**. `QueryIterator.table` calls `items[0]._cache.shrink(*keys)` — a private attribute access that the original code already did. If `LazyCall` ever changes its internal cache reference (e.g. lazy-resolves via `weakref`), table breaks silently. No fix here, but worth a note for future refactors.

annoying, probably best to give the query iterator a cache reference in the original BaseCache.query()
 
> **The variadic API's amortisation benefit only kicks in when keys are batched.** Calling `cache.shrink(k1); cache.shrink(k2); ...` separately pays one `list()` per call. The `c._cache.shrink(*keys)` call in `table` is what unlocks the speedup; other call sites (e.g. `transfer`'s log line in `QueryIterator.transfer`, `Digest.shrink`) still shrink one key at a time and don't benefit. Probably fine — those aren't hot paths — but worth knowing if someone re-uses the API in a loop.

+1

> **Empty cache shrink semantics**: `cache.shrink()` (no args) raises `TypeError`; `cache.shrink(k)` on an empty cache raises `KeyError`. Both fine, but inconsistent in error class — the no-args case could plausibly return `()` for zero-arg convenience. Picked the strict route.

both are fine
 
> **Algorithmic change to `KeyManagement.shrink`** affects every storage backend, not just the table path. Reads `list()` instead of iterating `expand(prefix)`. For backends where `list()` is more expensive than `expand` (none I can think of, but `Sql.list` does fetch every key as a single column), this is a regression on the _single-key_ path. Mitigated by the fact that the bisect is O(log N) per key after a single sort; the only worse case is "one key, huge cache", which is rare.

iirc expand calls list internally in teh default impl (correct me if wrong), so nothing lost imo.  list simple alternatives if possible

> **`CacheStack.shrink` calls `cache.contains(k)` per layer × per key** to decide which keys to batch into each layer. For an N-key query against an M-layer stack of Sql backends, that's N×M `SELECT EXISTS(...)` round-trips, _plus_ one batched shrink per layer. Could be improved with a `contains(*keys)` API, but out of scope here.
> ### Tests
> 
>     * The "longest of two storages" coverage went away with [refactor(caches): dispatch Cache.shrink by which sub-storage owns the key #577](https://github.com/pmrv/fleche/pull/577) — replaced by `dispatches_to_call_storage` / `dispatches_to_value_storage` / `missing_key_raises`.  The substantive correctness property survives (each sub-storage's shrink is exercised).
> 
>     * I did not add a batched-shrink-of-mixed-keyspaces test (e.g. one call key + one value key).  Would be a clean unit test to write before merge.

I don't see the utility.  add a docstring that such madness is undefined behavior.

>     * No test for `shrink(*keys)` against a `CacheStack` directly — the existing `CacheStack.shrink` test still uses single-arg.  Would be worth adding.

indeed

> 
> ### Where I'd push back on myself
> 
> The arity-dispatched return type is the design decision I'm least happy with. It's idiomatic Python and you asked for `shrink(*keys)` to stay backwards compatible, but in a strongly-typed codebase the union return type leaks into every caller. An alternative is to keep `shrink(key) -> Digest` as-is and add a separate batched method (`shrink_many(keys) -> tuple[Digest, ...]`). Two methods, two return types, no surprise. Happy to refactor if you want — the call sites are localised.
> 

pmrv:
Pushed af1fbfe addressing the review. Summary of what landed:

| Feedback | Change |
|---|---|
| arity-dispatch via `typing.overload` | Added `@overload` signatures on every `shrink(*keys)` site (`KeyManagement`, `BaseCache`, `Cache`, `CacheStack`, `CacheWrapper`); dropped the `isinstance` assert in `Digest.shrink`. `ty check` clean. |
| LCP correctness comment | Added a paragraph above `_shrink_one` explaining why lexicographic adjacency implies prefix adjacency, so `max(LCP_left, LCP_right) + 1` is provably the shortest unambiguous length. |
| SHA-collision dispatch note | Folded into `BaseCache.shrink` docstring as "mixing call keys and value keys in a single call is undefined behaviour". |
| Re-enter `_operation_context` | `KeyManagement.shrink` opens a `contextlib.ExitStack` and enters the per-key context for every key before snapshotting `list()`. |
| Ambiguity is impossible mod SHA collision | Removed the `try/except AmbiguousDigestError: pass` in `QueryIterator.table` — let it propagate so a broken hash invariant surfaces. |
| `LazyCall._cache` reach | `QueryIterator` grows an optional `cache` field, set by `BaseCache.query()` and propagated through `take`/`skip`/`filter`/`sorted`/`unique`/`groupby`. `.table()` uses `self.cache` (with the `_cache` fallback only for manually-constructed iterators). |
| `shrink_many` refactor | Skipped — overloads make the union return type invisible to callers, which was my only real concern. |
| Undefined behaviour for mixed-keyspace shrink | Documented in `BaseCache.shrink` docstring (last paragraph of the args section). |
| `CacheStack` variadic test | Two new tests in `test_expand_shrink.py`: `_batches_per_layer` (both layers contain both keys; one batched call per layer) and `_skips_layers_without_key` (per-layer key filtering). |

Skipped per your responses: CHANGELOG/AGENTS entry, dedicated mixed-keyspace test (covered by the docstring), `shrink_many` two-method split.

906 tests pass locally; `ty check src/` clean.

---

pmrv:
@claude make a note in an appropriate issue that we should probably include shrink/expand in hte benchmark suite.

PR #589, opened 2026-05-28
fix(config): intern the default cache under None so it is reused

pmrv:
`load_cache_config(None)` interned the resolved default cache under its resolved name (`"default"` for an inline `[default.cache]` table, the alias name for a string default) but not under `None`, so every call rebuilt the cache — re-spawning a default `SshCache` subprocess and reopening file handles instead of reusing the interned instance. Fix: always intern under the requested `name`.

https://claude.ai/code/session_01RjdebitoJvSPDoCF7u5Q7U

### DIFF
--- a/src/fleche/config.py
+++ b/src/fleche/config.py
@@ -536,7 +536,9 @@ def load_cache_config(name: str | None = None) -> caches.BaseCache:
             return _default_memory_cache(None, "no default cache configured")
         default_cache = config["default"]["cache"]
         if isinstance(default_cache, str):
-            return load_cache_config(default_cache)
+            cache = load_cache_config(default_cache)
+            _live_caches[None] = cache
+            return cache
         cache_name, cache_config = "default", default_cache
     else:
         if name not in config:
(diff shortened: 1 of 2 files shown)


src/fleche/config.py
pmrv, line 1: wouldn't it be easier to ensure name and cache_name are the same and well set ?

pmrv: Good call — done in 17a642d. I dropped the separate `cache_name` entirely and now always intern under the requested `name` (the same key callers look up by). The old `"default"` key was never actually usable anyway: `load_cache_config("default")` would try to build a cache out of the `[default]` table and fail. For a string-alias default, the recursive `load_cache_config(alias)` interns under the alias and the outer call interns the same instance under `None`, so both keys map to one object.

---

pmrv (commented):
rewrite op succintly

pmrv:
@claude Now I also want `cache('default')` to work in the expected way, i.e. it sets active whatever is the default conf, either named or inline.

PR #590, opened 2026-05-28
feat(config): make 'default' a special name resolving to the configured default cache

pmrv:
`cache('default')` now activates whatever is the configured default cache (string-alias or inline-table), same as calling `cache()` with no arguments.

Follows on from #589 which fixed the `None`-key interning.


pmrv:
document this in `fleche.cache` as well.  Your comment 'same as calling cache() with no argument` is wrong.  That returns the currently active cache instead.

pmrv:
> document this in `fleche.cache` as well. Your comment 'same as calling cache() with no argument` is wrong. That returns the currently active cache instead.

@claude

PR #598, opened 2026-05-29
refactor: deduplicate transfer policy between BaseCache.transfer and QueryIterator.transfer (#592)

### LINKED ISSUE #592: Refactor: deduplicate transfer policy between BaseCache.transfer and QueryIterator.transfer
## Location

- `src/fleche/caches.py:57-80` — `BaseCache.transfer`
- `src/fleche/query.py:181-201` — `QueryIterator.transfer` (added in PR #566)

## Problem

The "replay calls into another cache" conflict policy is implemented twice with near-identical bodies:

```python
# BaseCache.transfer
tpl = call.QueryCall(name=None, arguments=None, metadata=None, module=None, version=None, result=None)
for c in self.query(tpl):
    c = c.fetch()
    key = c.to_lookup_key()
    conflict = not overwrite and other.contains(key)
    if not conflict:
        other.save(c)
    if pop:
        if conflict:
            logger.warning("Not evicting %s from source: already exists in target ...", key)
        else:
            self.evict(key)

# QueryIterator.transfer
for c in self:
    key = c.to_lookup_key()
    conflict = not overwrite and target.contains(key)
    if conflict:
        logger.warning("Not transferring %s: already exists in target ...", target.shrink(key))
        continue
    target.save(c.fetch())
    if pop:
        c._cache.evict(key)
```

Both encode the same `pop` / `overwrite` / conflict-skip semantics. `BaseCache.transfer` even builds a fully-wildcard `QueryCall` and iterates it — which is exactly what `self.query()` (no args) already returns as a `QueryIterator`. So one of these is a re-implementation of the other.

Keeping two copies means every behavioural tweak (and every bug fix) has to be mirrored, and they have **already drifted** in small ways:

- Warning message differs (full `key` vs `target.shrink(key)`); `BaseCache.transfer` warns only in the `pop` branch, `QueryIterator.transfer` warns on every skipped conflict.
- Source eviction differs (`self.evict` vs `c._cache.evict`) — equivalent for `self.query()` but spelled differently.

This duplication is also why the planned TOCTOU lock fix in #452 (`contains→save` race in `BaseCache.transfer`) would otherwise need to be applied in two places to be correct.

## Suggested refactor

Make `BaseCache.transfer` a thin delegator:

```python
def transfer(self, other, pop=False, overwrite=False):
    self.query().transfer(other, pop=pop, overwrite=overwrite)
```

`cache.query()` with no template already yields a fully-wildcard `QueryIterator` bound to `self`, so `c._cache.evict(...)` resolves to `self.evict(...)`. Reconcile the two warning messages into one (prefer the shrunk-key form, but only shrink when actually warning to avoid the cost on the happy path). The conflict-skip-warning behaviour should be unified for both `pop=True` and `pop=False`.

## Pros

- Single source of truth for the transfer/conflict policy; #452's lock fix then lands in exactly one method.
- Removes ~20 lines and the redundant hand-spelled wildcard `QueryCall`.
- `QueryIterator.transfer` is already the more general form (it can run post-filter), so collapsing onto it is the natural direction.

## Cons / risks

- `QueryIterator.transfer` calls `target.shrink(key)` for its warning, which is an extra cache op; ensure that only runs on the conflict path (it already does) and ideally guard behind the logger level.
- Tests asserting on the exact warning text (if any) would need updating.
- `transfer` is covered by tests for both surfaces; the equivalence (`self.evict` ≡ `c._cache.evict` for `self.query()`) should be pinned by a regression test before collapsing.

Low. Pure consolidation onto an existing, more-general implementation; no new dependencies.

---

pmrv:
Closes #592

## Summary

- Pins the `_cache is source` invariant with a regression test before collapsing
- Reduces `BaseCache.transfer` to a one-line delegator onto `QueryIterator.transfer`
- Unifies conflict-skip warning (now always emitted, shrunk key, `fleche.query` logger)

## Test plan

- [x] `tests/regression/test_issue_592.py` — 7 new tests pinning functional equivalence
- [x] Existing `test_base_cache_transfer*` tests still pass
- [x] Existing `test_transfer_*` tests in `test_query_iterator.py` still pass
- [x] Full unit + regression suite: 959 passed

### DIFF
--- a/tests/regression/test_issue_592.py
+++ b/tests/regression/test_issue_592.py
@@ -1,0 +1,140 @@
+"""Regression test for issue #592: BaseCache.transfer and QueryIterator.transfer equivalence.
+
+Pins two properties that make it safe to collapse BaseCache.transfer into a
+one-line delegator onto QueryIterator.transfer:
+
+1. cache.query() with no template binds ``_cache`` to the source cache on
+   every yielded LazyCall, so ``self.evict(key)`` ≡ ``c._cache.evict(key)``.
+2. Both surfaces produce identical data decisions (save/skip/evict) across all
+   combinations of pop/overwrite/conflict.
+"""
+
+import pytest
+from fleche.call import Call
+from fleche.caches import Cache
+from fleche.storage.memory import ValueMemory, CallMemory
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _cache(*calls):
+    c = Cache(values=ValueMemory({}), calls=CallMemory({}))
+    for call in calls:
+        c.save(call)
+    return c
+
+
+_call_a = Call(name="f", arguments={"x": 1}, result=10)
+_call_b = Call(name="f", arguments={"x": 2}, result=20)
+_call_a_conflict = Call(name="f", arguments={"x": 1}, result=999)  # same key as _call_a
+
+
+# ---------------------------------------------------------------------------
+# 1. _cache identity on wildcard query
+# ---------------------------------------------------------------------------
+
+def test_wildcard_query_lazycall_cache_is_source():
+    """Every LazyCall from cache.query() has _cache pointing to that cache.
+
+    This is the invariant that makes self.evict(key) == c._cache.evict(key)
+    after the refactor.
+    """
+    src = _cache(_call_a, _call_b)
+    for lc in src.query():
+        assert lc._cache is src
+
+
+# ---------------------------------------------------------------------------
+# 2. Functional data equivalence: no conflict
+# ---------------------------------------------------------------------------
+
+def test_both_surfaces_copy_all_calls_no_conflict():
+    """Plain transfer: both surfaces save all source calls in the target."""
+    src1, dst1 = _cache(_call_a, _call_b), _cache()
+    src2, dst2 = _cache(_call_a, _call_b), _cache()
+
+    src1.transfer(dst1)
+    src2.query().transfer(dst2)
+
+    for call in (_call_a, _call_b):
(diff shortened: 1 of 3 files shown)


tests/regression/test_issue_592.py
```diff
@@ -0,0 +1,140 @@
...
+
+def test_both_surfaces_overwrite_target_on_conflict():
+    """overwrite=True: both surfaces replace the target entry."""
+    src1, dst1 = _cache(_call_a), _cache(_call_a_conflict)
+    src2, dst2 = _cache(_call_a), _cache(_call_a_conflict)
+
+    src1.transfer(dst1, overwrite=True)
+    src2.query().transfer(dst2, overwrite=True)
+
+    assert dst1.load(_call_a.to_lookup_key()).result == 10
```
pmrv: Presumably 10 == _call_a.result, but != _call_a_conflict.result.  Make that more obvious.

tests/regression/test_issue_592.py
```diff
@@ -0,0 +1,140 @@
...
+    """Conflict + overwrite=False: target entry kept on both surfaces."""
+    src1, dst1 = _cache(_call_a), _cache(_call_a_conflict)
+    src2, dst2 = _cache(_call_a), _cache(_call_a_conflict)
+
+    src1.transfer(dst1, overwrite=False)
+    src2.query().transfer(dst2, overwrite=False)
+
+    # Target retains conflicting value on both surfaces
+    assert dst1.load(_call_a.to_lookup_key()).result == 999
+    assert dst2.load(_call_a.to_lookup_key()).result == 999
```
pmrv: see below

pmrv (commented):
@claude revise

PR #599, opened 2026-05-29
feat: add LazyCall.to_digested_call() — public inverse of DigestedCall.fetch()

### LINKED ISSUE #586: emote._strip_cache reaches into LazyCall private fields — promote to a public LazyCall.to_digested_call()
fleche.remote._strip_cache builds the cache-free DigestedCall shadow of a LazyCall by reading its private attributes directly:

def _strip_cache(lc: LazyCall) -> DigestedCall:
    return DigestedCall(
        name=lc.name,
        arguments=dict(lc._arguments),   # private
        result=lc._result,               # private
        metadata=dict(lc.metadata),
        module=lc.module,
        version=lc.version,
        code_digest=lc.code_digest,
    )

This is needed on the server side because a LazyCall's _cache reference can't travel across the wire (it points back into the server process); the client re-binds the returned DigestedCall to itself via DigestedCall.fetch().

The problem: _arguments / _result are LazyCall internals. If that representation changes (e.g. the two fields are merged into a single _digests mapping), _strip_cache breaks silently — no test asserts on the internal layout, and call.py and remote.py would drift apart with nothing flagging it.

Proposed fix: add a public LazyCall.to_digested_call() -> DigestedCall method in call.py that returns the digest-only shadow (the inverse of the existing DigestedCall.fetch()), and have _strip_cache call it. The coupling then lives next to the data it depends on, and a LazyCall representation change forces the method to be updated in lockstep.

Notes:

    Low priority / not user-facing; surfaced in the #551 self-review (item F10).
    Worth a small LazyCall ↔ DigestedCall round-trip test once the method exists, asserting field equality so future refactors of the lazy representation can't silently regress it.

pmrv:
Closes #586.

Promotes the `LazyCall`→`DigestedCall` conversion from `remote._strip_cache` (which read private `_arguments`/`_result` directly) into a public `LazyCall.to_digested_call()` method, so representation changes in `call.py` are caught at the source rather than silently breaking `remote.py`.

### DIFF
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -397,7 +397,7 @@ class LazyCall:
             result=self.result
         )
 
-    def to_digested_call(self) -> "DigestedCall":
+    def detach(self) -> "DigestedCall":
         """Return the cache-free :class:`DigestedCall` shadow of this :class:`LazyCall`.
 
         The inverse of :meth:`DigestedCall.fetch`: strips the ``_cache``
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -397,6 +397,23 @@ class LazyCall:
             result=self.result
         )
 
+    def to_digested_call(self) -> "DigestedCall":
+        """Return the cache-free :class:`DigestedCall` shadow of this :class:`LazyCall`.
+
+        The inverse of :meth:`DigestedCall.fetch`: strips the ``_cache``
+        reference so the result can cross process boundaries (e.g. over the
+        wire in :mod:`fleche.remote`) without carrying a live cache pointer.
+        """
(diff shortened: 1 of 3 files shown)


src/fleche/call.py
```diff
@@ -397,6 +397,23 @@ def fetch(self) -> Call:
             result=self.result
         )
 
+    def to_digested_call(self) -> "DigestedCall":
```
pmrv: @claude Name needs to be snappier, like .fetch, or .stash.  List some examples, not .digest, since we are using this already.

pmrv: @claude detach it is.

PR #600, opened 2026-05-29
refactor: consolidate shrink boilerplate via Template Method pattern

### LINKED ISSUE #595: Refactor: consolidate repeated batched-shrink scaffolding across cache/storage layers
## Location

The variadic `shrink(*keys)` contract is re-spelled across the codebase:

**`@overload` stub pair + signature** (3 lines, copy-pasted in 6 classes):
- `src/fleche/caches.py:106-111` (`BaseCache`, abstract)
- `src/fleche/caches.py:299-303` (`Cache`)
- `src/fleche/caches.py:457-462` (`CacheWrapper`)
- `src/fleche/caches.py:580-584` (`CacheStack`)
- `src/fleche/remote.py:745-752` (`SshCache`)
- `src/fleche/storage/base.py:122-126` (`KeyManagement`)

**The partition / call-sub / repack body** (`if not keys: raise TypeError`, partition keys by owner, call each sub-`shrink`, then `out[0] if len(keys) == 1 else out`):
- `Cache.shrink` (`caches.py:303-325`)
- `CacheStack.shrink` (`caches.py:584-603`)
- `KeyManagement.shrink` (`storage/base.py:126-144`)

The single-vs-tuple unwrap idiom (`if len(keys) == 1: r = (r,)`) also recurs in `QueryIterator.table` (`query.py:256-259`) and the remote `_dispatch` shrink path.

## Problem

Three different `shrink` implementations each re-derive the same boilerplate:

```python
if not keys:
    raise TypeError("shrink() requires at least one key")
... # partition / per-sub shrink, with the single-key special case:
if len(present) == 1:
    r = (r,)
...
return out[0] if len(keys) == 1 else out
```

The "pack one result or many into the right return shape" rule lives in 4+ places; the empty-keys guard in 3; the overload stub pair in 6. A change to the batched contract (e.g. accepting an empty call to return `()`) means touching all of them.

## Suggested refactor

Two small shared helpers in one module (e.g. `storage/base.py` or a `_shrink` util):

```python
def _require_keys(keys):
    if not keys:
        raise TypeError("shrink() requires at least one key")

def _as_result(keys, per_key):
    """per_key: dict-or-seq of one Digest per input key, in input order."""
    out = tuple(per_key[k] for k in keys)
    return out[0] if len(keys) == 1 else out
```

Each concrete `shrink` keeps only its own partition logic (which is genuinely different — `Cache` partitions call vs value, `CacheStack` fans out across layers, `KeyManagement` slices a sorted snapshot) and routes the packing/guard through the shared helpers.

## Pros

- One definition of the batched-`shrink` return contract instead of 4 hand-rolled copies.
- The empty-keys guard and single/tuple unwrap stop drifting.
- Makes it obvious that the *only* difference between the implementations is the partition strategy.

## Cons / risks

- The `@overload` stub pair **cannot** be inherited in a way static checkers (`ty`) honour, so the 3-line signature duplication across the 6 classes is largely unavoidable; this refactor only removes the *body* duplication, not the type-stub duplication. Worth stating up front so reviewers don't expect the overloads to disappear.
- Low net LOC; this is a low-hanging-fruit cleanup, not a structural win.
- `shrink` is exercised broadly (cross-backend `value_storage`/`call_storage` fixtures, `test_expand_shrink.py`, `test_short_digest.py`); behaviour must stay identical.

Low. Mechanical, behaviour-preserving, no new dependencies.

---

pmrv:
Implements the Template Method pattern for `shrink` across the `BaseCache` hierarchy, as suggested in #595.

`BaseCache.shrink` is now concrete and owns the empty-keys guard and single/tuple return-value unwrap. Subclasses implement the new abstract `_shrink(*keys) -> tuple[Digest, ...]` with only their partition logic.

Closes #595

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -455,10 +455,9 @@ class CacheWrapper(BaseCache):
         return self.cache.expand(key)
 
     def _shrink(self, *keys: Digest | str) -> "tuple[Digest, ...]":
-        r = self.cache.shrink(*keys)
         if len(keys) == 1:
-            r = (r,)
-        return r
+            return (self.cache.shrink(keys[0]),)
+        return self.cache.shrink(*keys)
 
     def _query(self, call: call.QueryCall) -> Iterable[LazyCall]:
         return self.cache.query(call)
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -107,7 +107,6 @@ class BaseCache(ABC):
     def shrink(self, key: Digest | str, /) -> Digest: ...
     @overload
     def shrink(self, key: Digest | str, /, *keys: Digest | str) -> "tuple[Digest, ...]": ...
-    @abstractmethod
     def shrink(self, *keys: Digest | str) -> "Digest | tuple[Digest, ...]":
         """
         Find the shortest substring(s) that unambiguously reference each call.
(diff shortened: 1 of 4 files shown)


pmrv:
@claude If the ty inference is correct this is a real bug, write a test and confirm.

Issue #569, opened 2026-05-25
Refactor: extract `OperationContext` base shared by `KeyManagement` and `BaseCache`


pmrv:
**Parent:** #444

## Goal

Extract `_operation_context` into a standalone base class that both `KeyManagement` (storage layer) and `BaseCache` (cache layer) inherit from, so the same thread-safety mixins (`SerializingMixin`, `PerKeyLockMixin`, and any future ones) attach to either layer without duplication.

Motivated by the audit in #444: the storage layer is thread-safe (#449 closed) but the cache layer has TOCTOU windows that storage locks cannot see across (#217, #448, #450, #451, #452, #454). Today `_operation_context` lives only on `KeyManagement` (`storage/base.py:58–79`); `BaseCache` has no equivalent hook, so the storage mixins cannot be reused at the cache layer as-is.

## Proposed shape

```python
class OperationContext(ABC):
    @contextlib.contextmanager
    def _operation_context(self, key, *, intent=None):
        yield                                    # default no-op

class KeyManagement(OperationContext): ...       # storage, unchanged otherwise
class BaseCache(OperationContext): ...           # new: cache enters the hook too

class SerializingMixin(OperationContext): ...    # mixin descends from common base
class PerKeyLockMixin(OperationContext): ...     # → attachable to either layer
```

Existing MROs stay valid: storages reach `OperationContext` through `KeyManagement`; mixins reach it directly; concrete `class CallMemory(PerKeyLockMixin, CallMixin, MemoryBackend)` is unchanged.

The `intent` parameter is the design from #453 — recording it here so the new base lands the signature once.

## Decisions taken

1. **Which `BaseCache` methods enter the context.** Per-key methods (`save`, `load`, `load_value`, `evict`, `contains`, `expand`, `shrink`) acquire with `intent=None`. Whole-cache ops (`_query`, `gc`, `redigest`, `transfer`, `table`) acquire with `intent="scan"`. `save(call)` derives the key from `call.to_lookup_key()` and enters the hook inside `save`, mirroring `CallMixin.save`. Forwarding wrappers (`CacheWrapper`, `ReadOnlyMixin`, `FilteringMixin`, `RefreshingCache`) do **not** enter their own context — they hold no state of their own and delegate; the wrapped cache enters the hook.
[...]

PR #601, opened 2026-05-29
feat(storage): add `intent` parameter to `_operation_context` (step 1 of #569)

pmrv:
Adds `intent: str = "write"` (keyword-only) to `KeyManagement._operation_context` and all overriding implementations (`SerializingMixin`, `PerKeyLockMixin`, `Sql`).

This is step 1 of #569: pins the `intent` signature once at the storage layer. Currently the only defined value is "write", which is also the default, so all existing callers continue to work unchanged.


pmrv:
@claude intent is an (Str if available)Enum.

PR #603, opened 2026-05-29
refactor(metadata): replace if/elif chain with @configurable decorator registry

### LINKED ISSUE #593: Refactor: replace load_default_metadata if/elif chain with a MetaData name registry
## Location

`src/fleche/config.py:238-262` — `load_default_metadata`.

## Problem

`MetaData` is designed as a pluggable ABC (`metadata.py`), but the config loader resolves the configured metadata names through a closed, hand-maintained `if/elif` chain:

```python
for name in meta_names:
    if name == "Tags":
        raise ValueError("Tags metadata cannot be configured from the config file.")
    elif name == "Runtime":
        meta_objects.append(metadata.Runtime())
    elif name == "Environment":
        meta_objects.append(metadata.Environment())
    elif name == "Git":
        meta_objects.append(metadata.Git())
    else:
        raise ValueError(f"Unknown metadata type in config: {name}")
```

Issues:

- **Every new built-in metadata is a manual touch-point here.** `Resources` (#567) is already planned and will need another `elif`; it is easy to add the class and forget to wire it into the loader.
- The mapping name→class is knowledge that is essentially already implied by the class itself; the storage side already solves the analogous problem with a lookup table (`_STORAGE_NAME_MAPPING`), so metadata is inconsistent with its sibling subsystem.
- Note the config keys are the **class names** (`"Runtime"`), not the `.name` attribute (`"runtime"`), so the registry must key on whichever is chosen deliberately.

## Suggested refactor

Introduce an explicit registry — preferably co-located in `metadata.py` next to the classes — mapping the config name to a zero-arg constructor, and drive the loader from it:

```python
# metadata.py
CONFIGURABLE: dict[str, type[MetaData]] = {
    "Runtime": Runtime,
    "Environment": Environment,
    "Git": Git,
    # "Tags" deliberately excluded: requires arguments
}
```

```python
# config.py
for name in meta_names:
    cls = metadata.CONFIGURABLE.get(name)
    if cls is None:
        raise ValueError(f"Unknown or non-configurable metadata type in config: {name}")
    meta_objects.append(cls())
```

Adding a new built-in then becomes a single entry next to its class definition.

## Pros

- New built-in metadata (e.g. #567 `Resources`) becomes a one-line registration beside the class, not a separate edit in `config.py`.
- Mirrors the existing `_STORAGE_NAME_MAPPING` pattern — consistent design across config subsystems.
- Single list of "what can be configured" that docs / a future `fleche config check` can introspect; feeds naturally into the `[metadata]` section of the config redesign (#568).

## Cons / risks

- Auto-discovery via `MetaData.__subclasses__()` is tempting but fragile (requires the modules to be imported, would pick up user subclasses and argument-taking classes like `Tags`); an **explicit** registry is the safer choice and keeps `Tags` deliberately excluded.
- Very small net LOC change; value is in maintainability/consistency, not line count.

Low. Localised change with existing coverage in `tests/unit/config/` and `tests/unit/metadata/`.

---

pmrv:
Introduce a `configurable` class decorator in `metadata.py` that sets `cls.name` and registers the class in a `CONFIGURABLE` dict, replacing the closed if/elif chain in `config.load_default_metadata`.

Closes #593

### DIFF
--- a/src/fleche/config.py
+++ b/src/fleche/config.py
@@ -248,16 +248,10 @@ def load_default_metadata():
 
     meta_objects = []
     for name in meta_names:
-        if name == "Tags":
-            raise ValueError("Tags metadata cannot be configured from the config file.")
-        elif name == "Runtime":
-            meta_objects.append(metadata.Runtime())
-        elif name == "Environment":
-            meta_objects.append(metadata.Environment())
-        elif name == "Git":
-            meta_objects.append(metadata.Git())
-        else:
-            raise ValueError(f"Unknown metadata type in config: {name}")
+        cls = metadata.CONFIGURABLE.get(name)
+        if cls is None:
+            raise ValueError(f"Unknown or non-configurable metadata type in config: {name}")
+        meta_objects.append(cls())
 
     return tuple(meta_objects)
 
--- a/src/fleche/metadata.py
+++ b/src/fleche/metadata.py
@@ -65,16 +65,8 @@ class MetaData(ABC):
         """
         ...
 
-    @property
-    @abstractmethod
-    def name(self) -> str:
-        """
-        The unique name of this metadata type.
-
-        Returns:
-            str: The name of the metadata type.
-        """
-        ...
+    name: str
+    """The unique name of this metadata type."""
 
 
 CONFIGURABLE: dict[str, type["MetaData"]] = {}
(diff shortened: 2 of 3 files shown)


pmrv:
@claude Garçon!

PR #616, opened 2026-06-02
perf: avoid double H5Bag file open on load

### LINKED ISSUE #527: Perf audit: 2026-05-28
**Verdict**: No threshold regressions in the 2026-05-28 run vs the HEAD CSV; hot-spots #1 (hex encoding) and #2 (per-element SHA256 allocation) remain open; SqlMemory evict improved 4.3× (990 µs → 230 µs), confirming the full benefit of the PR #535 bulk-DELETE fix is now reflected in a fresh run.

---

## Regressions (≥1.5× and ≥1 µs, or ≥3×) — 2026-05-28 run vs HEAD

| Topic | Config | Workload | Function | HEAD | Fresh | Ratio | Note |
|---|---|---|---|---|---|---|---|
| Digest | | String (len>100) | digest | 230 µs | 350 µs | **1.52×** | Hypothesis random-length input noise (>100 chars); not a code regression |

No baseline rows missing from the fresh run.

### Resolved since 2026-05-21

| Issue | Fix | Evidence |
|---|---|---|
| Hot-spot #3: `SqlFile evict` ORM materialisation | Bulk DELETE at `sql.py:314–320`, PR #535 | SqlFile evict: 1.2 ms → 990 µs; SqlMemory evict: 990 µs → 230 µs (4.3× improvement — full benefit now visible) |

---

### 1. `_digest_bytes` feeds 64-byte hex strings into parent hashes — `digest.py:303`

```python
return m.hexdigest().encode()   # 64 ASCII bytes fed to parent m.update()
```

Every recursive `_digest_bytes` call returns `m.hexdigest().encode()` — 64 UTF-8 hex bytes — which the parent SHA256 ingests via `m.update(...)`. The raw digest is 32 bytes, so every level of nesting doubles the SHA256 input size. For a 100-element integer list the outer hasher receives 100 × 64 = 6 400 bytes of hex ASCII instead of 100 × 32 = 3 200 bytes of raw digest. The code already documents the coordinated fix at `digest.py:162–169`.

`List (integers, len>100)`: **27 ms** this run (HEAD: 29 ms — noise-level change, structural issue unchanged).

- **Cheap fix**: change `return m.hexdigest().encode()` → `return m.digest()` per the existing comment roadmap (`digest.py:162–169`), update `digest()` to call `.hex()` instead of `.decode()`, and convert the three early-return encode paths to `bytes.fromhex(...)`. Requires a `hash_version` bump + `Cache.redigest` migration.
- **Structural change**: adopt BLAKE3 tree-hashing so composable sub-hashes are produced natively at 32-byte width with no hex round-trip.

### 2. `_digest_bytes` allocates a fresh `sha256()` context on every recursive call — `digest.py:172`

```python
m = hashlib.sha256()
```

`hashlib.sha256()` fires at the top of every `_digest_bytes` invocation — once per element of every list, dict key, or numpy shape tuple. CPython's `sha256` constructor costs ≈200 ns; for a 100-element integer list that is ≈20 µs of pure context-allocation overhead before any bytes are hashed. `List (integers, len>100)` at 27 ms vs ~1.1 ms for `len<100` (~25× slower for ~3× more elements) is consistent with O(N) SHA256 object allocation dominating the cost.

- **Cheap fix**: pass the parent hasher `m` into recursive calls and call `m.update(...)` directly for container types (iterables, dicts, tuples), avoiding the child SHA256 object entirely. No wire-format change is required for these paths.
- **Structural change**: adopt a Merkle scheme where leaves are hashed independently (memoizable) and parent nodes combine children with a keyed concat step rather than a new SHA256 context.

### 3. BagOfHoldingH5File opens and closes the HDF5 file on every operation — `bagofholding_file.py:36,45`

`_to_file` calls `H5Bag.save(value, path)` (line 36) and `_from_file` calls `H5Bag(path).load()` (line 45) — each opens and closes the HDF5 file independently. BagOfHoldingH5File is consistently **7–15× slower** than PickleFile across all workloads:

| Workload | BagOfHolding load | PickleFile load | Ratio |
|---|---|---|---|
| small_strings | 1.8 ms | 160 µs | **11×** |
| nested_structures | 1.3 ms | 260 µs | **5×** |
| numpy_arrays | 1.9 ms | 180 µs | **11×** |

Not a regression, but consistently the slowest value backend by a wide margin. The hot path pays HDF5's per-file open/flush/close cost on every single `put`/`get`.

- **Cheap fix**: investigate whether h5py SWMR mode or flush-deferral options can be threaded through the `H5Bag` call site to reduce open/close overhead.
- **Structural change**: pool open file handles in a thread-local inside `BagOfHoldingH5FileBackend` to amortise open/close cost across consecutive `put`/`get` calls on the same bag.

---

## Skipped / crashed benchmarks

None — all three benchmark scripts (`benchmark_digest.py`, `benchmark_storage.py`, `benchmark_integration.py`) completed without error after installing test dependencies (`pip install -e ".[tests]"`).

---

_perf: `perf record` was not available in this environment; CPU profiling done via source-code tracing only._

pmrv:
## Summary

- `H5Bag.__init__` eagerly calls `_unpack_bag_info()` when the file already exists, opening and closing the HDF5 file to read bag metadata. The subsequent `H5Bag.load()` then opens it a second time to read the actual payload — two `openat` syscalls per load where one would do.
- Fix: pass `_skip_load=True` in `_from_file` and `rebag`, available since bagofholding 0.1.12, which skips the constructor's metadata read. The load call itself validates the bag info, so no correctness is lost.

Closes #527 (hot-spot #3).

## Benchmark (1000-element float64 numpy array, 200 reps)

| | µs/load |
|---|---|
| `H5Bag(_skip_load=False)` (old) | 826 |
| `H5Bag(_skip_load=True)` (new) | 531 |
| `ValueBagOfHoldingH5File` via storage (branch) | 755 |
| `PickleFile` reference | 105 |

The fix gives ~1.56× speedup on the raw H5Bag call and narrows the gap to PickleFile from 7.9× to 5.1×. The remaining gap is intrinsic HDF5 format overhead, unaffected by this change.

## Test plan

- [ ] Existing storage tests pass
- [ ] Strace confirms single `O_RDONLY` per load (see PR comment)

### DIFF
--- a/src/fleche/storage/bagofholding_file.py
+++ b/src/fleche/storage/bagofholding_file.py
@@ -39,9 +39,13 @@ class BagOfHoldingH5FileBackend(FileStorage):
 
     def _from_file(self, path: Path) -> Any:
         try:
+            # _skip_load=True skips the constructor's _load_existing_bag_info() call,
+            # which would otherwise open and close the file just to read bag metadata
+            # before load() opens it a second time to read the actual payload.
+            bag = H5Bag(path, _skip_load=True)
             if self.version_validator is not None:
-                return H5Bag(path).load(version_validator=self.version_validator)
-            return H5Bag(path).load()
+                return bag.load(version_validator=self.version_validator)
+            return bag.load()
         except FileNotFoundError:
             raise KeyError(path) from None
         except OSError as e:
@@ -58,7 +62,7 @@ class BagOfHoldingH5FileBackend(FileStorage):
             path = self._path(key)
             with self._operation_context(key):
                 try:
-                    value = H5Bag(path).load(version_validator=version_validator)
+                    value = H5Bag(path, _skip_load=True).load(version_validator=version_validator)
                     H5Bag.save(value, path)
                 except OSError as e:
                     logger.warning("Failed to rebag %s: %s", key, e)
--- a/tests/unit/storage/test_bagofholding_file.py
+++ b/tests/unit/storage/test_bagofholding_file.py
@@ -106,6 +106,42 @@ def test_rebag_skips_oserror(tmp_path, monkeypatch, caplog):
     assert "Failed to rebag" in caplog.text
 
 
+def test_from_file_passes_skip_load_to_h5bag(tmp_path, monkeypatch):
+    """_from_file must pass _skip_load=True so the file is opened only once per get."""
+    pytest.importorskip("bagofholding")
+    import fleche.storage.bagofholding_file as boh_mod
+
+    mock_h5bag = MagicMock()
+    mock_h5bag.return_value.load.return_value = 7
+    monkeypatch.setattr(boh_mod, "H5Bag", mock_h5bag)
+
+    s = BagOfHoldingH5FileBackend(tmp_path)
+    key = Digest("skip_load_key")
+    s.get(key)
+
+    _, kwargs = mock_h5bag.call_args
+    assert kwargs.get("_skip_load") is True
+
+
+def test_rebag_passes_skip_load_to_h5bag(tmp_path, monkeypatch):
+    """rebag must pass _skip_load=True so the file is opened only once per entry."""
+    pytest.importorskip("bagofholding")
+    import fleche.storage.bagofholding_file as boh_mod
+
+    mock_h5bag = MagicMock()
+    mock_h5bag.return_value.load.return_value = 1
+    monkeypatch.setattr(boh_mod, "H5Bag", mock_h5bag)
+
+    s = BagOfHoldingH5FileBackend(tmp_path)
+    key = Digest("rebag_skip_load_key")
+    s._path(key).write_bytes(b"dummy")
+
+    s.rebag()
+
+    _, kwargs = mock_h5bag.call_args
+    assert kwargs.get("_skip_load") is True
+
+
 def test_rebag_default_validator_is_none(tmp_path, monkeypatch):
     pytest.importorskip("bagofholding")
     import fleche.storage.bagofholding_file as boh_mod


pmrv:
## strace verification

Ran `strace -e trace=openat` on the branch against a single save + load of a 1000-element `np.float64` array (one file, no destructuring). Trimmed to the payload file only (lock file omitted).

**Save phase** — two opens as expected: first an `O_RDWR` existence probe (returns `ENOENT`), then `O_RDWR|O_CREAT|O_TRUNC` to write:

```
=== SAVE ===
openat(".../af3a35fc...", O_RDWR) = -1 ENOENT
openat(".../af3a35fc...", O_RDWR|O_CREAT|O_TRUNC, 0666) = 5
```

**Load phase (branch, `_skip_load=True`)** — exactly one `O_RDONLY`:

```
=== LOAD ===
openat(".../af3a35fc...", O_RDONLY) = 5
=== DONE ===
```

For comparison, before this fix (`_skip_load=False` / `H5Bag(path)`) the same load produced **two** `O_RDONLY` opens:

```
=== LOAD ===
openat(".../af3a35fc...", O_RDONLY) = 5   ← Bag.__init__ reads bag_info, then closes
openat(".../af3a35fc...", O_RDONLY) = 5   ← H5Bag.load() opens again for the payload
```

The extra open came from `Bag.__init__` (bagofholding `bag.py:142`) calling `self._unpack_bag_info()` whenever the file already exists, which `H5Bag._unpack_bag_info` wraps in a `with self:` context — open, read attrs, close — before `H5Bag.load()` immediately opens it again. `_skip_load=True` bypasses that constructor path entirely.

Issue #614, opened 2026-06-02
Blake Hash Recon


pmrv:
[Blake2](https://www.blake2.net/blake2_20130129.pdf) is an alternative hash also supported in stdlib python.  It has two interesting features:

1. a tree mode, conserving exactly the merkle property we hand rolled into `digest`
2. claims to be ~3 times faster than SHA

pmrv:
@claude Investigate the tree mode as it is exposed in the stdlib, particularly wrt #527 that we reinit hash contexts a lot for nested objects.

pmrv:
@claude  Do a performance comparison of you current digest and one that uses blake2b with equal digest size.  I want plots of hash cost as a function of input size for strings, lists and numpy arrays.  Build also simple tree by nested 2-tuples with given depth, this depths should be the x axis in another plot.

pmrv:
@claude look at https://www.ietf.org/archive/id/draft-aumasson-blake3-00.html and https://pypi.org/project/blake3/

pmrv:
@claude  Install the blake3 package in your env after fleche and repeat the benchmarks with the scripts from above.  Modify the script slightly so that the size of strings, lists and arrays is all measured in total bytes.  I want comparisons between all three hashing functions.

pmrv:
@claude verify for me quickly that string and array cases are exactly the same.  your plot uses different xylimits making this difficult to tell.  Both should be the same 'bulk bytes hashing' limit of application.

pmrv:
@claude seems blake2b is a good candidate, draft a migration plan.

pmrv:
@claude open a new issue for option A only.

pmrv:
@claude do we have a good test coverage for redigest?

pmrv:
@claude break the gaps into a new issue.  Start working on gap 1 there, i want tests for both 'only one type changes its digest impl' and 'the entire hash function changes'.

PR #619, opened 2026-06-02
test: add gap-1 redigest coverage for single-type and full hash changes

pmrv:
Two new tests in `tests/unit/caches/test_redigest.py` covering gap 1 from #618 (the actual hash-change migration scenario).

After a hash-function change, `redigest()` re-keys calls but leaves old value entries as orphans; a subsequent `gc()` removes them.

- Single-type change: patches `_digest_bytes` to flip int hashes only
- Full primitive swap: replaces `digest.hashlib.sha256` with blake2b(digest_size=32)

Closes gap 1 from #618. Relates to #614.

### DIFF
--- a/tests/unit/caches/test_redigest.py
+++ b/tests/unit/caches/test_redigest.py
@@ -3,6 +3,7 @@ from fleche.storage import ValueMemory, CallMemory
 from fleche.caches import Cache
 from fleche.call import Call
 from fleche.digest import Digest
+import fleche.digest as _fd
 
 
 def _flip_last_nibble(hexstr: str) -> str:
@@ -145,18 +146,8 @@ def test_redigest_noop_if_digest_unchanged(monkeypatch):
 
     import fleche.digest as fd
 
-    def passthrough(v):
-        return fd.digest.__wrapped__(v) if hasattr(fd.digest, "__wrapped__") else fd._digest(v)  # type: ignore[attr-defined]
-
-    # If the original digest isn't wrapped, just reuse it directly
-    # to represent "no change" semantics
-    try:
-        orig = fd.digest
-    except Exception:  # pragma: no cover - defensive
-        orig = None
-
-    if orig is not None:
-        monkeypatch.setattr(fd, "digest", orig, raising=True)
+    # Re-patch with the same function: digest is unchanged, redigest must be a no-op.
+    monkeypatch.setattr(fd, "digest", fd.digest, raising=True)
 
     cache.redigest()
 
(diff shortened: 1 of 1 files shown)


tests/unit/caches/test_redigest.py
```diff
@@ -145,18 +146,8 @@ def test_redigest_noop_if_digest_unchanged(monkeypatch):
...
-    # to represent "no change" semantics
-    try:
-        orig = fd.digest
-    except Exception:  # pragma: no cover - defensive
-        orig = None
-
-    if orig is not None:
-        monkeypatch.setattr(fd, "digest", orig, raising=True)
+    # Re-patch with the same function: digest is unchanged, redigest must be a no-op.
+    monkeypatch.setattr(fd, "digest", fd.digest, raising=True)
```
pmrv: Is there any purpose in monkey patching here?  Just call redigest without doing anything else?

tests/unit/caches/test_redigest.py
```diff
@@ -170,3 +161,116 @@ def passthrough(v):
...
+    # New value entries were added under the new int-derived keys.
+    assert values_after > values_before
+
+    # Loading by the new key round-trips correctly.
+    loaded = cache.load(new_key).fetch()
+    assert loaded.arguments["a"] == [1, 2, (3, 4)]
+    assert loaded.arguments["b"] == {"k": 10}
+    assert loaded.result == ("x", {"y": 5})
+
+    # gc() removes the orphaned old entries.
```
pmrv: gc tests are not required, imo, since they are tested separately.  remove or convince me otherwise

tests/unit/caches/test_redigest.py
```diff
@@ -170,3 +161,116 @@ def passthrough(v):
...
+    loaded = cache.load(new_key).fetch()
+    assert loaded.arguments["a"] == [1, 2, (3, 4)]
+    assert loaded.arguments["b"] == {"k": 10}
+    assert loaded.result == ("x", {"y": 5})
+
+    # gc() removes every orphaned old entry — the entire pre-migration value set.
+    evicted = cache.gc()
+    assert evicted == values_before
+    values_final = set(cache.values.list())
+    assert not (evicted & values_final)
```
pmrv: same as above.

pmrv (commented):
@claude revise

tests/unit/caches/test_redigest.py
```diff
@@ -135,38 +136,121 @@ def test_redigest_updates_call_keys_on_call_hash_change(monkeypatch):
...
 
-    if orig is not None:
-        monkeypatch.setattr(fd, "digest", orig, raising=True)
+    Simulates a scenario like "int hashing gained a version prefix". Calls containing
+    int arguments get new lookup keys; redigest re-saves them under those new keys and
+    stores new value entries. The old value entries (saved before the change) are left
+    behind as orphans — only a subsequent gc() removes them.
+    """
+    cache = _make_cache()
+    original = _sample_call()
```
pmrv: why are these not fixtures?

pmrv (commented):
@claude revise

tests/unit/caches/test_redigest.py
```diff
@@ -80,14 +81,13 @@ def patched(value):
     return patched
 
 
-def _make_cache():
-    values = ValueMemory({})
-    calls = CallMemory({})
-    return Cache(values, calls)
+@pytest.fixture
+def cache():
+    return Cache(ValueMemory({}), CallMemory({}))
```
pmrv: @claude for sure there is a global fixture in the support modules.

PR #622, opened 2026-06-03
feat(cache): BaseCache(OperationContext) + PerKeyLockMixin on Cache (steps 4-5 of #569)

pmrv:
Implements steps 4 and 5 of #569.

**Step 4**: `BaseCache` now inherits from `OperationContext`, giving the cache layer the same `_operation_context` hook that the storage layer already has. Per-key methods on `Cache` (`save`, `load`, `load_value`, `evict`, `contains`, `expand`) each enter the hook around their work; `save` derives the key via `call.to_lookup_key()` before entering, mirroring `CallMixin.save`.

**Step 5**: `PerKeyLockMixin` is applied directly to `Cache` (`Cache(PerKeyLockMixin, BaseCache)` --- no separate `PerKeyLockCache` class. `Cache` adopts `__hash__ = object.__hash__` so that storage backends with unhashable dict fields (e.g. `ValueMemory`) do not break the `WeakKeyDictionary` in `PerKeyLockMixin`.

Whole-cache ops (`gc`, `redigest`, `_query`) are intentionally left unwired; they will get `intent="scan"` once that `Intent` value lands.

Closes steps 4-5 of #569.

Generated with [Claude Code](https://claude.ai/code))

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -264,9 +264,6 @@ def _combine_shrink(key: "Digest | str", results: "Iterable[Digest]") -> "Digest
 class Cache(PerKeyLockMixin, BaseCache):
     values: storage.ValueStorage
     calls: storage.CallStorage
-    # Identity hash: values/calls storages may carry unhashable dict fields
-    # (e.g. MemoryBackend.storage), mirroring ValueMemory/CallMemory convention.
-    __hash__ = object.__hash__
 
     def load_value(self, key):
         with self._operation_context(key):
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -10,8 +10,9 @@ import pandas as pd
 from . import digest as _digest
 from .digest import Digest  # type hint convenience
 from . import storage
-from .storage.base import _longest_common_prefix_length
+from .storage.base import _longest_common_prefix_length, OperationContext
 from .storage.destructuring import HasChildDigests
+from .storage.thread_safe import PerKeyLockMixin
 from .call import Call, DigestedCall, LazyCall, QueryCall
 from . import call
 from . import query
@@ -30,7 +31,7 @@ DigestedIterable = storage.destructuring.DigestedIterable
 DigestedDict = storage.destructuring.DigestedDict
 
 
-class BaseCache(ABC):
+class BaseCache(OperationContext):
 
     @abstractmethod
     def save(self, call: Call) -> str:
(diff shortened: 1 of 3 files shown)


src/fleche/caches.py
```diff
@@ -260,34 +261,43 @@ def _combine_shrink(key: "Digest | str", results: "Iterable[Digest]") -> "Digest
 
 
 @dataclass(frozen=True)
-class Cache(BaseCache):
+class Cache(PerKeyLockMixin, BaseCache):
     values: storage.ValueStorage
     calls: storage.CallStorage
+    # Identity hash: values/calls storages may carry unhashable dict fields
+    # (e.g. MemoryBackend.storage), mirroring ValueMemory/CallMemory convention.
+    __hash__ = object.__hash__
```
pmrv: Isn't unsafe_hash the canonical choice?

tests/unit/caches/test_operation_context_cache.py
```diff
@@ -0,0 +1,215 @@
...
+from fleche.storage.memory import ValueMemory, CallMemory
+from fleche.storage.thread_safe import PerKeyLockMixin
+
+
+def make_call(name: str, x: int, result: int) -> Call:
+    return Call(name=name, arguments={"x": x}, result=result, module="test", version=1, metadata={})
+
+
+def make_cache() -> Cache:
+    return Cache(values=ValueMemory({}), calls=CallMemory({}))
```
pmrv: use global fixtures, they exists for these exact cases.
 
on a side note find out, why I need to remind you all the time of this when it should already be in your memory/agent.md instructions

tests/unit/caches/test_operation_context_cache.py
```diff
@@ -0,0 +1,215 @@
...
+def test_cache_instance_is_perkey_lock_mixin():
+    c = make_cache()
+    assert isinstance(c, PerKeyLockMixin)
+
+
+# ---------------------------------------------------------------------------
+# Per-key methods enter _operation_context
+# ---------------------------------------------------------------------------
+
+def _make_tracking_cache():
```
pmrv: use fixtures

tests/unit/caches/test_operation_context_cache.py
```diff
@@ -0,0 +1,215 @@
...
+
+    t1 = threading.Thread(target=load_key, args=(key1, "t1"))
+    t2 = threading.Thread(target=load_key, args=(key2, "t2"))
+    t1.start()
+    t2.start()
+    t1.join(timeout=5)
+    t2.join(timeout=5)
+
+    assert results.get("t1") == 10
+    assert results.get("t2") == 20
```
pmrv: not sure what this tests, but definitely not that the two threads are not blocking each other.

pmrv (commented):
@claude revise

tests/unit/caches/test_gc.py
```diff
@@ -19,6 +19,10 @@ class FlatValueMemory(ValueMixin, MemoryBackend):
     :class:`fleche.storage.destructuring.HasChildDigests`.  Used to exercise
     the GC path that walks only direct call references.
     """
+    # @dataclass(frozen=True) regenerates __hash__ from fields, including the
+    # inherited storage: dict, which is unhashable. Mirror MemoryBackend's
+    # convention to keep identity-based hashing.
+    __hash__ = object.__hash__
```
pmrv: would it make sense to ad MemoryBackend?  If it's already there, why aren't we inheriting it here?

pmrv (commented):
@claude revise

PR #627, opened 2026-06-04
fix(wrapper): narrow _in_flight to the cache-write gap, not full computation

### LINKED ISSUE #485: fix: narrow the _in_flight window to cover only the cache-write gap, not the full computation
## Background

CPython's `Future.set_result()` releases the condition lock **before** calling `_invoke_callbacks()`. This means a thread blocked on `future.result()` can be unblocked and proceed while the done-callback that saves to the cache has not yet run.

PR #478 fixed the symptom by tracking every in-flight `Future` in a per-wrapper `_in_flight: dict[str, Future]` dict. A second call with the same key finds the future in `_in_flight`, waits on its result, then re-checks the cache (or returns the value directly if the callback lost the race).

## The race window (ASCII art)

```
Main Thread                         Worker Thread
──────────────────────────────────  ──────────────────────────────────
compute(5)
  cache miss
  f = func(5)          ────────────►  (expensive computation runs…)
  _in_flight[key] = f               Future.set_result(25):
  f.add_done_callback(_cache)         lock.acquire()
  return f                            _state = FINISHED
                                      condition.notify_all()
f.result() ◄─ BLOCKS                  lock.release()     ← unblocks f.result()
f.result() RETURNS 25  ◄────────────
                                    ╔══════════════════════════════╗
compute(5) called again             ║  GAP: cache still empty!     ║
  cache miss                        ╚══════════════════════════════╝
  check _in_flight → found            _invoke_callbacks() runs NOW
  wait on future (already done)       _cache_and_cleanup()
  re-check cache or return val        cache.save(call)   ← too late
```

The **gap** is between `lock.release()` inside `set_result()` and the completion of `_invoke_callbacks()` (which runs `_cache_and_cleanup`). It is narrow but real.

## Problem with the current PR #478 design

In PR #478 the future is added to `_in_flight` **immediately after `func()` returns it**, which means fleche tracks the future for the **entire duration of the background computation** — potentially minutes.

This is undesirable because:

1. **Fleche is not an executor.** Tracking long-running futures is the executor's responsibility. Fleche's only responsibility is saving the result to the cache once the future resolves.
2. **Concurrent callers are explicitly out of scope.** The stampede problem (two threads calling `compute(5)` at the same time, both finding a cache miss before either can add the future to `_in_flight`) is already documented as a known limitation. The `_in_flight` dict in its current form gives the appearance of solving stampedes while actually only helping with the much narrower sequential re-call case.
3. **Exposure to executor state.** Holding a reference to an in-flight `Future` for the entire computation couples fleche to the executor's lifecycle in subtle ways (cancellation, exception propagation, `BoundWrapper` pickling concerns, etc.).

## Proposed design

The future should be entered into `_in_flight` **only inside the done callback** — i.e., only for the duration of the cache-write window — not while the computation is still running.

Conceptual sketch:

```python
def _cache_and_cleanup(f):
    with _in_flight_lock:
        _in_flight[key] = f          # enter only now — result is in our hands
    try:
        _cache(f)
    finally:
        with _in_flight_lock:
            _in_flight.pop(key, None)  # exit after save completes

result.add_done_callback(_cache_and_cleanup)
return result
# _in_flight is NOT populated before this point
```

A second caller that hits the window checks `_in_flight` after a cache miss: if the key is present, the future is already resolved (done callbacks only fire on resolved futures) so `.result()` returns immediately, and the caller re-checks the cache or uses the value directly.

### Trade-offs

| | Current PR #478 | Proposed narrower design |
|---|---|---|
| Covers narrow sequential re-call window | ✅ partially* | ✅ (when callback has started) |
| Covers concurrent stampede | ❌ not reliably | ❌ intentionally out of scope |
| Tracks futures during computation | ✅ (undesirable) | ❌ |
| Fleche exposure to executor lifecycle | high | minimal |

\* The current design also has a window: between `func()` returning the future and `_in_flight[key] = result` being set under the lock. In practice this window is tiny, but it exists.

### Sequential re-call fix

For the common case of a single thread calling `compute(5)`, then awaiting the result, then calling `compute(5)` again too quickly, the correct fix is in the **test** (not the wrapper): wait for `fn.contains(arg)` to become `True` after `future.result()` returns before re-invoking. PR #484 implements this test fix.

## Implementation notes

- `_in_flight` keyed by the same cache key string used for `cache.load`/`cache.save`.
- The check in the wrapper (on cache miss) remains the same: look up `_in_flight`, call `.result()` (returns immediately since callback only fires on resolved futures), then try `cache.load(key)` again, falling back to the direct value.
- If `_in_flight` is not populated during computation, the stampede behaviour is unchanged from the pre-#478 baseline (two concurrent callers both see a miss and both execute). This is consistent with the documented limitation.
- Add a regression test that explicitly exercises the gap: use `threading.Barrier` to pause the done callback mid-save, confirm a concurrent `compute(5)` call blocks on `_in_flight`, then release the barrier and verify the cache is populated exactly once.

## Related

- Closes / supersedes #477
- PR #478 (current implementation — too broad)
- PR #484 (test-side fix for the sequential re-call case)

pmrv:
Closes #485

## Summary

- `_in_flight[key]` is now populated inside `_cache()` (the done-callback), before `cache.save()`, not while the computation is still running
- `_in_flight` is removed in a `finally` clause after `save()` completes
- Lock management lives entirely in `_cache()`; wrapper body does a lockless `dict.get` read
- Adds `tests/regression/test_issue_485.py` with a `threading.Barrier`-based regression test

### DIFF
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -214,13 +214,10 @@ def make_wrapper(func, policy, meta, isolate, get_call):
 
         # A done-callback for the same key may be mid-save right now (the
         # CPython gap between condition.notify_all() and _invoke_callbacks()).
-        # If so, the future is already resolved — .result() returns immediately.
+        # If so, the future is already resolved — .result() returns immediately
+        # and the future's result is exactly the function result.
         if (f := _in_flight.get(key)) is not None:
-            f.result()
-            try:
-                return cache.load(key).result
-            except KeyError:
-                return f.result()
+            return f.result()
 
         def _run_and_cache():
             active_meta = state._METADATA.get() + tuple(meta)
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -1,4 +1,5 @@
 import os
+import threading
 from pathlib import Path
 from functools import wraps, partial
 from types import SimpleNamespace
@@ -169,6 +170,13 @@ def make_rerun(func, wrapper):
 
 def make_wrapper(func, policy, meta, isolate, get_call):
     """Build the cached wrapper returned by :func:`fleche`."""
+    # Tracks futures whose done-callback is currently executing the cache write.
+    # Populated at the start of _cache() and removed in its finally clause, so
+    # the entry only exists during the narrow window between Future.set_result()
+    # releasing the condition lock and cache.save() completing.
+    _in_flight: dict[str, Future] = {}
+    _in_flight_lock = threading.Lock()
+
     @wraps(func)
     def wrapper(*args: Any, **kwargs: Any) -> _T:
         cache: BaseCache = state._CACHE.get()
@@ -204,6 +212,16 @@ def make_wrapper(func, policy, meta, isolate, get_call):
         except KeyError:
             logger.debug("Cache miss for %s with key %s", call.name, key)
 
+        # A done-callback for the same key may be mid-save right now (the
+        # CPython gap between condition.notify_all() and _invoke_callbacks()).
+        # If so, the future is already resolved — .result() returns immediately.
+        if (f := _in_flight.get(key)) is not None:
+            f.result()
+            try:
+                return cache.load(key).result
+            except KeyError:
+                return f.result()
+
         def _run_and_cache():
             active_meta = state._METADATA.get() + tuple(meta)
             metadata: Dict[str, Any] = defaultdict(dict)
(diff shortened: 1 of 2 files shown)


src/fleche/wrapper.py
```diff
@@ -204,6 +212,16 @@ def wrapper(*args: Any, **kwargs: Any) -> _T:
...
 
+        # A done-callback for the same key may be mid-save right now (the
+        # CPython gap between condition.notify_all() and _invoke_callbacks()).
+        # If so, the future is already resolved — .result() returns immediately.
+        if (f := _in_flight.get(key)) is not None:
+            f.result()
+            try:
+                return cache.load(key).result
+            except KeyError:
+                return f.result()
```
pmrv: the result of the future is the function result, so just `return f.result()` should be enough, no?

tests/regression/test_issue_485.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    _in_flight[key] is set before cache.save so a second compute() call that
+    arrives while the save is in progress gets the value without re-executing
+    the function.
+
+    Threading.Barrier is used to hold the done-callback at the start of
+    cache.save.  While it is paused, a second compute() call is made; it
+    should find the key in _in_flight and return without incrementing
+    call_count.  Releasing the barrier lets save complete, and the cache
+    should contain exactly one entry.
```
pmrv: the cache will only hold one key either way, better to check call metadata.

pmrv: i see you don't actually check the cache count.  that's good, but update the description of the test.

pmrv (commented):
@claude revise

PR #629, opened 2026-06-04
fix(caches): serialize CacheStack.load() auto-transfer to base cache

### LINKED ISSUE #217: `CacheStack.load()` auto-transfer to base cache is not serialized
**File:** `src/fleche/caches.py:493-497`

When a key is found in a higher-level cache but not the base, `CacheStack.load()` automatically transfers it to the base. This is unprotected: multiple threads loading the same key will all miss the base, all hit the higher cache, and all simultaneously call `self.save(call.fetch())` on the base — triggering the non-atomic check-evict-save multiple times concurrently.

**MWE:**
```python
import concurrent.futures, contextvars
import fleche
from fleche.caches import Cache, CacheStack

@fleche.fleche
def compute(x):
    return x * 2

mem_top_v, mem_top_c = Memory({}), Memory({})
mem_base_v, mem_base_c = Memory({}), Memory({})
cache_top = Cache(mem_top_v, mem_top_c)
cache_base = Cache(mem_base_v, mem_base_c)
stack = CacheStack((cache_base, cache_top))

with fleche.cache(cache_top):
    compute(42)  # only in top cache

with fleche.cache(stack):
        futures = [pool.submit(ctx.run, compute, 42) for _ in range(8)]
        [f.result() for f in futures]
# All 8 threads simultaneously attempt to transfer from top to base

### PR DESCRIPTION
**Pull request by pmrv**
Serializes the per-key back-fill in `CacheStack.load()` so concurrent loads of the same missing key transfer the hit into the base cache exactly once, instead of all racing into the base's non-atomic check-evict-save.

Closes #217

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -557,9 +557,11 @@ class CacheStack(PerKeyLockMixin, BaseCache):
         """Transfer a hit from a higher cache into the base cache.
 
         Serialized per key so that concurrent loads of the same missing key do
-        not all run the base cache's non-atomic check-evict-save at once.  The
-        double-checked :meth:`~BaseCache.contains` lets every waiter past the
-        first skip the redundant save once the key has been back-filled.
+        not all run the base cache's non-atomic check-evict-save at once.  All
+        concurrent loaders block on the per-key :meth:`_operation_context` lock;
+        the first one past the lock does the transfer, and every later waiter
+        finds the key already present via :meth:`~BaseCache.contains` and returns
+        without repeating the save.
         """
         with self._operation_context(key):
             if self.stack[0].contains(key):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/fleche/caches.py [hunk 1]
```diff
@@ -544,16 +547,29 @@ def load(self, key) -> LazyCall:
...
                 continue
         raise KeyError(key)
 
+    def _backfill(self, key, lc: LazyCall) -> None:
+        """Transfer a hit from a higher cache into the base cache.
+
+        Serialized per key so that concurrent loads of the same missing key do
+        not all run the base cache's non-atomic check-evict-save at once.  The
+        double-checked :meth:`~BaseCache.contains` lets every waiter past the
+        first skip the redundant save once the key has been back-filled.
```
pmrv: @claude what do you mean by double checked?  if I read correctly.  all concurrent loaders will block on the operation_context lock and wait until the first past the lock does the transfering.  I think that's ok behaviour but your prose is not clear about this.

PR #630, opened 2026-06-04
fix(query): make transfer check-then-save atomic via target per-key lock

### LINKED ISSUE #452: Cache race: BaseCache.transfer() contains→save TOCTOU

## Problem

`BaseCache.transfer()` at `caches.py:56–79` checks for an existing key and then saves, with no lock covering the gap:

```python
conflict = not overwrite and other.contains(key)   # CHECK
if not conflict:
    other.save(c)                                   # SAVE (unguarded)
if pop:
    ...
    self.evict(key)                                 # EVICT source
```

Two concurrent `transfer()` calls (or a transfer racing with a direct `save()`) can both pass the `contains` check and both call `save()`.  With `overwrite=False`, the intent is to skip already-present keys, but the window between check and save is wide enough for a race.

If `pop=True`, both callers also evict the source entry — the evict itself is harmless (idempotent), but the duplicate save in the target is silent and may produce an inconsistent state.

## Impact

Unintended duplicate writes with `overwrite=False`; silently violated "no overwrite" semantics.

## Possible fix

Acquire a per-key lock on `other` (or a cache-level lock) before the `contains` check and hold it until after `save()` returns, making the check-then-save atomic.

pmrv:
Fixes #452.

`QueryIterator.transfer` (and `BaseCache.transfer`, which delegates to it) checked `target.contains(key)` then `target.save()` with no lock spanning the gap, so concurrent transfers could both pass the `overwrite=False` check and both write. The fix holds the target's reentrant per-key `_operation_context(key)` across the whole `contains` → `shrink` → `save` sequence, making the check-then-save atomic and also closing the `shrink` `KeyError` window.

Regression-pinned in `tests/regression/test_issue_452.py`.


src/fleche/query.py
```diff
@@ -189,14 +189,24 @@ def transfer(self, target, pop: bool = False, overwrite: bool = False) -> None:
...
+            # same key does not deadlock.  Caches without per-key locking
+            # inherit a no-op context, so this is at worst a no-op for them.
+            with target._operation_context(key):
+                conflict = not overwrite and target.contains(key)
+                if conflict:
+                    logger.warning(
+                        "Not transferring %s: already exists in target and overwrite=False",
+                        target.shrink(key),
+                    )
+                    continue
```
pmrv: @claude annoying to break API here.  Do you see alternatives?

pmrv: @claude Let it be done! make `_transfer_one` happen.

pmrv: @claude I storngly want to avoid a separate transfer_context.  The issue is the that some basecache subclasses, don't provide a locking operations context.  I think for the wrapping cache, the issue is simply solved by just forwarding the underlying context.  We didn't do this on introduction because we thought it wouldn't be necessary and it'd save us an extra lock, but I think that was a mistake.  the underlying and the wrapping cache cannot meaningfully be accessed separately anyway.

More difficult for the stack.  brainstorm some options.

pmrv: @claude Option 2 would give us a good reason to introduce Intent.READ (we called it scan earlier, but read is a bit more direct I think).  Discuss and check the deadlock potential a bit more.  We'd basically always take Intent.READ on all stack members and Intent.WRITE on stack[0] iff that's our given intent.  For a start Intent.READ would be a noop, but reader-writer locks in the future could be an option.

pmrv: @claude check out the cost of implementing rwlocks.

pmrv: @claude Check `fasteners`, it has a reentrant threading reader/writer, but not upgradable.  How far does that get us.

pmrv:
@claude Based on previous discussion add the simplest resolve to this issue, imo that's making Intent.READ no-op everywhere and call it a day.

pmrv:
the .shrink call also needs to be removed from transfer.  internal api calls should enver expand or shrink.

pmrv:
> the .shrink call also needs to be removed from transfer. internal api calls should enver expand or shrink.

@claude see to it.

pmrv:
@claude rebase on main to avoid fix the merge conflict, then squash and merge.

PR #651, opened 2026-06-14
fix(digest): handle type objects without raising Indigestible

### LINKED ISSUE #469: digest() broken for types
In 0.8.2 fl.D(C) where C is a dataclass type breaks, this may have been fixed in some update, but in general basic types throw Unhashable as well, but shouldn't.

pmrv:
Closes #469.

## Problem

`digest(int)`, `digest(str)`, and `digest(SomeDataclass)` (a class object, not an instance) all failed:

- Built-in types raised `Indigestible` — no match arm handled them.
- Dataclass *classes* hit the `dataclasses.is_dataclass` arm and then crashed with `AttributeError` on `getattr(cls, field_name)` for required fields that have no class-level attribute.

`dataclasses.is_dataclass()` returns `True` for both the class and its instances, which was the source of the class/instance confusion.

## Fix

Two changes in `_digest_bytes` (`src/fleche/digest.py`):

1. Insert `case _ if isinstance(value, type) and value.__module__ == 'builtins':` **before** the dataclass arm. Built-in types (`int`, `str`, `list`, …) are digested by their qualified name (`builtins.<qualname>`). This is deliberately **restricted to the `builtins` module** — user-defined classes (including dataclass *classes*) still raise `Indigestible` as before.
2. Add a `and not isinstance(value, type)` guard to the dataclass arm, so it is only reached by dataclass *instances*, never the class object.

## Tests

New tests in `tests/unit/digest/test_digest.py` covering:
- Built-in types are digestible without error
- Built-in type digest is stable (`digest(int) == digest(int)`)
- Different built-in types produce different digests
- Built-in type digest differs from instance digest (`digest(int) != digest(0)`)
- User-defined classes raise `Indigestible`
- Dataclass *classes* raise `Indigestible`
- Dataclass *instances* still digest correctly (type guard doesn't break the instance path)

https://claude.ai/code/session_01UTuB8gbhCrdtR5Q2rvMoSS

---
_Generated by [Claude Code](https://claude.ai/code/session_01UTuB8gbhCrdtR5Q2rvMoSS)_

### DIFF
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -273,6 +273,12 @@ def _digest_bytes(value: Any) -> bytes:
             m.update(_digest_bytes(value.__func__))
         case property():
             m.update(_digest_bytes((value.fget, value.fset, value.fdel)))
+        case _ if isinstance(value, type):
+            # Digest a class (type object) by its fully-qualified name: module + qualname.
+            # This case must precede the dataclasses check because dataclasses.is_dataclass()
+            # returns True for both the class itself and its instances, but
+            # getattr(cls, field_name) raises AttributeError for required fields.
+            m.update(_digest_bytes(f"{getattr(value, '__module__', '')}.{getattr(value, '__qualname__', value.__name__)}"))
         case _ if dataclasses.is_dataclass(value):
             # cannot use asdict because it recursively converts values which destroys digests
             # instead (flat-) convert to dictionaries, salt with type name, then fallback to dictionary case.
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -279,7 +279,7 @@ def _digest_bytes(value: Any) -> bytes:
             # This case must precede the dataclasses check: is_dataclass() returns True
             # for both a class and its instances, but getattr(cls, field) raises
             # AttributeError for required fields that have no class-level default.
-            m.update(_digest_bytes(f"builtins.{value.__qualname__}"))
+            return _digest_bytes(f"builtins.{value.__qualname__}")
         case _ if dataclasses.is_dataclass(value) and not isinstance(value, type):
             # cannot use asdict because it recursively converts values which destroys digests
             # instead (flat-) convert to dictionaries, salt with type name, then fallback to dictionary case.
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -273,13 +273,14 @@ def _digest_bytes(value: Any) -> bytes:
             m.update(_digest_bytes(value.__func__))
         case property():
             m.update(_digest_bytes((value.fget, value.fset, value.fdel)))
-        case _ if isinstance(value, type):
-            # Digest a class (type object) by its fully-qualified name: module + qualname.
-            # This case must precede the dataclasses check because dataclasses.is_dataclass()
-            # returns True for both the class itself and its instances, but
-            # getattr(cls, field_name) raises AttributeError for required fields.
-            m.update(_digest_bytes(f"{getattr(value, '__module__', '')}.{getattr(value, '__qualname__', value.__name__)}"))
-        case _ if dataclasses.is_dataclass(value):
+        case _ if isinstance(value, type) and value.__module__ == 'builtins':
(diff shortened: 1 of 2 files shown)


pmrv (commented):
This should only be applied to builtin classes from stdlib.  Everyone else should fail as before.  Do add a `and not isinstance(value, type)` guard to the dataclass value arm.

pmrv:
@claude update OP

src/fleche/digest.py
```diff
@@ -273,7 +273,14 @@ def _digest_bytes(value: Any) -> bytes:
...
         case property():
             m.update(_digest_bytes((value.fget, value.fset, value.fdel)))
-        case _ if dataclasses.is_dataclass(value):
+        case _ if isinstance(value, type) and value.__module__ == 'builtins':
+            # Digest a built-in type (int, str, list, …) by its qualified name.
+            # Restricted to the builtins module; user-defined types remain Indigestible.
+            # This case must precede the dataclasses check: is_dataclass() returns True
+            # for both a class and its instances, but getattr(cls, field) raises
+            # AttributeError for required fields that have no class-level default.
+            m.update(_digest_bytes(f"builtins.{value.__qualname__}"))
```
pmrv: @claude return directly to avoid double hash

tests/unit/digest/test_digest.py
```diff
@@ -841,3 +841,62 @@ def f2(x):
...
 
     assert digest(m1) == digest(m2)
+
+
+# --- Tests for digesting types (classes) — issue #469 ---
+
+
+def test_builtin_types_are_digestible():
+    """Built-in types (builtins module) must not raise Indigestible."""
+    for t in (int, str, float, bool, bytes, list, dict, tuple, set, type, object):
```
pmrv: pytest parametrize on the outside

tests/unit/digest/test_digest.py
```diff
@@ -841,3 +841,62 @@ def f2(x):
...
+        assert isinstance(digest(t), Digest)
+
+
+def test_builtin_type_digest_is_stable():
```
pmrv: useless, remove

tests/unit/digest/test_digest.py
```diff
@@ -841,3 +841,62 @@ def f2(x):
...
+    """Digesting the same built-in type twice returns the same value."""
+    assert digest(int) == digest(int)
+    assert digest(str) == digest(str)
+
+
+def test_different_builtin_types_have_different_digests():
```
pmrv: do a parametrized version that goes over product of supported values doing a kind of if t1!=t2 -> D(t1) != D(t2) else == -> ==

tests/unit/digest/test_digest.py
```diff
@@ -841,3 +841,62 @@ def f2(x):
...
+    """int, str, and float must produce distinct digests."""
+    assert digest(int) != digest(str)
+    assert digest(int) != digest(float)
+
+
+def test_builtin_type_digest_differs_from_instance_digest():
```
pmrv: probably better to add the types to the generic value fixtures and make that work

pmrv (commented):
@claude revise

tests/unit/digest/test_digest.py
```diff
@@ -841,3 +846,60 @@ def f2(x):
...
+
+@pytest.mark.parametrize("t", BUILTIN_TYPES)
+def test_builtin_type_is_digestible(t):
+    """Built-in types (builtins module) must not raise Indigestible."""
+    assert isinstance(digest(t), Digest)
+
+
+@pytest.mark.parametrize("t2", BUILTIN_TYPES)
+@pytest.mark.parametrize("t1", BUILTIN_TYPES)
+def test_builtin_type_digests_distinct(t1, t2):
```
pmrv: is this not a subtest of the general testbwe have for this?

pmrv (commented):
@Claude revise and compress tests as best as possible without losing coverage

pmrv:
@claude rebase to integrate merge conflict

PR #655, opened 2026-06-15
docs(lazy_call): fix undefined key variable in LazyCall example

pmrv:
## Problem

The introductory code example in `docs/usage/lazy_call.rst` called `cache().load(key)` without ever defining `key`:

```python
from fleche import cache

# Default: returns a LazyCall — cheap, no deserialization yet
lazy_call = cache().load(key)   # NameError: name 'key' is not defined
```

A reader copying this snippet would get a `NameError` immediately. The example also split `.fetch()` into a separate disconnected code block, breaking the narrative flow.

## Fix

Rewrite the example as a single runnable snippet that:
1. Defines a `@fleche`-decorated function
2. Populates the cache with one call
3. Derives `key` via `process.digest()` — the standard way to get a lookup key
4. Calls `cache().load(key)` and shows lazy attribute access
5. Calls `.fetch()` in the same block

```python
from fleche import fleche, cache

@fleche
def process(x):
    return x * 2

with cache("memory"):
    process(21)                       # populate the cache
    key = process.digest(21)          # SHA-256 digest for this call

    lazy_call = cache().load(key)

    print(lazy_call.result)           # 42 — loaded on access
    print(lazy_call.arguments['x'])   # 21 — loaded on access

    call = lazy_call.fetch()
```

Verified: snippet runs without errors and prints expected values.

https://claude.ai/code/session_01KmpuvojoLsfzyLZDAdAxDk

---
_Generated by [Claude Code](https://claude.ai/code/session_01KmpuvojoLsfzyLZDAdAxDk)_

### DIFF
--- a/docs/usage/lazy_call.rst
+++ b/docs/usage/lazy_call.rst
@@ -12,21 +12,25 @@ When you call ``cache().load(key)`` or iterate over ``cache().query(...)``, you 
 
 .. code-block:: python
 
-   from fleche import cache
+   from fleche import fleche, cache
 
-   # Default: returns a LazyCall — cheap, no deserialization yet
-   lazy_call = cache().load(key)
+   @fleche
+   def process(x):
+       return x * 2
 
-   # Arguments and results are fetched only when accessed
-   print(lazy_call.result)          # Triggers a load from value storage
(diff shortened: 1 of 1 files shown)


pmrv:
@claude cache('memory') is default, so you should be able to leave the context manager to get a leaner example.

PR #674, opened 2026-06-22
docs(parallel): correct cache stampede description re: _in_flight mechanism

pmrv:
## Summary

- The "Cache Stampede" section claimed "there is no reservation mechanism between the cache miss check and the cache write in `wrapper.py`" — this is inaccurate.
- An `_in_flight: dict[str, Future]` does exist in `make_wrapper()` and provides a reservation mechanism, but it is narrow in scope:
  - Only active for **Future-based** decorated functions (populated in the `add_done_callback` path)
  - Only covers the brief window between `future.set_result()` and `cache.save()` completing
  - Has no effect across processes
- The real gap is the common stampede scenario: multiple workers calling a **non-Future-returning** function with identical arguments before any result is cached — no protection exists there.
- Updated the section to accurately describe both what `_in_flight` does and where the limitation remains.

**Root code reference:** `src/fleche/wrapper.py`, `make_wrapper()` — the `_in_flight` dict and the `if (f := _in_flight.get(key)) is not None` check.

## Test plan
- [ ] Read `src/fleche/wrapper.py` `make_wrapper()`: confirm `_in_flight[key] = future` is only set when `future is not None` (i.e. inside `add_done_callback`)
- [ ] Confirm the `_in_flight` check fires only for same-process, same-thread duplicate calls during the save window
- [ ] Review updated prose in `docs/parallel_execution.rst` for accuracy

https://claude.ai/code/session_01Jg8uCZRPfHe1NEcj7azztu

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jg8uCZRPfHe1NEcj7azztu)_

### DIFF
--- a/docs/parallel_execution.rst
+++ b/docs/parallel_execution.rst
@@ -406,14 +406,21 @@ computes the function and writes the result — wasting redundant work and
 potentially producing inconsistent call metadata (e.g. ``Runtime`` values will
 differ between the duplicate records).
 
-The root cause is that there is no reservation mechanism between the cache
-miss check and the cache write in ``wrapper.py``.  A correct fix would require
-pre-allocating a "pending" slot in the cache so that other workers can detect
-the in-flight computation and wait, rather than starting their own.
-Implementing this correctly is non-trivial: it requires key-scoped locking,
-timeout handling for stale allocations (e.g. if the computing worker crashes),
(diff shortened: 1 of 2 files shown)


pmrv:
I don't want to promise this, so let's not document it.

PR #675, opened 2026-06-23
fix(digest): hash pandas DataFrame/Series/Index by content

pmrv:
## Summary

Bug report: pandas DataFrames with the same columns but different row values produce identical digests, so cached results are incorrectly reused across distinct inputs.

**Root cause** (`src/fleche/digest.py`): `pd.DataFrame` is not a `Mapping` but *is* an `Iterable`, and iterating a DataFrame yields its column names only — not the data. The `case Iterable():` arm hashed just `["a", "b", ...]`, so any two DataFrames sharing column names collided:

```python
df1 = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})  # 57dbc0...
df2 = pd.DataFrame({'a': [10,20,30], 'b': [40,50,60]})  # 57dbc0... (collision)
```

`pd.Series` and `pd.Index` had related problems (Series iterates values but ignores `name`/`dtype`/`index`).

## Fix

Add explicit pre-`Iterable` arms for `DataFrame`, `Series`, and `Index` that fold in:
- columns / name
- dtype(s)
- index
- `pd.util.hash_pandas_object(value).values.tobytes()` for the values themselves

pandas is already a hard dependency, so the import is unconditional (mirrors the existing `np.ndarray` arm).

No `hash_version` bump / migration shim — per request, this changes pandas digests silently; any pre-existing cache entries keyed off pandas inputs are now unreachable.

## Test plan

- [x] New regression tests in `tests/unit/digest/test_digest.py`:
  - `test_pandas_dataframe_hashes_by_content` — the exact collision from the bug report
  - `test_pandas_dataframe_distinguishes_columns_dtype_index` — column order, dtype, index, renamed columns each produce distinct digests
  - `test_pandas_series_hashes_by_content` — values, equality, `name`
  - `test_pandas_index_hashes_by_content` — values, equality, `name`
- [x] Full `tests/unit/` suite passes (1425 passed)

---
_Generated by [Claude Code](https://claude.ai/code/session_015SYyTiDpq9MADVyeGnmkAW)_

### DIFF
--- a/src/fleche/digest.py
+++ b/src/fleche/digest.py
@@ -242,11 +242,22 @@ def _digest_bytes(value: Any) -> bytes:
         case pd.DataFrame():
             # DataFrame is Iterable (yields column names) but NOT Mapping, so without
             # this arm two DataFrames sharing column names collide regardless of values.
+            #
+            # We digest columns/dtypes/index ourselves and pass index=False to
+            # hash_pandas_object rather than letting index=True fold the index in.
+            # hash_pandas_object only mixes per-element value bytes — it ignores
+            # column names, column dtypes, index.name, and index.dtype — so
+            # index=True alone would still collide DataFrames whose indices differ
+            # only in name or dtype.  Recursing through _digest_bytes on
+            # value.index also reuses the pd.Index arm below, keeping a standalone
+            # Index and a DataFrame's .index digest-consistent.
             m.update(_digest_bytes(list(value.columns)))
             m.update(_digest_bytes([str(d) for d in value.dtypes]))
             m.update(_digest_bytes(value.index))
             m.update(pd.util.hash_pandas_object(value, index=False).values.tobytes())
         case pd.Series():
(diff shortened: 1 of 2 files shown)


src/fleche/digest.py
```diff
@@ -238,6 +239,22 @@ def _digest_bytes(value: Any) -> bytes:
         case np.bool_():
             # np.bool_ ∉ Number so this arm is reachable
             return _digest_bytes(bool(value))
+        case pd.DataFrame():
+            # DataFrame is Iterable (yields column names) but NOT Mapping, so without
+            # this arm two DataFrames sharing column names collide regardless of values.
+            m.update(_digest_bytes(list(value.columns)))
+            m.update(_digest_bytes([str(d) for d in value.dtypes]))
+            m.update(_digest_bytes(value.index))
```
pmrv, line 256: debate in a comment pro/cons of including index by ourselves vs. setting index=True in the hash helper.

PR #689, opened 2026-06-26
feat: add CachePool, a read-only collection of caches

pmrv:
## What

Adds `CachePool` — an **unordered, read-only collection of caches** queried as one. Where `CacheStack` is an *ordered, writable* hierarchy (saves land on `stack[0]`, hits back-fill downward), a `CachePool` never writes to any member: it just exposes several independent caches as a single read source.

Use it to surface a teammate's results directory, a shared immutable archive, or last month's run as one cache you can `load`/`contains`/`query`/`expand`/`shrink` against — without risking a write to any of them.

### Semantics

- `load` / `load_value` — first member to hold the key wins.
- `contains` — true if *any* member holds the key.
- `query` — union across members, deduplicated by lookup key.
- `expand` / `shrink` — combined across members.
- `save` / `evict` — raise `Rejected`.
- Unlike `CacheStack`, `load` **never back-fills**, so members are never mutated as a side effect of reading. Member order only decides which copy a `load` collision returns; every member is an equally valid read source.

## How

- **Extracted `_MultiCache(BaseCache)`** — the read fan-out that `CacheStack` already had (the `_first_hit` / `_collect` / `_foreach` traversal helpers plus `contains`/`load_value`/`expand`/`_shrink`/`_query`) now lives in one shared abstract base. Members are exposed through an abstract `_members` property. `CacheStack(PerKeyLockMixin, _MultiCache)` and `CachePool(_MultiCache)` both subclass it, so the fan-out has a single implementation. `CacheStack` keeps its own write/back-fill/locking behaviour; no behavioural change.
- **Config wiring** — `cache_from_config` builds a `CachePool` from a dict with a `pool` key (a list of member configs, naturally expressed in TOML as a `[[name.pool]]` array-of-tables). `cache_to_config` round-trips it back to `{"pool": [...]}`.
- **Docs** — new section in `docs/storage/configuration.rst`, plus `AGENTS.md` updates.

## Tests

- `tests/unit/caches/test_cache_pool.py` — save/evict rejection, first-hit load with no back-fill, contains/query/expand/shrink fan-out, empty pool, transfer-out, heterogeneous members.
- Config round-trip tests in `test_cache_from_config.py` / `test_cache_to_config.py`.
- Full unit suite (1444 tests) green; `ty check src/` passes. Verified pickling and an end-to-end `@fleche` call served from a pool without recompute.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NAwpJXnXGcD4Sfexx96tRk)_


src/fleche/caches.py
```diff
@@ -777,33 +715,163 @@ def _foreach(
...
+
+    def push(self, cache: BaseCache) -> "CacheStack":
+        return CacheStack((cache, *self.stack))
+
+    def evict(self, key: str | Digest) -> None:
+        self._foreach(lambda c: c.evict(key))
+
+
+@dataclass(frozen=True)
+class CachePool(_MultiCache):
```
pmrv: Does it make sense to pull in the readonly cache here?

pmrv: Good call — done in 831c978. I'd hand-rolled `save`/`evict` raising `Rejected`, which also meant `remote._is_read_only` (it keys on `isinstance(cache, ReadOnlyMixin)`) wouldn't recognise a pool, so an `SshCache` serving one would miss the read-only short-circuit.

The blocker to reusing it directly was that `ReadOnlyMixin` extended `CacheWrapper` (single-`cache` field), which doesn't fit the multi-cache pool. But its `save`/`evict` never touch `self.cache`, so I made `ReadOnlyMixin` a field-free, base-free behaviour mixin and compose it into both layouts:

- `ReadOnlyCache(ReadOnlyMixin, CacheWrapper)`
- `FilteredCache(ReadOnlyMixin, FilteringMixin)`
- `CachePool(ReadOnlyMixin, _MultiCache)`

`ReadOnlyMixin` goes first in the bases so its `save`/`evict` win over the forwarding/aggregating impls (MRO verified). `remote._is_read_only` now recognises a pool for free — pinned by `test_cache_pool_is_recognised_as_read_only`. Full unit suite green (1445) + `ty` clean.

---

PR #694, opened 2026-06-29
docs(custom_digests): fix pycon block with undefined TypeA/TypeB

pmrv:
## Problem

The "Registering Entry Points" section in `docs/dev/custom_digests.rst` contained a `.. code-block:: pycon` block (interactive Python session format with `>>>` prompts) that shows what a user's `my_package/hooks.py` might look like:

```pycon
>>> # my_package/hooks.py
>>> from fleche.digest import digest, Digest, Hook

>>> def type_a_digest(obj: TypeA) -> Digest:
...     return digest((type(obj).__name__, obj.field1, obj.field2))

>>> def type_b_digest(obj: TypeB) -> Digest:
...     return digest((type(obj).__name__, obj.relevant_field))

>>> digest_hooks = [
...     Hook(TypeA, type_a_digest),
...     (TypeB, type_b_digest),
... ]
```

`TypeA` and `TypeB` are never defined. Any attempt to execute this as a doctest (or by copying it into an interpreter) raises:

```
NameError: name 'TypeA' is not defined
```

## Fix

Change the block from `.. code-block:: pycon` to `.. code-block:: python`. The `pycon` directive signals an interactive Python session that can be run as-is; the `python` directive signals a plain source file excerpt — which is the correct framing here, since this snippet represents the contents of `my_package/hooks.py`, not commands to paste at a REPL.

No logic changes; only the RST directive and prompt formatting are affected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WkD6NJ284JhQiwB2NbxGsm)_

### DIFF
--- a/docs/dev/custom_digests.rst
+++ b/docs/dev/custom_digests.rst
@@ -75,21 +75,21 @@ These must be **module-level objects**, not callables that return hooks.
 ``fleche`` loads the entry point with ``importlib.metadata.EntryPoint.load()``,
 which returns the object at the given path directly — it does **not** call it.
 
-.. code-block:: pycon
+.. code-block:: python
 
-   >>> # my_package/hooks.py
-   >>> from fleche.digest import digest, Digest, Hook
+   # my_package/hooks.py
+   from fleche.digest import digest, Digest, Hook
 
-   >>> def type_a_digest(obj: TypeA) -> Digest:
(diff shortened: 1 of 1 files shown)


pmrv:
@claude [haiku] leave as before, but add stub class defs so that it would run interactively.

PR #718, opened 2026-07-06
docs(helpers): fix 'hashable' terminology and document metadata shadowing in .query()

pmrv:
## Summary

Two issues in `docs/usage/helpers.rst` found during cross-reference against source:

### 1. Wrong terminology: "hashable" vs "digestible"

The "Usage with Decorated Methods" note reads:

> For `fleche` to cache calls that include `self`, the class must be hashable — i.e. it must implement a `__digest__` method

"Hashable" is a Python term meaning `__hash__` is defined. `__digest__` is a completely separate fleche-specific protocol. Conflating them is incorrect and misleading — a class can be `__hash__`-able without having `__digest__`, and vice versa.

**Fix**: replaced "hashable" with "digestible" and added a one-line clarification that this is unrelated to `__hash__`.

### 2. Undocumented gotcha: `metadata` parameter shadowing in `.query()`

`.query()` accepts a `metadata=` keyword argument to filter results by metadata tags. If the decorated function itself has a parameter named `metadata`, that parameter **cannot** be passed as a keyword to `.query()` — it is silently replaced by the filter dict.

The source (`wrapper.py`, `make_query()`) already logs a `WARNING` when it detects this, but the situation was entirely undocumented for users. The only workaround — passing the `metadata` argument positionally — was not mentioned anywhere.

```python
# function signature: def fetch(user_id, metadata):
fetch.query(123, "v1")                    # positional — works
fetch.query(user_id=123, metadata="v1")   # shadowed — 'metadata' is the filter dict
```

**Fix**: added a `.. warning::` admonition to the `.query()` description that explains the conflict and shows the positional workaround.

## Verification

- `wrapper.py` `make_query()` confirmed: `logger.warning("Function argument 'metadata' shadowed by query argument")` fires when `"metadata" in call.arguments`.
- `digest.py` and `call.py` confirmed: `__digest__` and Python `__hash__` are entirely orthogonal — `digest()` dispatches on `__digest__`, not `__hash__`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BsP3xV7s9KRMPMwh3gnHHX)_

### DIFF
--- a/docs/usage/helpers.rst
+++ b/docs/usage/helpers.rst
@@ -33,6 +33,19 @@ Returns ``True`` if the result for the given call is already present in the cach
 
 Returns matching cached calls from the active cache. Any argument passed as ``None`` acts as a wildcard, matching any stored value for that parameter. The ``metadata`` keyword argument accepts a dictionary of metadata tags to further filter results (e.g., ``metadata={"tags": {"project": "alpha"}}``).
 
+.. warning::
+
+   If the decorated function has a parameter named ``metadata``, that parameter
+   is shadowed by ``.query()``'s own ``metadata=`` keyword and **cannot** be
+   passed as a keyword argument.  Pass it positionally instead::
+
+      # function signature: def fetch(user_id, metadata):
+      fetch.query(123, "v1")           # positional — works
+      fetch.query(user_id=123, metadata="v1")  # shadowed — 'metadata' is
+                                               # interpreted as the filter dict
+
+   ``fleche`` logs a ``WARNING`` when this situation is detected.
+
 See :doc:`query` for a detailed guide on querying cached calls.
 
 ``.rerun(*args, **kwargs)``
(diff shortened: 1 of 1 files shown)


docs/usage/helpers.rst
```diff
@@ -149,8 +162,10 @@ Usage with Decorated Methods
    However, this helper is a plain function — not a bound method — so ``obj`` is not
    pre-applied; you must pass the instance explicitly as the first positional argument.
 
-   For ``fleche`` to cache calls that include ``self``, the class must be hashable —
-   i.e. it must implement a ``__digest__`` method (see :doc:`/dev/custom_digests`).
+   For ``fleche`` to cache calls that include ``self``, the class must be
+   *digestible* — i.e. it must implement a ``__digest__`` method
+   (see :doc:`/dev/custom_digests`).  This is unrelated to Python's
+   ``__hash__``/hashability.
```
pmrv, line 168:
```suggestion
   For ``fleche`` to cache calls that include ``self``, the class must be
   *digestible* --- i.e. it must implement a ``__digest__`` method
   (see :doc:`/dev/custom_digests`).

```

PR #719, opened 2026-07-06
feat(config): add [default] root tag to stop upward config merge

pmrv:
## What

Adds a `root` boolean tag to the `[default]` section of `fleche.toml`. When set to `true`, the config-discovery walk stops at that file: `fleche.toml` files farther **up** the directory tree — and the `$XDG_CONFIG_HOME` fallback — are no longer merged in. Files **closer** to the CWD still merge on top as usual.

```toml
[default]
cache = "persistent"
root = true              # ignore any fleche.toml farther up the tree
```

## Why

Previously every discovered config on the walk from the CWD up to `$HOME` (plus the XDG fallback) was unconditionally shallow-merged, so a project inherited whatever `fleche.toml` happened to live in a parent directory or `$HOME`. The new tag lets a project pin its configuration and act as the top of the hierarchy — the same convention as ESLint's `root: true`.

## How

- `_load_merged_config()` now collects files closest-first and breaks the walk at the first file whose `[default].root` is truthy, then shallow-merges the retained files as before (closest wins). A small `_is_root_config()` helper does the check defensively (only treats `[default]` as a source when it is a table).
- Since the XDG fallback is appended last on the walk path, stopping at an earlier `root` file naturally excludes it too.

## Docs

- Module docstring in `src/fleche/config.py` (discovery section + new "Stopping the walk").
- `docs/storage/configuration.rst`: discovery note + a `root` subsection under `[default]`.
- `agents/USAGE.md`: config-discovery paragraph.

## Tests

Added four tests in `tests/unit/config/test_config.py`:
- `test_walk_root_stops_upward_merge` — a `root` file blocks a farther file from merging.
- `test_walk_root_keeps_closer_files` — closer files still merge on top of a `root` file.
- `test_walk_root_ignores_xdg_fallback` — `root` also excludes the XDG layer.
- `test_walk_root_false_still_merges` — `root = false` is a no-op.

All 44 config tests pass; `ty check src/` is clean. Verified end-to-end that `root = true` drops the parent's cache section from the merged config while removing it lets the parent merge back in.

https://claude.ai/code/session_012SLcyqXWPSY9vWZbW76qtM

---
_Generated by [Claude Code](https://claude.ai/code/session_012SLcyqXWPSY9vWZbW76qtM)_

### DIFF
--- a/docs/storage/configuration.rst
+++ b/docs/storage/configuration.rst
@@ -16,6 +16,11 @@ merged).  This makes it easy to keep a base config at ``~/fleche.toml``
 (or in XDG) and override individual cache definitions in project
 subdirectories.
 
+A file can opt out of this upward inheritance by setting ``root = true`` in
+its ``[default]`` table (see :ref:`the default section <default-section>`
+below).  The walk stops at the closest such file: files farther up the tree
+— and the ``$XDG_CONFIG_HOME`` fallback — are not merged in.
+
 If no configuration file is discovered, ``fleche`` falls back to a default
 in-memory cache.
 
@@ -78,6 +83,8 @@ without changing anything.
    ...     # The config-file default cache is active inside this block.
    ...     ...
 
+.. _default-section:
+
 The ``[default]`` section
 -------------------------
 
@@ -108,6 +115,27 @@ Example:
    metadata = ["Runtime"]
 
 **Note:** The ``Tags`` metadata cannot be configured from the config file, as it requires arguments.
+
+``root``
+~~~~~~~~
+
+The ``root`` key is a boolean (default ``false``) that marks this file as the
+top of the config hierarchy.  When set to ``true``, the discovery walk stops
+here: any ``fleche.toml`` farther up the directory tree, along with the
+``$XDG_CONFIG_HOME`` fallback, is ignored.  Files *closer* to the current
+directory are still merged on top as usual.
+
+This mirrors the ESLint ``root: true`` convention — use it to pin a
+project's configuration so it does not inherit whatever ``fleche.toml``
+happens to live in a parent directory or ``$HOME``.
+
+Example:
+
+.. code-block:: toml
+
+   [default]
+   cache = "mycache"
+   root = true
 
 Cache sections
 --------------
--- a/docs/storage/configuration.rst
+++ b/docs/storage/configuration.rst
@@ -125,7 +125,7 @@ here: any ``fleche.toml`` farther up the directory tree, along with the
 ``$XDG_CONFIG_HOME`` fallback, is ignored.  Files *closer* to the current
 directory are still merged on top as usual.
 
-This mirrors the ESLint ``root: true`` convention — use it to pin a
+Use it to pin a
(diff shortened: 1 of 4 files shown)


docs/storage/configuration.rst
```diff
@@ -93,6 +100,27 @@ Example:
...
+``root``
+~~~~~~~~
+
+The ``root`` key is a boolean (default ``false``) that marks this file as the
+top of the config hierarchy.  When set to ``true``, the discovery walk stops
+here: any ``fleche.toml`` farther up the directory tree, along with the
+``$XDG_CONFIG_HOME`` fallback, is ignored.  Files *closer* to the current
+directory are still merged on top as usual.
+
+This mirrors the ESLint ``root: true`` convention — use it to pin a
```
pmrv:
```suggestion
Use it to pin a
```

src/fleche/config.py
```diff
@@ -235,16 +249,37 @@ def _collect_config_paths() -> list[Path]:
...
-    for path in reversed(_collect_config_paths()):
+    configs: list[dict[str, Any]] = []
+    for path in _collect_config_paths():  # closest first
         config = _load_config(path)
+        configs.append(config)
+        if _is_root_config(config):
+            break
+
+    merged: dict[str, Any] = {}
+    for config in reversed(configs):  # farthest retained first, closest wins
```
pmrv: is something like

```python
merged: dict[str, Any] = {}
for path in _collect_config_paths():  # closest first
        config = _load_config(path)
        merged.update(config)
        if _is_root_config(config):
            break
```

not shorter?

tests/unit/config/test_config.py
```diff
@@ -741,3 +741,130 @@ def test_walk_merged_metadata(monkeypatch, tmp_path):
...
+    assert isinstance(cache_obj.values, storage.ValueMemory)
+
+
+def test_walk_root_ignores_xdg_fallback(monkeypatch, tmp_path):
+    """`root = true` also excludes the XDG lowest-priority fallback."""
+    home = tmp_path / "home"
+    home.mkdir()
+    sub = home / "project"
+    sub.mkdir()
+    xdg = tmp_path / "xdg"
```
pmrv, line 817: lots of setup code in all cases that obscure test logic, factor out.

src/fleche/config.py
pmrv: Good instinct on collapsing to one loop, but `.update()` closest-first inverts precedence: the closest file gets written first and then the farther files overwrite its keys, so the farthest would win (breaking `test_walk_closer_overrides_farther`). Kept the single closest-first loop + `break` but switched to `setdefault` so the first (closest) writer of each top-level key wins. Done in 3f15892.

---

tests/unit/config/test_config.py
pmrv: Factored out in 3f15892: a `home_and_project` fixture handles the `$HOME`/CWD/XDG setup and a `_write_config` helper writes the dedented TOML, so each test now shows only the configs it exercises.

---

pmrv (commented):
fix merge conflicts by rebase

PR #722, opened 2026-07-06
fix: always bind non-fleche callables in wrap_executor.submit

### LINKED ISSUE #691: How to get cache hits for indirect calls on separate processes?
I've been struggling today to figure out how to make a given cache available in some off-process call (e.g. from a `ProcessPoolExecutor`) when the function being submit is not itself a `@fleche` function, but only references one. E.g., consider some reference file (`scratch2.py` in my example):

```python
import time

from fleche import fleche

@fleche
def slow(t):
    time.sleep(t)
    return t

# @fleche
def carries_slow(t):
    s = slow(t)
    return s
```

If I try to run it twice with pickle storage:

```python
import pathlib
import shutil
import time
from concurrent import futures

from fleche import cache, wrap_executor
from fleche.storage import ValuePickleFile, CallPickleFile

import scratch2

cache_dir = pathlib.Path(".pickle_cache")
shutil.rmtree('.pickle_cache', ignore_errors=True)
pickle_cache = Cache(
    values=ValuePickleFile.with_pickle(root='.pickle_cache/values'),
    calls=CallPickleFile.with_pickle(root='.pickle_cache/calls')
)

print(cache_dir.is_dir())

t = time.time()
with cache(pickle_cache):
    with futures.ProcessPoolExecutor() as exe:
        wrap_executor(exe)
        f = exe.submit(scratch2.carries_slow, 3)
dt_first = time.time() - t
print(f.result())


t = time.time()
with cache(pickle_cache):
    with futures.ProcessPoolExecutor() as exe:
        wrap_executor(exe)
        f = exe.submit(scratch2.carries_slow, 3)
dt_second = time.time() - t
print(f.result())
print("first time", dt_first)
print("second time", dt_second)

```

I get the right answer, but get no cache hit and speedup:

```
False
3
3
first time 3.666440010070801
second time 3.6199047565460205
```

Of course, if I un-comment the decorator on `carries_slow` itself, the second pass is lightning fast.

My best understanding is that this is somehow just a matter of getting the submit call to carry along the cache in a way that it's accessible to the entire call chain on the separate process. I tried playing around with `wrap_executor` and with manually `BoundWrapper.bind`ing the outer function, but couldn't get it to work. I feel like I'm missing something simple here and that this should be doable...

pmrv:
## Summary
- `wrap_executor(...).submit()` now always applies `BoundWrapper.bind` to non-fleche callables before forwarding them to the underlying executor, so a plain function that merely calls a `@fleche`-decorated function still gets the active cache/metadata state propagated into the worker process.
- Added a regression test modeled on the original issue snippet (`slow`/`carries_slow`), using a file-backed cache and `ProcessPoolExecutor(max_tasks_per_child=1)` to confirm the inner fleche call only executes once across two submissions.
- Updated docs/parallel_execution.rst to match the new behavior.

Fixes #691

## Test plan
- [x] `pytest tests/` — 1528 passed, 11 skipped

### DIFF
--- a/src/fleche/executor.py
+++ b/src/fleche/executor.py
@@ -3,13 +3,18 @@
 Motivation: users passing a :func:`fleche.fleche`-decorated function to
 ``executor.submit(...)`` have to remember to call :meth:`.BoundWrapper.bind`
 themselves to carry the active cache/metadata state into the worker.  They also
-pay the submit/serialisation cost even when the result is already cached.
+pay the submit/serialisation cost even when the result is already cached.  The
+same problem affects plain (non-fleche) callables that merely *call* a
+fleche-decorated function somewhere in their body: unless they are bound too,
+the active cache/metadata state never reaches the worker process.
 
 :func:`wrap_executor` patches the ``submit`` method of an executor *instance*
 (we cannot subclass, since callers pass us instances of third-party executors)
 so that:
 
-* non-fleche callables are forwarded unchanged,
+* non-fleche callables are bound via :meth:`.BoundWrapper.bind` and submitted
+  to the original ``submit`` unchanged otherwise, so that any fleche calls
+  nested inside them still see the active cache/metadata state,
 * fleche callables whose result is already cached are returned via an
   already-completed :class:`~concurrent.futures.Future` without touching the
   executor, and
@@ -25,6 +30,8 @@ remaining keyword arguments are bound as part of the function payload.
 from concurrent.futures import Future
 from inspect import signature, Parameter
 from types import SimpleNamespace
+
+from . import state
 
 
 __all__ = ["wrap_executor"]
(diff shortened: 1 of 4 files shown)


tests/integration/test_parallel_execution.py
```diff
@@ -442,3 +443,74 @@ def test_executorlib_wrap_executor(file_cache):
 
     assert result == 42, f"Expected 42, got {result}"
     assert file_cache.contains(double.digest(21))
+
+
+# ---------------------------------------------------------------------------
+# Regression test for #691
+# ---------------------------------------------------------------------------
```
pmrv: in a seperate regression folder

pmrv: @claude ^

src/fleche/executor.py
```diff
@@ -80,7 +87,8 @@ def wrap_executor(executor):
 
     def submit(func, *args, **kwargs):
         if not _is_fleche_function(func):
-            return original_submit(func, *args, **kwargs)
+            bound = state.BoundWrapper.bind(func)
```
pmrv, line 90:
```suggestion
            if not isinstance(func, state.BoundWrapper)
                  bound = state.BoundWrapper.bind(func)
```

pmrv: @claude new PR for this

PR #731, opened 2026-07-08
Logging cleanup: config.py levels, drop dead pickle_file logger

pmrv:
## Summary

First round of the logging cleanup from the audit overview:

- `config.py` `_load_config`: kept the failure log at `error` (per review), but added `exc_info=True` so the traceback is preserved. Formatting already used lazy `%s` args.
- `config.py` `_default_memory_cache`: downgraded the "using default memory cache" fallback log from `warning` to `info` — it's an expected fallback, not a problem.
- `storage/pickle_file.py`: removed the unused `logger = logging.getLogger(...)` (never called anywhere in the file) and the now-unused `import logging`.
- Added `exc_info=True` to the other `error`-level logs that catch a live exception but were dropping the traceback: `digest.py`, `storage/file.py`, `storage/bagofholding_file.py`, `caches.py`, `remote.py` (two call sites). Also switched the two f-string-formatted calls (`bagofholding_file.py`, `caches.py`) to lazy `%s` args for consistency.
- Left `security.py`'s `error` logs untouched — those fire outside any `except` block (they're validation failures before raising), so there's no live exception for `exc_info` to attach.

## Test plan
- [x] `uv run pytest tests/unit -q` — 1458 passed
- [x] `uv run python -c "import fleche.storage.pickle_file"` — imports cleanly

### DIFF
--- a/src/fleche/config.py
+++ b/src/fleche/config.py
@@ -204,7 +204,7 @@ def _load_config(path: Path) -> dict[str, Any]:
         with open(path, "rb") as f:
             return tomllib.load(f)
     except Exception as e:
-        logger.info("Failed to load configuration from %s: %s", path, e, exc_info=True)
+        logger.error("Failed to load configuration from %s: %s", path, e, exc_info=True)
         return {}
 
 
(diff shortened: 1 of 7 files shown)


src/fleche/config.py
```diff
@@ -204,7 +204,7 @@ def _load_config(path: Path) -> dict[str, Any]:
         with open(path, "rb") as f:
             return tomllib.load(f)
     except Exception as e:
-        logger.error("Failed to load configuration from %s: %s", path, e)
+        logger.info("Failed to load configuration from %s: %s", path, e, exc_info=True)
```
pmrv: keep it as error.   There's another warning call here that I want to be info

PR #734, opened 2026-07-09
refactor(storage,caches): fold _combine_expand into _resolve_prefix

### LINKED ISSUE #661: Refactor: consolidate `_combine_expand` (caches.py) and `_resolve_prefix` (storage/base.py)
## Problem

Two helpers implement the same "given candidates for a digest prefix, return the unique full key or raise" logic:

`src/fleche/storage/base.py:38` — `_resolve_prefix(key, candidates)`
```python
if not candidates:
    raise KeyError(key)
if len(candidates) == 1:
    return candidates[0]
lcp = _longest_common_prefix_length(candidates[0], candidates[1])
raise AmbiguousDigestError(f"Short digest {key} is ambiguous; need at least {lcp+1} characters.")
```

`src/fleche/caches.py:233` — `_combine_expand(key, results)`
```python
unique = sorted(set(results))
if not unique:
    raise KeyError(key)
if len(unique) == 1:
    return unique[0]
lcp = _longest_common_prefix_length(unique[0], unique[1])
raise AmbiguousDigestError(f"Short digest {key} is ambiguous; expands to: {unique}; need at least {lcp + 1} characters.")
```

The only meaningful differences are (a) `_combine_expand` dedupes via `sorted(set(...))` because multiple sub-storages can yield the same digest, while `_resolve_prefix` callers (`KeyManagement.expand`, `Sql.expand`) already pre-slice to two candidates; and (b) the ambiguous-error message is slightly richer in `_combine_expand`.

Call sites are `KeyManagement.expand` (storage/base.py:144) + `Sql.expand` (storage/sql.py:312) for `_resolve_prefix`, and `Cache.expand` (caches.py:297) + `CacheStack.expand` (caches.py:600) for `_combine_expand`.

## Proposal

Move the shared core into `storage/base.py` (or a tiny `storage/_prefix.py`) as one helper, e.g.

```python
def resolve_prefix(key, candidates, *, dedupe: bool = False) -> Digest:
    if dedupe:
        candidates = sorted(set(candidates))
    if not candidates:
        raise KeyError(key)
    if len(candidates) == 1:
        return candidates[0]
    lcp = _longest_common_prefix_length(candidates[0], candidates[1])
    raise AmbiguousDigestError(
        f"Short digest {key} is ambiguous; expands to: {list(candidates)}; need at least {lcp + 1} characters."
    )
```

Cache-layer callers pass `dedupe=True`; storage-layer callers keep `dedupe=False` (default).

Note: `_combine_shrink` (caches.py:251) is **not** part of this refactor — it picks the *longest* prefix from multiple sub-storages and has different semantics (no LCP, no `AmbiguousDigestError`).

## Pros

- Two near-identical functions become one.
- The richer "expands to: ..." error message becomes available at the storage layer too, which is useful when debugging SQL `expand` collisions.
- Single place to fix wording / add a `len > 2` debugging aid (currently both helpers only look at the first two candidates).

## Cons

- Adds a `dedupe` flag that only one of the two call sites uses. Mild API surface bloat in exchange for the deduplication.
- The cache-layer caller currently lives in `caches.py`; the helper would live in `storage/base.py`. That's a small additional cross-module import (`caches.py` already imports `_longest_common_prefix_length` from `storage.base`, so the dependency direction is already correct).
- Net code delta: ~10 lines removed. Small win.

## Scope

- One helper added in `storage/base.py` (or a tiny `storage/_prefix.py`).
- `_combine_expand` deleted from `caches.py`; the two `Cache`/`CacheStack` callsites adjusted to pass `dedupe=True`.
- No tests need to move — both call paths are already covered by `tests/unit/caches/test_expand_shrink.py` and `tests/unit/storage/test_short_digest.py`.

pmrv:
Closes #661.

## Problem

`_resolve_prefix` (`storage/base.py`) and `_combine_expand` (`caches.py`) implemented the same "unique candidate digest or raise `KeyError`/`AmbiguousDigestError`" logic. The only real differences were: `_combine_expand` deduplicates its input first (multiple sub-storages can yield the same digest, while `_resolve_prefix`'s callers already pre-slice to at most two candidates), and `_combine_expand`'s ambiguous-error message additionally listed the candidates ("expands to: ...").

## Change

- `_resolve_prefix` (`storage/base.py`) gained a `dedupe: bool = False` keyword-only argument. When `True`, candidates are reduced to `sorted(set(candidates))` before the existing unique/raise logic runs. The error message now always includes the "expands to: [...]" detail, so the storage layer gets the richer debugging message too (useful for SQL `expand` collisions).
- `_combine_expand` is deleted from `caches.py`. `Cache.expand` and `CacheStack.expand` now call `_resolve_prefix(key, results, dedupe=True)` instead.
- `_combine_shrink` (`caches.py`) is untouched — it picks the *longest* prefix from multiple sub-storages and has different semantics (no LCP, no `AmbiguousDigestError`), as noted in the issue.
- Updated the two `agents/DEVELOPING.md` references to the old `_combine_expand` name.

## Tests

No test relocation needed — both call paths were already covered by `tests/unit/caches/test_expand_shrink.py` and `tests/unit/storage/test_short_digest.py`; both pass unchanged (the latter's `"need at least 11 characters" in str(excinfo.value)` assertion still holds since it's a substring match, and the message is now more detailed rather than different).

Full suite: `1522 passed, 2 failed, 11 skipped` on this branch — the 2 failures (`tests/integration/test_parallel_execution.py::test_process_executor_bound_wrapper`, `test_wrap_executor_does_not_double_bind_bound_wrapper`) are pre-existing test-order flakiness, confirmed to fail identically on unmodified `main` when run as part of the same full `tests/unit tests/regression tests/integration` invocation, and to pass in isolation on both `main` and this branch.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---
_Generated by [Claude Code](https://claude.ai/code/session_012JTMyPKmLxEQyJtXqiCu7M)_


pmrv:
@claude ^

pmrv (commented):
@claude fix merge conflict, then merge

pmrv:
> @claude fix merge conflict, then merge

@claude One more time.

PR #743, opened 2026-07-13
docs: document that functions returning None are never cached

pmrv:
## Summary

`wrapper.py` silently skips caching when a decorated function returns `None`, emitting only a `WARNING` log. This behaviour was entirely absent from the documentation, and the `.rerun()` description incorrectly claimed it "saves the newly computed result to the cache" without qualification.

### Changes in `docs/usage/helpers.rst`

- **Corrected `.rerun()` description** — added "(unless the result is `None` — see below)" so the claim is accurate.
- **New "Functions returning `None`" section** — explains the behaviour, lists the affected code paths (normal calls and `.rerun()`), and provides the recommended workaround (return a sentinel value instead of `None`).

## Why the existing docs were wrong

From `src/fleche/wrapper.py` (`make_wrapper` → `_cache` closure):

```python
if call.result is None:
    logger.warning("Function returned None, not caching")
    return None
```

This check runs unconditionally for every code path that reaches `_cache()`, including `.rerun()`. A user calling `.rerun()` on a function that now returns `None` would expect the cache to be updated, but no write happens and no error is raised — only a WARNING in the logs.

---
_Generated by [Claude Code](https://claude.ai/code/session_012BWYfzZA1BrqkYsHfbwzJr)_

### DIFF
--- a/docs/storage/configuration.rst
+++ b/docs/storage/configuration.rst
@@ -447,3 +447,27 @@ Below is an example of a complete configuration file demonstrating several featu
    values.root = "/shared/team/.cache/fleche/values"
    calls.type = "cloudpickle"
    calls.root = "/shared/team/.cache/fleche/calls"
+
+.. _none-not-cached:
+
+Functions Returning ``None``
+-----------------------------
+
+Functions that return ``None`` are **never cached**.  When a decorated function
+returns ``None``, ``fleche`` logs a ``WARNING`` and skips the save step
+entirely.  Subsequent calls will execute the function again rather than
+returning a cached value.
+
+This applies to all code paths, with one difference for ``.rerun()``:
+
+- A normal call that returns ``None`` does not cache.
+- ``.rerun()`` re-executes the function and still does not cache if the new
+  result is ``None`` — but since a prior cached entry may now be stale, it is
+  **evicted** so that later calls fall through to re-execution rather than
+  returning the old value. If the active cache rejects the eviction (e.g. a
+  read-only cache), ``fleche`` only logs a ``WARNING`` — the stale entry is
+  left in place, and you must evict it yourself::
+
+     >>> from fleche import cache
+     >>> key = my_func.digest(*args, **kwargs)
+     >>> cache().evict(key)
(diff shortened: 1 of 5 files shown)


pmrv:
@claude sounds to me like the most consistent thing would be to evict a prior results under .rerun iff new result is None.  Build a minimum viable PR, I only want to judge whether being consistent here is worth the code complexity.

docs/usage/helpers.rst
```diff
@@ -83,6 +83,26 @@ Optionally pre-applies ``*args`` and ``**kwargs`` via :func:`functools.partial`.
...
+This applies uniformly to all code paths:
+
+- A normal call that returns ``None`` does not cache.
+- ``.rerun()`` re-executes the function but still does not cache if the new
+  result is ``None``.
+
+To cache the fact that a function produced "no result", return a sentinel value
+instead of ``None`` (e.g. ``[]``, ``""``, or a dedicated singleton object), and
+convert it back to ``None`` at the call site.
+
```
pmrv: Do not advertise work arounds.  If anything, state that evict has to be called by users.

docs/usage/helpers.rst
```diff
@@ -83,6 +83,26 @@ Optionally pre-applies ``*args`` and ``**kwargs`` via :func:`functools.partial`.
 
 See :class:`~fleche.state.BoundWrapper` for the full API, including pickling support.
 
+.. _none-not-cached:
+
+Functions returning ``None``
```
pmrv: Should live on a more generic page, since this is not specific to helpers.

pmrv:
@claude Fix comment s^

docs/storage/configuration.rst
pmrv, line 1: @claude not here either.  split it up actually.  Mention that None functions are never cached in the new TL;DR section (rebase on main) and move the rerun specific section back into the helper page.

pmrv:
@claude

> @claude not here either. split it up actually. Mention that None functions are never cached in the new TL;DR section (rebase on main) and move the rerun specific section back into the helper page.

PR #744, opened 2026-07-13
docs: document that CacheStack cannot be nested inside another CacheStack

pmrv:
## Summary

`CacheStack.__post_init__` raises `ValueError` when any member of the stack is itself a `CacheStack`. This constraint was not mentioned anywhere in the documentation, so users who tried to compose stacks by nesting them would hit an opaque runtime error with no guidance on the correct alternative.

### Changes in `docs/storage/cache_stack.rst`

- **New "Restrictions" section** — states the constraint, shows the failing pattern and the error it produces, and provides the correct fix (flatten all layers into a single `CacheStack`).

## Why this matters

From `src/fleche/caches.py`:

```python
def __post_init__(self):
    for c in self.stack:
        if isinstance(c, CacheStack):
            raise ValueError("CacheStack cannot be nested inside another CacheStack")
```

A user who reads that `CacheStack` composes multiple caches into a hierarchy would reasonably try:

```python
inner = CacheStack((fast_cache, medium_cache))
outer = CacheStack((inner, slow_cache))  # ValueError!
```

The fix is straightforward — flatten to `CacheStack((fast_cache, medium_cache, slow_cache))` — but nothing in the docs pointed this out.

---
_Generated by [Claude Code](https://claude.ai/code/session_012BWYfzZA1BrqkYsHfbwzJr)_

### DIFF
--- a/docs/storage/cache_stack.rst
+++ b/docs/storage/cache_stack.rst
@@ -41,3 +41,16 @@ Example
     ...     my_function(10)
 
 Automatic hit transfer only applies to full function calls (``Call`` objects) and not to individual values loaded via ``load_value``.
+
+Restrictions
+------------
+
+``CacheStack`` **cannot be nested**: passing a ``CacheStack`` as a member of
+another ``CacheStack`` raises ``ValueError`` immediately::
+
+   inner = CacheStack((cache_a, cache_b))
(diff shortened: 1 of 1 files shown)


pmrv (commented):
@claude advertise the stack method

PR #748, opened 2026-07-13
feat: refix() re-shards bagofholding prefix length; validate layout at init

pmrv:
## What

Prefix-length lifecycle management for `BagOfHoldingH5FileBackend`:

- **`prefix_length` reinterpreted** — `0` is one file per key (TOML-expressible, replacing the old `None`), `None` infers the length from the files already in `root` (default `2` on an empty root, error on a mixed one), so the field is always an int after construction. Since multi-bag files are named `{prefix}.h5` and per-key files by the full 64-char digest, the length any file was written with can be read off its name.
- **Init-time layout check** — `__post_init__` validates `prefix_length` and raises `ValueError` when files already in `root` were written with a different prefix length, instead of silently leaving those entries unreachable. The init-only `check_consistency=False` kwarg (requires an explicit `prefix_length`) skips the check and makes the instance *blind*: it operates only on files of exactly its own length, which is how one layout of a mixed root is addressed.
- **`refix(n)`** — re-shards every stored entry of a live storage to an explicit target length (`0` = per-key), in any direction. Each entry is copied before its original is removed; per-key originals unlink immediately and old bags unlink whole as soon as they are drained (HDF5 files don't shrink on group deletion), bounding transient extra disk usage to one entry / one bag. Unreadable entries abort with `RuntimeError` — silently skipping would fail the *next* instantiation — and entries already present in the target layout are skipped, so aborted migrations resume without re-copying.
- **`consolidate(root, n)`** — classmethod repair constructor: refixes every other prefix length found in `root` to `n` and returns the final, consistency-checked storage.
- **`put` hashes when no key is given** — `StorageBackend.put(value)` digests the value across the file/memory/void backends.

## Testing

- 21 tests in `tests/unit/storage/test_bagofholding_file.py` covering every refix direction with multiple keys per bag (split, merge, from/to per-key, with h5py-level sibling-group assertions), noop refix never touching the filesystem, drained bags unlinking before the next bag is copied, fail-fast aborts on unreadable entries, blind instances, `None`-inference (existing / empty / mixed roots), mixed-root repair via `consolidate`, and init-check rejection/tolerance cases.
- Config round-trip tests updated for the resolved-int semantics.
- Full suite passes (1605 passed, 11 skipped); `ty check src/` clean.

Docs updated in `docs/storage/configuration.rst` (with `:meth:` cross-references), `src/fleche/config.py`'s docstring, `agents/USAGE.md`, and `agents/DEVELOPING.md`.

https://claude.ai/code/session_01X6svvmiw1mv3NejCUGv6sH


src/fleche/storage/bagofholding_file.py
```diff
@@ -37,6 +46,68 @@ class BagOfHoldingH5FileBackend(FileStorage):
...
+            if observed != self.prefix_length:
+                raise ValueError(
+                    f"prefix_length={self.prefix_length} does not match "
+                    f"{p.name!r} in {self.root}, which was written with "
+                    f"prefix_length={observed}; open the storage with "
+                    f"prefix_length={observed} and call "
+                    f"refix({self.prefix_length}) to migrate it."
+                )
+
+    def refix(self, prefix_length: int | None) -> None:
```
pmrv: need to ensure original files are deleted, and as soon as possible to avoid ballooning storage.

pmrv: Addressed in 2051204 follow-up. Per-key originals were already unlinked immediately after each copy, but the multi-bag side had a real problem: evicting per key only deletes the HDF5 *group*, and HDF5 files don't shrink on group deletion — every old bag would have stayed at full size until its last entry migrated. `refix` now drains one bag at a time and unlinks the whole file as soon as all of its entries are copied out, so transient extra disk usage is bounded by a single bag (or a single entry in per-key mode). Entries that fail to read stay behind in the old bag rather than being dropped.

---

src/fleche/storage/bagofholding_file.py
```diff
@@ -37,6 +46,117 @@ class BagOfHoldingH5FileBackend(FileStorage):
...
+                        value = self.get(key)
+                    except KeyError:
+                        logger.warning(
+                            "Skipping unreadable entry %s during prefix-length migration", key
+                        )
+                        continue
+                    target.put(value, key)
+                    self._evict(key)
+        else:
+            for bag_path in sorted(self.root.glob("*.h5")):
```
pmrv: try `sorted(self.list())` instead to force collection before migration starts.

docs/storage/configuration.rst
```diff
@@ -277,12 +277,17 @@ Since digests are SHA256 hex strings, the default ``prefix_length = 2``
...
-from Python — TOML cannot express ``None``) for the one-file-per-key layout.
+bucket sizes stay roughly even without any extra bookkeeping. Set it to
+``None`` (only possible when constructing the backend from Python — TOML
+cannot express ``None``) for the one-file-per-key layout.
+
+``prefix_length`` is checked against the files already present in ``root``
+when the storage is constructed: opening an existing cache directory with a
+different ``prefix_length`` raises a ``ValueError`` instead of silently
+leaving the old entries unreachable.  To re-shard an existing cache, open it
+with its current ``prefix_length`` and call ``refix(new_prefix_length)``,
```
pmrv: full sphinx reference to API doc page

src/fleche/storage/bagofholding_file.py
```diff
@@ -37,6 +47,113 @@ class BagOfHoldingH5FileBackend(FileStorage):
...
+            if observed != self.prefix_length:
+                raise ValueError(
+                    f"prefix_length={self.prefix_length} does not match "
+                    f"{p.name!r} in {self.root}, which was written with "
+                    f"prefix_length={observed}; open the storage with "
+                    f"prefix_length={observed} and call "
+                    f"refix({self.prefix_length}) to migrate it."
+                )
+
+    def refix(self, prefix_length: int | None) -> None:
```
pmrv: need a method that is able to be called from an inconsistent state, so that consumers can fix broken state when previously aborting a refix.

Guard the consistency check in post init behind an init-only kwarg, default True.  It can only be False if an explicit prefix length is given.  The storage should then blindly operate only on files of that length.

refix should have the previos shape of operating on two separate storage instances again.

Add a classmethod constructor that inspects all present prefix lengths and takes a target length, converting all other lengths it finds by their refix method and returns the final storage

tests/unit/storage/test_bagofholding_file.py
```diff
@@ -255,6 +255,189 @@ def test_multi_bag_different_prefix_lengths(tmp_path):
     assert s4.get(key) == "long-prefix"
 
 
+def test_refix_from_per_key(tmp_path):
+    pytest.importorskip("bagofholding")
+    s = BagOfHoldingH5FileBackend(tmp_path, prefix_length=None)
+    key1 = Digest("ab" + "1" * 62)
+    key2 = Digest("cd" + "2" * 62)
+    s.put("first", key1)
+    s.put("second", key2)
```
pmrv:
```suggestion
    key1 = s.put("first")
    key2 = s.put("second")
```

pmrv: same it test cases below

tests/unit/storage/test_bagofholding_file.py
```diff
@@ -255,6 +255,189 @@ def test_multi_bag_different_prefix_lengths(tmp_path):
...
+
+    s.refix(None)
+
+    assert s.prefix_length is None
+    assert not list(tmp_path.glob("*.h5"))
+    assert (tmp_path / key1).is_file()
+    assert (tmp_path / key2).is_file()
+    assert s.get(key1) == "first"
+    assert s.get(key2) == "second"
```
pmrv: disallow `refix(None)` unnecessary API surface.  If someone wants to change prefix they better be specific

tests/unit/storage/test_bagofholding_file.py
```diff
@@ -255,6 +255,189 @@ def test_multi_bag_different_prefix_lengths(tmp_path):
...
+
+
+def test_refix_noop(tmp_path):
+    pytest.importorskip("bagofholding")
+    s = BagOfHoldingH5FileBackend(tmp_path, prefix_length=2)
+    key = _digest_like("a")
+    s.put("value", key)
+
+    s.refix(2)
```
pmrv: additionally test that this never this filesystem.

tests/unit/storage/test_bagofholding_file.py
```diff
@@ -255,6 +255,189 @@ def test_multi_bag_different_prefix_lengths(tmp_path):
...
+    assert s.prefix_length == 4
+    assert {p.name for p in tmp_path.glob("*.h5")} == {"ab11.h5", "ab22.h5"}
+    assert s.get(key1) == "first"
+    assert s.get(key2) == "second"
+
+
+def test_refix_noop(tmp_path):
+    pytest.importorskip("bagofholding")
+    s = BagOfHoldingH5FileBackend(tmp_path, prefix_length=2)
+    key = _digest_like("a")
```
pmrv: useless complication, see above, just let put hash

tests/unit/storage/test_bagofholding_file.py
```diff
@@ -255,6 +255,189 @@ def test_multi_bag_different_prefix_lengths(tmp_path):
...
+        if path.name == key1:
+            raise KeyError(path)
+        return real_from_file(self, path)
+
+    monkeypatch.setattr(BagOfHoldingH5FileBackend, "_from_file", failing_from_file)
+
+    s.refix(None)
+
+    # the migrated sibling moved out and was dropped from the old bag, the
+    # unreadable entry stayed behind
```
pmrv: refix should probably fail here to alert users.  otherwise their *next* instantiation is going to fail

src/fleche/storage/bagofholding_file.py
```diff
@@ -37,6 +47,113 @@ class BagOfHoldingH5FileBackend(FileStorage):
     def __post_init__(self):
         if hasattr(super(), "__post_init__"):
             super().__post_init__()
+        _validate_prefix_length(self.prefix_length)
+        self._check_prefix_consistency()
+
+    def _check_prefix_consistency(self) -> None:
```
pmrv, line 99: reinterpret prefix_length.  `0` is one file per key, `None` is automatic from existing values, but check consistency.

tests/unit/storage/test_bagofholding_file.py
pmrv: misinterpreted argument value on my part, but see very top comment, this should be 0.

pmrv (commented):
none of the tests exercise the case where multiple values actually land in one file,  add more keys to all tests.

src/fleche/storage/bagofholding_file.py
pmrv: All implemented in 1fb4e38, together with the other threads from this round:

- `prefix_length` reinterpreted: `0` = one file per key, `None` = inferred from the files already in `root` (default 2 on an empty root, error on a mixed one), so it's always an int after construction. TOML can now express the per-key layout too.
- `check_consistency` is an `InitVar` (default `True`); `False` requires an explicit `prefix_length` and makes the instance blind — `list`/`contains`/`get` only see files of exactly that length.
- `refix` builds its target with the real constructor (`dataclasses.replace(..., check_consistency=False)`) instead of the `copy.copy` hack, rejects `None`/out-of-range targets, and raises `RuntimeError` on an unreadable entry instead of silently skipping (which would have failed the next instantiation). Entries already present in the target layout are skipped, so an aborted migration resumes instead of re-copying.
- `consolidate(root, n)` classmethod refixes every other prefix length found in `root` to `n` and returns the final (consistency-checked) storage — the repair path for aborted refixes.
- Tests: every refix direction now moves multiple keys per bag (split, merge, from/to per-key, with h5py-level sibling-group assertions), the noop refix is asserted to never touch the filesystem (no `H5Bag.save` calls, identical mtimes/sizes), and `put` hashes the value when no key is passed so tests only hand-craft digests where prefix co-location is the point.
- `configuration.rst` now uses full `:meth:` cross-references for `refix`/`consolidate`.

---

src/fleche/storage/bagofholding_file.py
```diff
@@ -24,19 +25,191 @@
...
+            raise ValueError(
+                f"prefix_length={self.prefix_length} does not match existing "
+                f"files in {self.root} written with prefix length(s) "
+                f"{sorted(others)} (0 = one file per key); open the storage "
+                f"with the matching prefix_length and call "
+                f"refix({self.prefix_length}) to migrate, or use "
+                f"{type(self).__name__}.consolidate() to unify a mixed root."
+            )
+
+    def refix(self, prefix_length: int) -> None:
```
pmrv: docstring for argument google style

pmrv: should return the refix'd storage

src/fleche/storage/bagofholding_file.py
```diff
@@ -24,19 +25,191 @@
...
+                    self._evict(key)
+        else:
+            for bag_path, bag_keys in itertools.groupby(keys, key=self._bag_file):
+                migrated = []
+                for key in bag_keys:
+                    with self._operation_context(key):
+                        self._refix_one(target, key)
+                        migrated.append(key)
+                self._drop_bag(bag_path, migrated)
+        object.__setattr__(self, "prefix_length", prefix_length)
```
pmrv: leave self invariant

src/fleche/storage/bagofholding_file.py
```diff
@@ -24,19 +25,191 @@
...
+        _validate_prefix_length(prefix_length)
+        root = Path(root)
+        probe = cls(root, prefix_length=prefix_length, check_consistency=False, **kwargs)
+        for length in sorted(_observed_prefix_lengths(probe.root) - {prefix_length}):
+            cls(root, prefix_length=length, check_consistency=False, **kwargs).refix(
+                prefix_length
+            )
+        return cls(root, prefix_length=prefix_length, **kwargs)
+
+    def _drop_bag(self, bag_path: Path, migrated: list[Digest]) -> None:
```
pmrv: is it really necessary (ie. are we missing something in our _evict?), that evict drop the groups and files as soon as possible.

src/fleche/storage/base.py
```diff
@@ -237,7 +237,8 @@ class StorageBackend(KeyManagement):
     """
 
     @abstractmethod
-    def put(self, value: Any, key: Digest) -> Digest: ...
+    def put(self, value: Any, key: Digest | None = None) -> Digest:
+        """Store *value* under *key*, digesting the value when no key is given."""
```
pmrv: I gave bad feedback on last review round.  Do not mess with the ABC's

tests/unit/storage/test_bagofholding_file.py
```diff
@@ -255,6 +255,342 @@ def test_multi_bag_different_prefix_lengths(tmp_path):
     assert s4.get(key) == "long-prefix"
 
 
+def test_refix_from_per_key(tmp_path):
```
pmrv: build yourself a fixture of prefilling storage. hypothesis indexed by size?  anything

src/fleche/storage/bagofholding_file.py
pmrv: You're right — `_evict` already unlinks the bag the moment its last group goes, and since keys migrate in sorted order each bag is fully drained before the next is touched, so plain per-key eviction gives exactly the same disk-usage bound. `_drop_bag` only saved N−1 redundant group deletions per bag, which doesn't justify its existence. Removed in 68964c5; `refix` is back to one loop over `sorted(self.list())` with `_evict` per key. That commit also covers the rest of this round: `refix` returns the re-fixed storage and leaves `self` invariant, Google-style docstrings on `refix`/`consolidate`, the `put` ABC change reverted (tests prefill through a shared `_prefill` helper with controlled prefixes instead, parametrized by entries per bag), and the docs updated to match.

---

pmrv (commented):
lgtm

PR #749, opened 2026-07-13
feat(storage): default prefix_length=2 and remaining_depth=1

pmrv:
## Summary

Changes two storage defaults to better-performing values:

- **`BagOfHoldingH5FileBackend.prefix_length`: `None` → `2`** — the HDF5 backend now multiplexes keys into up to 256 shared `.h5` files (`00.h5` .. `ff.h5`) by default instead of writing one file per key, avoiding HDF5's fixed per-file overhead when caching many small results. Pass `prefix_length=None` for the old per-key layout (Python only — TOML cannot express `None`).
- **`DestructuringMixin.remaining_depth`: `0` → `1`** — scalars are now inlined into their parent container by default, so a flat list/dict of scalars becomes a single cache entry instead of one entry per leaf. Nested sub-collections are still destructured. Set `remaining_depth=0` for the old maximum splitting.

## Details

- Updated the config module docstring, `docs/storage/configuration.rst`, `agents/USAGE.md`, `agents/DEVELOPING.md`, and the `StorageBackends` notebook for the new defaults.
- Test updates:
  - `tests/fixtures.py`: the plain `h5` fixture params now pin `prefix_length=None` so both layouts stay covered alongside `h5_multi`.
  - `test_bagofholding_file.py`: tests that write per-key files via `_path()` pin `prefix_length=None`; added `test_prefix_length_default_is_2`.
  - Destructuring/GC tests that assert on the fully-split raw storage structure pin `remaining_depth=0` explicitly.
  - Config serialization expectations updated (`remaining_depth: 1`, default `prefix_length == 2`).

Note: `prefix_length` is fixed for the lifetime of a cache directory, so existing `bagofholding_hdf` caches written with the old per-key layout will not be found under the new default — users with existing caches need to set `prefix_length = None` (from Python) or migrate.

## Testing

- Full suite: 1584 passed, 11 skipped.
- `ty check src/` passes.

https://claude.ai/code/session_01WxsSw4sN2LvzhdP9EgGoEv

---
_Generated by [Claude Code](https://claude.ai/code/session_01WxsSw4sN2LvzhdP9EgGoEv)_


pmrv:
fix merge conflicts

PR #765, opened 2026-07-20
docs(README): fix three factual errors

pmrv:
## Summary

Three factual errors found by cross-referencing the README against source code:

1. **Non-existent JSON backend** — the features list claimed "File (pickle, JSON)" backends. No JSON storage backend exists; the file backends are `pickle`, `cloudpickle`, and `dill` only. Fixed to match what `config.py` actually maps.

2. **Wrong `isolate` description** — the decorator-options code block said `# Isolate cache by working directory`. `isolate=True` creates a temporary working directory for the *function call* via `os.chdir`; it does not affect the cache at all. Also added the thread-safety caveat (`os.chdir` is process-wide, so `isolate=True` is explicitly not safe with `ThreadPoolExecutor`).

3. **Unqualified "Thread-Safe" feature bullet** — the bullet said "Safe for use in multi-threaded environments" with no qualification, even though `isolate=True` is documented in `wrapper.py` as not thread-safe. Added pointer to the `isolate=True` caveat.

**Additional fixes:**
- Made the `fetch_data` query example self-contained (replaced undefined `get_data` call that would raise `NameError` if run literally)
- Fixed the inline comment on `fetch_data.query(user_id=123)` — omitted arguments act as wildcards, not arguments passed as `None`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gmvhzu4udySWtYVSoNuD2P)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ A persistent caching solution for arbitrary Python functions - like `lru_cache` 
 - **Query Support**: Search and retrieve cached results with metadata filtering
 - **Configurable**: Control what gets hashed (version, module, code, arguments)
 - **Multiple Backends**: File (pickle, cloudpickle, dill), SQLAlchemy, Bagofholding, and more
-- **Thread-Safe**: Safe for concurrent use; see `isolate=True` caveat below
+- **Thread-Safe**: Safe for use in multi-threaded environments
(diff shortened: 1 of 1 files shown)


pmrv (commented):
@claude drop any mention of isolate, we'll get rid of it soon

PR #777, opened 2026-07-20
feat(config): add cache templates and accept configs in cache()

### LINKED ISSUE #732: API suggestion -- new cache shortcut
I feel like it may be nice to extend `Cache` with a class method like

```python
@classmethod
def new(cls, style: str, **kwargs) -> Cache:
    match style:
        case "void":
            return cls(storage.ValueVoid(), storage.CallVoid())
        case "memory":
            return cls(storage.ValueMemory({}), storage.CallMemory({}))
        case "pickle":
            filename = kwargs.pop("filename")
            path = pathlib.Path(filename)
            return cls(
                values=ValuePickleFile.with_pickle(root=path / "values"),
                calls=CallPickleFile.with_pickle(root=path / "calls"),
            )
        case "bagofholding":
            ...
        ...
        case _:
            raise ValueError(f"Unrecognized cache style: {style}, please choose from...")

```

E.g. `my_cache = Cache.new("pickle", filename=".pickle_storage")`.

pmrv:
Closes #732.

## What

Adds a lightweight, flexible shorthand for building caches, resolving the API-shortcut request in #732 without the combinatorial `match`-per-backend template the issue sketched.

The core move follows the direction discussed on the issue: rather than a new public factory that enumerates every value×call backend combination, the shorthand is expressed **as config** and expanded inside the existing `cache_from_config` — so it works identically in `fleche.toml` and in Python.

### `template` key

A cache config may name a `template` plus the storage args that shape requires:

```toml
[terse]
template = "cloudpickle"          # -> values at root/values, calls at root/calls
root = "~/.fleche"

[sqlbacked]
template = "sql"                  # cloudpickle values + SQL call storage
root = "~/.fleche/values"
url = "sqlite:///~/.fleche/calls.db"
```

Two categories keep the surface small:

- **Symmetric** templates (`memory`, `void`, `pickle`, `cloudpickle`, `dill`, `bagofholding_hdf`) use one backend for both values and calls; the filesystem ones split a single `root` into `root/values` + `root/calls`.
- **`sql`** pairs `cloudpickle` values (`root`) with SQL call storage (`url`).

Each template consumes only the args it needs; a missing or unexpected arg raises a clear `ValueError` pointing back to the explicit `values`/`calls` form. `read_only` / `max_size` compose with templates, and templates compose inside stacks/pools via the existing recursion. Anything a template doesn't cover (mixed backends, per-backend options like `compress`/`secret_key`) falls back to the explicit config — the template stays small precisely because it declines to grow.

### `cache()` accepts configs

`cache()` now takes a `dict` or `list` and routes it through `cache_from_config`, so an ad-hoc cache can be built and activated inline:

```python
cache({"template": "pickle", "root": ".cache"})
```

No new public API function is added — the entry points are the existing `cache()` and `fleche.toml`.

## Why

Directly addresses the "too many value/call combinations" concern raised on #732: the machinery to turn config dicts into live caches already existed but wasn't reachable ergonomically. This exposes it through the terse-common-case template while leaving the full dict form for everything else.

## Testing

- New template tests in `tests/unit/config/test_cache_from_config.py` (symmetric transient/file, cloudpickle, sql, read_only/max_size composition, stack composition, no-mutation, and the three error paths).
- New `cache()`-accepts-config tests in `tests/unit/test_cache_sticky.py`.
- Full unit suite: `1553 passed`. `ty check src/`: clean. Module doctests pass. Verified end-to-end that a template built via `cache({...})` caches persistently and splits values/calls into separate subdirectories, and that the same works from a `fleche.toml`.

## Docs

Updated the `config.py` module docstring, `agents/USAGE.md`, and `docs/storage/configuration.rst`.

https://claude.ai/code/session_01RR4Kj28gDCL9WNYBP1FG5s

---
_Generated by [Claude Code](https://claude.ai/code/session_01RR4Kj28gDCL9WNYBP1FG5s)_


docs/storage/configuration.rst
```diff
@@ -144,6 +144,36 @@ You can define multiple cache configurations in the same file, each in its own s
...
+.. code-block:: toml
+
+   [terse]
+   template = "cloudpickle"     # both backends cloudpickle...
+   root = "~/.fleche"           # ...values at root/values, calls at root/calls
+
+   [sqlbacked]
+   template = "sql"             # cloudpickle values + SQL call storage
+   root = "~/.fleche/values"
+   url = "sqlite:///~/.fleche/calls.db"
```
pmrv: shouldn't be required in this template.  root would be `.fleche` (should have a different key name though, say folder, directory?) and root/url would derive from it

pmrv: correction: that's the default.  URL for sql should still be overridable.

src/fleche/config.py
```diff
@@ -440,6 +464,94 @@ def storage_to_config(s: storage.ValueStorage | storage.CallStorage) -> dict[str
...
+# pair of storage configs.  The builders take explicit keyword arguments so a
+# missing or unexpected key surfaces as a clear error (see cache_from_config).
+_CACHE_TEMPLATES: "dict[str, Callable[..., tuple[dict[str, Any], dict[str, Any]]]]" = {
+    "memory": _template_symmetric_transient("memory"),
+    "void": _template_symmetric_transient("void"),
+    "pickle": _template_symmetric_file("pickle"),
+    "cloudpickle": _template_symmetric_file("cloudpickle"),
+    "dill": _template_symmetric_file("dill"),
+    "bagofholding_hdf": _template_symmetric_file("bagofholding_hdf"),
+    "sql": _template_value_plus_sql("cloudpickle"),
```
pmrv: should be paired with any value storage, not just cloudpickle

pmrv (commented):
revise

pmrv:
Revised the `sql` template per the review (pushed in d8d48b5):

- **`url` no longer required** — it defaults to `sqlite:///root/calls.db`, derived from a single `root`, and is still overridable with an explicit `url`.
- **Value backend is selectable** via `values` (default `cloudpickle`), so `sql` pairs with any filesystem value storage rather than hardcoding cloudpickle.
- Values now live under `root/values`, consistent with the symmetric templates.

So the terse form is now just:

```toml
[sqlbacked]
template = "sql"
root = "~/.fleche"     # values -> root/values, calls -> sqlite:///root/calls.db
# optional: values = "pickle", url = "postgresql://..."
```

One open point I made a call on: you flagged the base key should maybe be renamed (`folder`/`directory`?). I kept it as **`root`** for now — it matches the symmetric templates and your own "root would be `.fleche`" wording. Happy to rename it (across all templates for consistency, or sql-only) if you'd prefer `directory`/`folder` — just say which.

---

src/fleche/config.py
```diff
@@ -440,6 +468,103 @@ def storage_to_config(s: storage.ValueStorage | storage.CallStorage) -> dict[str
...
+    def build(root: str) -> "tuple[dict[str, Any], dict[str, Any]]":
+        base = Path(root)
+        return (
+            {"type": style, "root": str(base / "values")},
+            {"type": style, "root": str(base / "calls")},
+        )
+    return build
+
+
+def _template_sql(default_value_style: str = "cloudpickle") -> "Callable[..., tuple[dict[str, Any], dict[str, Any]]]":
```
pmrv:
```suggestion
def _template_sql(default_value_style: str = "bagofholding_hdf") -> "Callable[..., tuple[dict[str, Any], dict[str, Any]]]":
```

src/fleche/config.py
```diff
@@ -440,6 +468,103 @@ def storage_to_config(s: storage.ValueStorage | storage.CallStorage) -> dict[str
...
+    return build
+
+
+# Named cache templates.  Each entry maps a ``template`` string to a builder
+# that turns the remaining (required) config keys into a ``(values, calls)``
+# pair of storage configs.  The builders take explicit keyword arguments so a
+# missing or unexpected key surfaces as a clear error (see cache_from_config).
+_CACHE_TEMPLATES: "dict[str, Callable[..., tuple[dict[str, Any], dict[str, Any]]]]" = {
+    "memory": _template_symmetric_transient("memory"),
+    "void": _template_symmetric_transient("void"),
```
pmrv: drop

pmrv (commented):
revise

pmrv:
@liamhuber  What do you think about this?  It's a terser dict accessible via `fleche.cache()` rather then a classmethod.  Your example should work out to `fleche.cache(dict(template="pickle", root="./././"))`.

src/fleche/state.py
```diff
@@ -38,12 +38,14 @@ def cache(new_cache: None = None, stack: bool = False) -> caches.BaseCache: ...
 
 @overload
 def cache(
-    new_cache: caches.BaseCache | str, stack: bool = False
+    new_cache: "caches.BaseCache | str | dict[str, Any] | list[dict[str, Any]]",
+    stack: bool = False,
 ) -> AbstractContextManager[None]: ...
 
 
 def cache(
```
pmrv, line 45: let's see if the signature is still readable if we also take **kwargs -> new_cache.

pmrv:
Tried the `**kwargs -> new_cache` idea (pushed in def32c0). The runtime body stays short — one guard up top:

```python
def cache(new_cache=None, stack=False, **kwargs):
    if kwargs:
        if new_cache is not None:
            raise TypeError("... positional or keyword config, not both")
        new_cache = kwargs
    ...
```

So both spellings work:

```python
cache(template="pickle", root="./././")          # new
cache({"template": "pickle", "root": "./././"})  # still fine
```

One wrinkle worth a look: the signature needs a **third `@overload`** to keep return types honest, because `cache()` (query → returns the active cache) and `cache(template=...)` (activate → returns the sticky context manager) both leave `new_cache` at its default:

```python
@overload
def cache(new_cache: None = None, stack: bool = False) -> BaseCache: ...           # query
@overload
def cache(new_cache: BaseCache | str | dict | list, stack: bool = False) -> AbstractContextManager[None]: ...
@overload
def cache(*, stack: bool = False, **kwargs: Any) -> AbstractContextManager[None]: ...  # kwargs form
```

`ty` is happy with it and the three overloads read cleanly enough, but it is one more overload and `**kwargs` loosens arg-checking (a typo'd key becomes an unknown-storage-key error at build time rather than a `TypeError` at the call). Happy to keep it or drop back to dict-only — your call on whether the terseness is worth the extra overload.

---

liamhuber:
> @liamhuber What do you think about this? It's a terser dict accessible via `fleche.cache()` rather then a classmethod. Your example should work out to `fleche.cache(dict(template="pickle", root="./././"))`.

Hm, not quite. In my application, I store a `Cache` instance, and then turn it into a context at a later date. Here, `fleche.state.cache` claims to return a `"caches.BaseCache | AbstractContextManager[None]"`, but in fact explicitly instantiates and returns a `fleche.state._StickyCache`, i.e. I always get the `AbstractContextManager[None]` and never the `caches.BaseCache`.

This means it works fine for JIT calls:

```python
import fleche

@fleche.fleche
def foo(x):
    return x

with fleche.cache(template="pickle", root="./foo"):
    y = foo(5)
```

But is useless as an actual `BaseCache` creator:

```python
import fleche

@fleche.fleche
def foo(x):
    return x

cache = fleche.cache(template="pickle", root="./foo")
with fleche.cache(cache):
    y = foo(5)
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[10], line 8
      5     return x
      7 cache = fleche.cache(template="pickle", root="./foo")
----> 8 with fleche.cache(cache):
      9     y = foo(5)
     10 y

File ~/dev/pyiron/fleche/src/fleche/state.py:103, in cache(new_cache, stack, **kwargs)
    101     resolved = config.cache_from_config(new_cache)
    102 else:
--> 103     raise ValueError(new_cache)
    105 if stack:
    106     resolved = _CACHE.get().push(resolved)

ValueError: <fleche.state._StickyContext object at 0x15bb0b190>
```

So, except for the misleading type hint, I don't have anything against these changes, but they don't provide the shorthand `Cache`-instance-creator I was hoping for.

liamhuber:
Unsurprisingly, I can't just carry around the `_StickyContext` in my case because I need to be able to ship it and 

```python
>>> import pickle
>>> pickle.loads(pickle.dumps(fleche.cache(template="pickle", root="./foo")))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[12], line 10
      7 cache = fleche.cache(template="pickle", root="./foo")
      9 import pickle
---> 10 pickle.loads(pickle.dumps(cache))

TypeError: cannot pickle '_contextvars.ContextVar' object

```

pmrv:
Right, I forgot the `cache()` gives you the result already wrapped in the context.  This function is terribly overloaded anyway.  I'll make a second function that you can call cleanly from outside.

pmrv:
Makes sense — reverted the `**kwargs` overload (0ab8f74), so `cache()` is back to just accepting a dict/list config. Left the rest in place for your dedicated function to build on.

For whenever you write it: `config.cache_from_config(dict | list)` is the reusable core that already does all the template/stack/pool/read_only/max_size expansion, so the clean external function can be a thin wrapper that returns the `BaseCache` directly (no context-manager wrapping). Happy to wire it up if you'd like, or I'll leave it to you.

---

pmrv:
> Makes sense — reverted the `**kwargs` overload ([0ab8f74](https://github.com/pmrv/fleche/commit/0ab8f74f16a21ce92025285a3b94d81a0767c8e4)), so `cache()` is back to just accepting a dict/list config. Left the rest in place for your dedicated function to build on.
> 
> For whenever you write it: `config.cache_from_config(dict | list)` is the reusable core that already does all the template/stack/pool/read_only/max_size expansion, so the clean external function can be a thin wrapper that returns the `BaseCache` directly (no context-manager wrapping). Happy to wire it up if you'd like, or I'll leave it to you.
> 

Expose `cache_from_config` in the global `fleche` API.  I think a classmethod wrapper on BaseCache would be fun, but will get us into recursive import territory.  Try both and I make the call later.

pmrv:
Both wired up in 1875b6c so you can compare:

**1. Top-level function**
```python
import fleche
c = fleche.cache_from_config({"template": "pickle", "root": ".cache"})
```

**2. Classmethod on `BaseCache`**
```python
c = Cache.from_config({"template": "pickle", "root": ".cache"})
```

The recursive-import worry is real but contained — `fleche.config` imports `caches` at module load, so `from_config` does the config import lazily inside the method body rather than at module top:

```python
@classmethod
def from_config(cls, config):
    from . import config as _config   # lazy: config imports this module
    return _config.cache_from_config(config)
```

Both delegate to the same `cache_from_config`, so the concrete type still comes from the config shape (`cls` is ignored — `Cache.from_config([...])` still returns a `CacheStack`). That "classmethod that ignores `cls`" is the one slightly odd thing about option 2, if that sways you. `import fleche; fleche.cache_from_config(...)` and `Cache.from_config(...)` both work now — tests and `ty` pass. Whichever you drop, I'll clean up.

---

liamhuber:
> Both wired up in https://github.com/pmrv/fleche/commit/1875b6cdd1b248d0c85ff8a8c09b4818d56811b1 so you can compare:

FWIW I like (2) better from the caller perspective, but (1) better from a maintainer perspective (I tolerate but dislike in-function imports). Both look like they solve my problem 🚀

docs/storage/configuration.rst
```diff
@@ -144,6 +144,39 @@ You can define multiple cache configurations in the same file, each in its own s
 
 Each cache section must define two storage backends: ``values`` and ``calls``. ``values`` is used to store the results of function calls, and ``calls`` is used to store the function call details.
 
+Cache templates
```
pmrv, line 147: you explanation is a bit muddy.  introduce general concept of template, then just explain the details and available keys of both

src/fleche/__init__.py
pmrv, line 1: drop in favor of 2

src/fleche/caches.py
```diff
@@ -33,6 +33,21 @@ class Rejected(Exception):
 
 class BaseCache(OperationContext):
 
+    @classmethod
+    def from_config(cls, config: "dict[str, Any] | list[dict[str, Any]]") -> "BaseCache":
```
pmrv, line 37: I like that this keeps my API surface contained, so let's take this route.

src/fleche/state.py
```diff
@@ -38,12 +38,14 @@ def cache(new_cache: None = None, stack: bool = False) -> caches.BaseCache: ...
 
 @overload
 def cache(
-    new_cache: caches.BaseCache | str, stack: bool = False
+    new_cache: "caches.BaseCache | str | dict[str, Any] | list[dict[str, Any]]",
```
pmrv: drop the dict path entirely, since we can now do `cache(BaseCache.from_config(...))`.

tests/unit/config/test_cache_from_config.py
```diff
@@ -175,3 +176,136 @@ def test_cache_sql(tmp_path):
...
+def test_template_with_read_only():
+    c = cache_from_config({"template": "memory", "read_only": True})
+    assert isinstance(c, ReadOnlyCache)
+    assert isinstance(c.cache, Cache)
+
+
+def test_template_with_max_size():
+    c = cache_from_config({"template": "memory", "max_size": 7})
+    assert isinstance(c, SizeLimitedCache)
+    assert c.max_size == 7
```
pmrv, line 259: parametrize better

pmrv (commented):
revise

pmrv:
> > Both wired up in [1875b6c](https://github.com/pmrv/fleche/commit/1875b6cdd1b248d0c85ff8a8c09b4818d56811b1) so you can compare:
> 
> FWIW I like (2) better from the caller perspective, but (1) better from a maintainer perspective (I tolerate but dislike in-function imports). Both look like they solve my problem 🚀

I'll go with 2 then, since it also avoids another top level public API function and that outweighs the import, imo.

pmrv (commented):
lgtm, merge

PR #778, opened 2026-07-20
fix(wrapper): make the cached wrapper cloudpicklable by value (drop raw lock from closure)

pmrv:
## What

Makes fleche's cached wrapper survivable through **by-value** cloudpickling, so `wrap_executor` payloads can actually be shipped to cross-process / cluster backends. Previously the wrapper closure closed over a raw `threading.Lock`, which is unpicklable.

## The bug

`wrap_executor`'s patched `submit` hands the underlying executor `func.fleche.bind(*args)` — a `BoundWrapper` wrapping `functools.partial(wrapper, *args)`, where `wrapper` is the closure built by `make_wrapper` (`src/fleche/wrapper.py`). That closure closed over:

```python
_in_flight_lock = threading.Lock()   # a raw _thread.lock
```

A cloudpickling cluster backend (executorlib, dask) must serialise that wrapper **by value** whenever the decorated function isn't importable by reference — and by value, cloudpickle walks the closure cells and hits the lock:

```
TypeError: cannot pickle '_thread.lock' object
```

### How easily by-value is triggered

More easily than `__main__`/notebooks or an explicit `register_pickle_by_value`. **Simply rebinding a decorated function to a different name is enough:**

```python
compute = fleche.fleche(some_other)   # name 'compute' != wrapper.__qualname__ 'some_other'
```

The wrapper carries `@wraps(some_other)`, so its `__qualname__` is `some_other`; cloudpickle's by-reference lookup finds the *plain* function (or nothing) under that name — not the wrapper — and falls back to by value. Verified in a plain importable module:

| pattern | by-ref lookup resolves to wrapper? | `cloudpickle.dumps(bind(3))` |
|---|---|---|
| `compute = fleche.fleche(some_other)` (rebind) | ✗ | **FAIL — `_thread.lock`** |
| `@fleche.fleche def shadowed(...)` (standard) | ✓ | OK (by reference) |

The standard `@fleche.fleche` form dodges it because the decorated name shadows the original and by-reference resolution succeeds — which is exactly why the happy-path docs examples (and the existing `test_executorlib_wrap_executor`, whose target `double` is a module-level function) never exercised this path.

## The fix

Bundle the in-flight map and its lock into a small picklable helper, `_InFlight`, that reconstructs fresh on unpickle:

```python
class _InFlight:
    __slots__ = ("map", "lock")
    def __init__(self):
        self.map = {}
        self.lock = threading.Lock()
    def __reduce__(self):
        return (_InFlight, ())   # fresh map + lock; both are process-local
```

In-flight state and the lock are process-local, so a worker that receives the wrapper by value correctly starts with an empty map and its own lock. Concurrency behaviour is unchanged (the peek at line 219 stays lock-free; set/pop stay under the lock). The lock was the *only* by-value blocker — every other closure cell (`func`, `policy`, `meta`, `isolate`, `get_call`) already pickles — so the payload now round-trips:

```
cloudpickle by-value: OK, 13797 bytes
restored(): 9
```

## Docs

A light `.. tip::` in `docs/parallel_execution.rst` nudges users, for process/cluster backends, to decorate a module-level function (wrapper ships by reference) and notes that rebinding to a different name — or `__main__`/notebook defs — forces heavier by-value serialisation that only cloudpickling backends support.

## Tests

`tests/regression/test_wrap_executor_cloudpickle_lock.py` now guards the fix:
- **deterministic, always-runs:** the wrapper closure carries no raw `threading.Lock` (version-independent — never invokes cloudpickle, so it holds even on the `test-minimum-deps` env whose older cloudpickle can't walk the wrapper bytecode by value).
- **end-to-end:** a cloudpickle by-value round-trip via the rebinding pattern, asserting it deserialises and returns `9`; skipped only when cloudpickle can't serialise the wrapper for a reason unrelated to the lock, while a re-emergence of the lock `TypeError` still fails loudly.

```
$ python -m pytest tests/regression/test_wrap_executor_cloudpickle_lock.py -v
test_wrap_executor_payload_has_no_raw_lock_in_closure PASSED
test_wrap_executor_payload_cloudpickles_by_value      PASSED
```

Existing wrapper / futures / bound-wrapper / parallel-execution suites pass unchanged.

> Earlier commits in this PR built the reproducer first (and a note there wrongly dismissed the `@wraps` mechanism — it's real; the missing piece was that by-value is triggered by the *rebinding* / non-importable-by-reference case, not by `@wraps` alone).

### DIFF
--- a/src/fleche/wrapper.py
+++ b/src/fleche/wrapper.py
@@ -168,14 +168,41 @@ def make_rerun(func, wrapper):
     return _rerun_func
 
 
+class _InFlight:
+    """Futures whose done-callback is currently executing the cache write.
+
+    ``map`` is populated at the start of ``_cache()`` and removed in its
+    finally clause, so an entry exists only during the narrow window between
+    ``Future.set_result()`` releasing the condition lock and ``cache.save()``
+    completing; ``lock`` guards mutations of ``map``.
+
+    A raw :class:`threading.Lock` is unpicklable, and it lives in the
+    ``make_wrapper`` closure that fleche's wrapper carries.  When that wrapper
+    is cloudpickled **by value** — the shape a cross-process / cluster backend
+    needs whenever the decorated function is not importable by reference (bound
+    to a name other than its ``__qualname__``, or defined in ``__main__`` / a
+    notebook) — the lock would abort serialisation with
+    ``cannot pickle '_thread.lock' object``.  Bundling map and lock here lets
+    the object pickle as a fresh, empty instance: in-flight state and the lock
+    are process-local, so a worker that receives the wrapper by value correctly
+    starts with an empty map and its own lock.
+    """
+
+    __slots__ = ("map", "lock")
+
+    def __init__(self):
+        self.map: dict[str, Future] = {}
+        self.lock = threading.Lock()
+
+    def __reduce__(self):
+        # Reconstruct fresh on unpickle; never ship the (unpicklable) lock or
+        # the process-local in-flight map.
+        return (_InFlight, ())
+
+
 def make_wrapper(func, policy, meta, isolate, get_call):
     """Build the cached wrapper returned by :func:`fleche`."""
-    # Tracks futures whose done-callback is currently executing the cache write.
-    # Populated at the start of _cache() and removed in its finally clause, so
-    # the entry only exists during the narrow window between Future.set_result()
-    # releasing the condition lock and cache.save() completing.
-    _in_flight: dict[str, Future] = {}
-    _in_flight_lock = threading.Lock()
+    _in_flight = _InFlight()
 
     @wraps(func)
     def wrapper(*args: Any, **kwargs: Any) -> _T:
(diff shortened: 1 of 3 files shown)


src/fleche/wrapper.py
```diff
@@ -168,14 +168,41 @@ def _rerun_func(*args, **kwargs):
     return _rerun_func
 
 
+class _InFlight:
```
pmrv: try with our picklable lock infra instead.

pmrv: Done in 01b5dce — dropped the bespoke `_InFlight` helper and swapped the raw `threading.Lock` in `make_wrapper`'s closure for `storage.thread_safe._PicklableLock`. `_in_flight` is back to a plain dict; concurrency behaviour is unchanged. By-value round-trip still verified (closure cell is now `_PicklableLock`, no raw `_thread.lock`).

---

pmrv (commented):
lgtm merge

PR #746, opened 2026-07-13
feat(storage): support fixed prefix-length multi-bagging for HDF5 backend

pmrv:
## Summary
- Adds an optional `prefix_length` to `BagOfHoldingH5FileBackend`. When set, keys sharing the first `prefix_length` characters are multiplexed as sibling groups (named by the full key) into one HDF5 file at `root/{prefix}.h5`, instead of each key getting its own file.
- Single level, fixed prefix length only — no adaptive splitting.
- Relies entirely on bagofholding's existing `file.h5/group` generalized path addressing (bagofholding 0.1.12); no new bagofholding API needed.
- Single-file mode (`prefix_length=None`, the default) is unchanged.

Closes #556

## Test plan
- [x] New unit tests for put/get/list/evict/contains/rebag in multi-bag mode
- [x] `h5_multi` fixture added to cross-backend contract test sweep
- [x] Config round-trip tests for `prefix_length`
- [x] Full test suite passes (1582 passed, 11 skipped)

Generated with [Claude Code](https://claude.ai/code)

Issue #625, opened 2026-06-04
Perf audit: 2026-09-17


pmrv:
**Verdict**: No new source-caused regressions — the only rows crossing the threshold are the chronic `BagOfHoldingH5File` `contains_hit`/`evict` cost (3.8–6.2× vs HEAD, same root cause as every prior audit, still unfixed). This is also the first audit where `perf record` actually worked in-container (prior runs had no matching kernel package); profiling three synthetic hot-path repros below confirms each of the 3 standing hot spots mechanistically rather than by source-reading alone.

---

## Regressions (≥1.5× and ≥1 µs, or ≥3×) — 2026-09-17 run vs HEAD (`benchmarks/results.csv` unchanged since `cc07be9`)

| Topic | Config | Workload | Function | HEAD | Fresh | Ratio |
|---|---|---|---|---|---|---|
| Value Storage | BagOfHoldingH5File | small_strings | evict | 79 µs | 490 µs | **6.20×** |
| Value Storage | BagOfHoldingH5File | numpy_arrays | evict | 85 µs | 520 µs | **6.12×** |
| Value Storage | BagOfHoldingH5File | small_strings | contains_hit | 28 µs | 160 µs | **5.71×** |
| Value Storage | BagOfHoldingH5File | nested_structures | contains_hit | 28 µs | 150 µs | **5.36×** |
| Value Storage | BagOfHoldingH5File | numpy_arrays | contains_hit | 29 µs | 150 µs | **5.17×** |
| Value Storage | BagOfHoldingH5File | nested_structures | evict | 53 µs | 200 µs | **3.77×** |

No baseline rows missing from the fresh run; no new rows appeared (240/240 both runs). Re-ran `benchmark_storage.py` standalone a second time — the fresh-side numbers reproduced within ~10%, so this isn't a one-off scheduling blip; it's the same chronic per-op HDF5-open cost flagged in every audit since 2026-05-14 (hot-spot #2), just landing above threshold this run because HEAD happened to sample low.

---

## Per-topic slowest workload/function and backend gaps

- **Digest**: slowest is `List (integers, len>100)` at **17 ms**, ~23× `len<100`'s 730 µs for only a few × more elements (hypothesis-sized, not fixed length) — same shape as every prior audit (hot-spot #1).
[...]

PR #786, opened 2026-07-23
fix(storage): cache open HDF5 read handles for multi-bag bag files

pmrv:
Fixes the `BagOfHoldingH5File` multi-bag regression flagged in #625: since PR #746, every `contains()` (and each bag in `list()`) paid a full HDF5 file open/close, ~8.6× slower than the per-key layout (240 µs vs 28 µs).

## What changed

`bagofholding_file.py` gains a process-wide cache of open **read-only** `h5py.File` handles (`_BagHandleCache`), keyed by absolute bag-file path:

- A `WeakValueDictionary` indexes the open handles; a bounded strong-ref MRU (`_MAX_OPEN_BAGS = 32`) keeps the most recently used ones alive between operations. A handle evicted from the MRU is never closed eagerly — the strong reference is dropped and the interpreter closes the file once the last in-flight user releases it.
- Every acquisition re-validates the handle against an `(inode, mtime_ns, size)` stat signature, so bags rewritten by other processes or other storage instances are detected and reopened. One stat (~1 µs) replaces one HDF5 open (~120 µs).
- Cached handles are opened with `locking=False`: a long-lived reader holding HDF5's shared OS lock would otherwise make every write from *another process* fail for as long as it lives. Cross-process write mutual exclusion was always provided by the `filelock` sidecar locks, not by HDF5's locking, so nothing is lost.
- Within a process, HDF5 refuses any open whose mode **or locking flags** differ from a live handle's. So every path that opens the bag file itself — `put`/`_to_file`, `get`/`_from_file`, `_evict`, `rebag` (H5Bag uses default flags) — runs inside `_bag_writer(key)`: it closes the cached handle and holds a per-bag in-process lock across the operation. Readers (`_contains`, `list`) hold the same lock while using the handle, so a writer can never close a handle mid-read.
- The cache is process-wide rather than literally per-instance for the same reason: a handle cached by one storage instance must be closable by any *other* instance (equal or not) addressing the same file, or that instance's writes would fail. The per-instance `WeakKeyDictionary` pattern used for lock tables would leave two non-equal instances on the same root able to brick each other's writes.
- Fork-safety: the cache detects a changed pid and drops all inherited state (read-only handles close harmlessly in the child); `locking=False` falls back to a default-flag open on h5py < 3.5 / HDF5 < 1.12.1.

## Numbers

| op | before | after | pre-regression (#625 baseline) |
|---|---|---|---|
| `contains` (hit) | 240 µs | **29 µs** | 28 µs |
| `contains` (miss) | ~240 µs | **17 µs** | — |
| `evict` | 490 µs | unchanged | 79 µs (per-key layout) |

`evict` is deliberately untouched: it genuinely mutates the bag, so it must pay the `filelock` acquisition (~100 µs) plus an HDF5 write open/del/close (~130 µs+). That cost is structural for a durable per-key delete inside a shared file; caching write handles across operations would hold HDF5's exclusive state hostage against every other process. If evict throughput ever matters, the realistic lever is batching (one write open per bag for N evictions, e.g. a `evict_many` on the storage or transaction-style deferral), not handle caching.

## `list()` — remaining cost and options

`list()` now reuses cached handles, so a warm repeated `list()` costs ~0 opens for up to `_MAX_OPEN_BAGS` bags. But a cold `list()` over a root with many bags still opens every file once, and roots with more than 32 bags will churn the MRU. Options, roughly in order of effort:

1. **Leave it** (this PR). Warm-path wins only; cold `list()` on a default `prefix_length=2` root is up to 256 opens (~30 ms worst case).
2. **Cache each bag's key list next to its handle**, invalidated by the same stat signature. Cheap to add on top of this PR's infrastructure; makes repeated `list()` on an unchanged root free even beyond the MRU bound (key lists survive handle eviction). Doesn't help the truly cold first call.
3. **A persistent per-root key index** (sidecar manifest updated under the bag's filelock on put/evict). Makes cold `list()` one file read, but adds cost and a consistency obligation to every write, plus a repair story for crashes between bag write and index write. This is the only option that fixes the cold path, and the only one with real correctness surface.
4. **`prefix_length=0`** already gives O(readdir) `list()` today for workloads that are list-heavy and don't hit the many-small-files problem multi-bagging exists to solve — worth remembering before building 3.

Happy to follow up with 2 (small) or 3 (needs design agreement) in a separate PR.

## Behavioral notes

- Code that opens a bag file directly with `h5py.File(path)` (default flags) in the same process while a cached handle is live will now get the flag-mismatch `OSError`; go through the storage API, or open with `locking=False`.
- The stat signature uses `st_mtime_ns`: on filesystems with coarse mtime granularity, an external rewrite that changes neither size nor inode within the same timestamp tick could go briefly unnoticed by `contains`. Same-process writes are exact (explicit invalidation); cross-process reads were already best-effort.
- Cached handles keep files open between operations; on Windows that makes deleting a bag file out from under a warm storage fail (`_evict` itself closes the handle under the bag lock before unlinking, so the backend's own deletes are safe; CI is Linux-only).

## Testing

- 8 new tests pin: handle reuse across `contains`/`list` (open counting), invalidation on sibling-key `put`, `get` after a warm `contains` (the flag-mismatch trap), evict with a warm cache including file removal, writes from a second non-equal instance on the same root, external-process writes being neither blocked nor masked (subprocess test — also pins `locking=False`), and the MRU bound.
- Full suite: 1640 passed, 11 skipped. `ty check src/` clean.

https://claude.ai/code/session_014DWRGyuEoH3qXNSQbSXddp

---
_Generated by [Claude Code](https://claude.ai/code/session_014DWRGyuEoH3qXNSQbSXddp)_


src/fleche/storage/bagofholding_file.py
```diff
@@ -56,6 +61,137 @@ def _observed_prefix_lengths(root: Path) -> set[int]:
     return observed
 
 
+# Cap on read-only bag-file handles kept open between operations.  Handles are
+# shared process-wide (keyed by absolute path), so this also bounds the
+# process's open-fd contribution regardless of how many storages exist.
+_MAX_OPEN_BAGS = 32
```
pmrv: let's try 8

src/fleche/storage/bagofholding_file.py
```diff
@@ -56,6 +61,137 @@ def _observed_prefix_lengths(root: Path) -> set[int]:
...
+                f = None
+        if f is None:
+            f = _open_readonly(path)
+        with self._meta_lock:
+            self._files[key] = f
+            self._signatures[key] = signature
+            self._recent[key] = f
+            self._recent.move_to_end(key)
+            while len(self._recent) > _MAX_OPEN_BAGS:
+                self._recent.popitem(last=False)
```
pmrv, line 183: does this not expose a race condition between both acquirations of meta_lock?

src/fleche/storage/bagofholding_file.py
```diff
@@ -205,26 +341,65 @@ def consolidate(
...
+            try:
+                yield _bag_handles.acquire(file_path)
+            except OSError:
+                _bag_handles.invalidate(file_path)
+                raise
+
+    @contextlib.contextmanager
+    def _bag_writer(self, key: str):
+        """Hold *key*'s in-process bag lock across an operation that opens the
+        bag file itself, closing any cached read handle first.  HDF5 refuses a
```
pmrv, line 376: I think we can make our lives easier by just opening all files in read-write mode, so that cached handles can just be given out to any operation that requests them

src/fleche/storage/bagofholding_file.py
```diff
@@ -291,8 +467,9 @@ def list(self) -> Iterable[Digest]:
             ):
                 continue
             try:
-                with h5py.File(p, "r") as f:
-                    keys.extend(Digest(name) for name in f.keys())
+                with self._bag_reader(p) as f:
+                    if f is not None:
+                        keys.extend(Digest(name) for name in f.keys())
             except OSError as e:
```
pmrv: f = None only on OSError, in which case this operation here would be skipped, rather than repeated as advised in by bagreaders docstring.

pmrv (commented):
I see, because bag of holding doesn't expose the file handles nor let's us inject them, your 'caching' caches raw h5py handles only to peek inside the bags without bag of holding and invalidates on any writing operation, then actually using the bag of holding API.  I don't like this.  

Sketch what API changes we'd need from bag of holding to be able to just hold onto bag instances we might reuse.

src/fleche/storage/bagofholding_file.py
pmrv: Done in 4243748.

---

src/fleche/storage/bagofholding_file.py
pmrv: No — the invariant that closes it is that every caller of `acquire()` holds the per-bag lock from `lock(path)` for the whole call (readers additionally for the duration of their use of the handle). So between the two `_meta_lock` sections no other thread can acquire *or* invalidate **this** path; the only interleaving the dropped `_meta_lock` admits is activity on *other* paths, and that never touches this path's entries — MRU eviction only drops other paths' strong references, it doesn't close anything. `_meta_lock` exists solely to keep the shared dicts internally consistent across different bags; per-path mutual exclusion comes from the bag lock.

4243748 adds a comment in `acquire()` stating this. The alternative — holding `_meta_lock` across the `h5py` open — would also be correct and simpler to reason about, at the cost of serializing concurrent cold opens of *different* bags (~120 µs each); happy to switch if you prefer the simpler model.

---

src/fleche/storage/bagofholding_file.py
pmrv: Right, good catch — fixed in 4243748. Reads now go through `_read_bag(file_path, reader)`, which on a mid-read error drops the broken handle and repeats the read once from a fresh open; only a failure of the *fresh* open still propagates (the file really is unreadable, matching the old per-open semantics). While testing this I also found that a closed h5py handle silently answers `False` to `key in f` instead of raising, so `acquire()` now additionally validates cached handles via h5py truthiness (`not f`) and reopens dead ones.

---

src/fleche/storage/bagofholding_file.py
pmrv: I probed this (h5py 3.16 / HDF5 2.0) and it splits cleanly along the process boundary:

**In-process it works exactly as you say.** A cached default-flag `r+` handle coexists with H5Bag's own `r` and `a` opens (verified — HDF5 shares the underlying file object when flags are compatible, which is also why the views stay coherent). All the invalidate-before-write and flag juggling would disappear, and `evict`/future batched ops could even go through the cached handle directly.

**Cross-process it's a trap.** A default-locking `r+` handle holds HDF5's *exclusive* lock for as long as it's cached — I verified that another process then fails **any** open, reads included, with `BlockingIOError`, until we happen to drop the handle (which for a hot bag is never). Today no handle outlives a single operation, so this would turn "two jobs sharing a cache root" from working into deadlocked-until-timeout. And the escape hatch of `r+` with `locking=False` just reintroduces the same flag-mismatch failure with H5Bag's default-flag opens that the current invalidation dance exists to avoid — verified as well.

So RW handles become a real simplification only once *every* open goes through us, i.e. once bagofholding can accept an injected handle/group — then one `r+`, `locking=False` handle per bag serves everything and cross-process coordination stays with the filelock sidecars, exactly as now. That's the sketch in the main thread. Alternatively, if you're happy declaring multi-bag roots single-process-at-a-time, I can switch to cached default-locking `r+` handles now and the code does get meaningfully smaller — your call.

---

pmrv:
Fair characterization — the cache peeks at raw h5py handles for `contains`/`list` and has to step aside (invalidate + flag-juggle) every time the bagofholding API opens the file itself. Here's the sketch of what bagofholding would need for fleche to hold onto reusable bag state instead, in two sizes:

## Option A — minimal: handle injection (two changes)

1. **Construct from an open group.** `H5Bag(source, ...)` accepts a live `h5py.Group`/`h5py.File` in addition to a path (or a dedicated `H5Bag.from_group(group)`). When given a handle, the constructor and `load()` read from it and never close what they didn't open. The `file.h5/subpath` addressing 0.1.12 already does internally maps 1:1 to "a group inside an open file", so this is mostly threading a parameter through.
2. **Save into an open group.** `H5Bag.save(value, target)` likewise accepts an open writable `h5py.Group` (caller does `f.require_group(key)`), writing in place and flushing rather than opening/closing the path itself.

That's sufficient. fleche then keeps exactly **one handle per bag file**, opened `"r+"` with `locking=False`, and routes *every* operation through it:

```python
f = cache.acquire(bag_path)              # the existing MRU + stat-signature machinery
key in f                                 # contains
f.keys()                                 # list
H5Bag(f[key]).load(...)                  # get
H5Bag.save(value, f.require_group(key))  # put   (+ f.flush())
del f[key]                               # evict (+ f.flush())
```

Since nothing ever double-opens the file, the whole invalidate-on-write/flag-mismatch dance disappears (`_bag_writer` and the H5Bag-vs-cache mode split are deleted); cross-process coordination is unchanged — `filelock` sidecars for writers, stat-signature reopen for staleness. Fallback to `"r"` on read-only filesystems.

## Option B — nicer: a first-class multi-bag container

Promote the multi-bag file to an object that owns the open handle:

```python
bags = bagofholding.BagFile(path, mode="r+", locking=False)   # long-lived
key in bags;  bags.keys()
bags.load(key, version_validator=...)
bags.save(value, key)
del bags[key]
bags.flush();  bags.close()
```

fleche's cache then holds `BagFile` instances directly — the `_BagHandleCache` machinery (per-path lock, MRU, stat signature, fork guard) carries over unchanged with `BagFile` in place of `h5py.File` — and `bagofholding_file.py` collapses to path layout + locking bookkeeping. This also gives bagofholding a natural home for future batched ops (`save_many`, `evict_many` = one flush), which is what an `evict` faster than ~250 µs would need anyway.

Supporting needs under either option: pass-through of open kwargs (at least `locking=`), an explicit `flush()`, and documented single-writer expectations (bagofholding shouldn't need to know about fleche's filelocks, it just must not assume it owns the file exclusively).

Option A is small enough to be one upstream PR and I'd start there; B can grow out of it. Happy to draft the bagofholding side if useful.

Status of the review otherwise: cap is 8, reads retry once from a fresh open instead of skipping (plus closed-handle detection), and the `_meta_lock` question is answered on its thread — all in 4243748.

---

pmrv:
I'd like to see a sketch of an option C as well, where bagofholding does everything transparently and keeps a file open behind a timeout for a while in case of repeated operations.

pmrv:
## Option C — transparent handle pool inside bagofholding

The public API stays path-based (`H5Bag(path).load()`, `H5Bag.save(value, path)`); internally every `h5py.File` open routes through a process-wide pool that keeps handles open for a while after use:

```python
# bagofholding internals
_pool = FilePool(max_open=8, idle_timeout=5.0)  # both configurable

class FilePool:
    def open(self, path, *, write=False) -> ContextManager[h5py.File]: ...
    # checkout: refcount += 1; checkin: refcount -= 1, stamp idle deadline
    def flush_all(self): ...   # close every idle handle now (tests, Windows, rsync)

# every internal open becomes
with _pool.open(path, write=True) as f:
    ...
    f.flush()   # writes flush at checkin — durability must not wait for the timeout
```

Key internals (all four are things this PR already had to solve, so they transplant directly):

- **One handle per absolute path, opened `r+` with `locking=False`**, falling back to `r` on read-only files. Because *all* opens go through the pool, the same-process mode/flag conflicts that force the invalidation dance in this PR vanish by construction — there is only ever the one handle. Cross-process exclusion stays with the caller (fleche's sidecar filelocks); the pool must not add OS-level locking of its own.
- **Stat-signature revalidation at checkout** (`inode, mtime_ns, size`) so files rewritten by other processes are reopened, and a fork guard (pid check) that abandons inherited handles.
- **Lifetime = refcount + idle deadline + LRU cap.** A checked-in handle isn't closed; it gets `deadline = now + idle_timeout`. Reaping is the one genuinely new design decision:
  - *(a) lazy* — every pool operation closes expired handles in passing. No background thread, but a handle can linger past its deadline in a quiet process (harmless on Linux; visible as held fds/locks on Windows). Needs the explicit `flush_all()` escape hatch.
  - *(b) daemon timer thread* — prompt reaping, but bagofholding grows a background thread with fork- and interpreter-shutdown care.
  - I'd start with (a) + `flush_all()` + an `atexit`/`weakref.finalize` sweep; (b) can be added behind the same interface.
- **Per-path checkout lock** (readers share, writers exclusive) — the direct analogue of this PR's per-bag lock, just living next to the handle it guards.

**What C does *not* remove:** fleche's `contains`/`list`/`evict` still need *something* to ask "is this group in the file / which groups / delete this group" — today that's the raw-h5py peeking you objected to. So C still wants a small query surface, which is exactly Option A's injected-handle constructor or Option B's `BagFile` container. In other words, **C is less an alternative than a placement decision**: the pool slots naturally *underneath* B (`BagFile(path)` checks out of the pool transparently; `contains` becomes `key in BagFile(path)` with zero opens when warm).

**Trade-offs vs A/B:**

| | A (inject handle) | B (`BagFile` container) | C (transparent pool) |
|---|---|---|---|
| upstream effort | smallest | moderate | largest (pool, reaping, config, docs) |
| fleche-side cache code | keeps `_BagHandleCache` | machinery moves onto `BagFile` | **deleted entirely** |
| who benefits | fleche | fleche + explicit users | **every bagofholding user** |
| handle lifetime | explicit (caller) | explicit (object) | implicit (timeout) — most surprising: fds stay open, files "busy" on Windows, after a bare `H5Bag.save()`; needs opt-out (`FilePool(idle_timeout=0)` ≡ today's behavior) |
| eviction/batching path | caller's problem | natural (`del bags[key]`, future `save_many`) | natural once B's surface exists |

My recommendation stands as a sequence rather than a choice: land A's injected-handle constructor (small, unblocks fleche dropping the raw-h5py peeking), grow B's `BagFile` on top of it, then move this PR's cache logic upstream *as* C's pool under `BagFile` — at which point `bagofholding_file.py` here reduces to path layout + filelocks, and the fleche-side cache is deleted rather than maintained in parallel. Happy to draft the bagofholding side starting from whichever rung you want.

---

pmrv:
Will let this sit for a bit and discuss with liam.   300ums per open hurt my pride, but are also not life threatening.

PR #793, opened 2026-07-26
feat(caches)!: seal call identity before the function body runs

pmrv:
## Problem

The wrapper digests arguments twice from live state: once before the function body runs (the lookup key, `wrapper.py`) and once after (when `Cache.save` re-stashed the call via `Call.stash`). A function that mutates an argument — appends to a received list, writes an output file into a directory it received — was therefore recorded under the **post-mutation** content:

- honest repeat calls with the original input could never hit (the function silently re-executes forever), and
- a call passing the *mutated* state would **false-hit** a result computed from different input.

## Design

fleche commits pure functions: the recorded identity of a call is the arguments **as passed**. This lands as a two-phase save protocol with a hybrid interface — the entry point on the cache, the lifecycle state in `call.py`:

- **`cache.prepare(call) -> PreparedCall`** — stores argument values and seals the lookup key *before* the body runs. Because `digest(x) == values.save(x)`, the sealed key always equals `to_lookup_key()`; lookup and record cannot drift. The `BaseCache` implementation is a digest-only admission (nothing written); `Cache` overrides it to stash into its value storage, wrappers delegate to their inner cache, and `CacheStack` prepares directly on `stack[0]`, where its saves land.
- **`PreparedCall.commit(result, metadata)`** — single-shot: hands the `PreparedCall` back to the cache's `save`, which stores the result *as returned* (a mutated argument passed back out is captured in its final state) and files the record. A second `commit`, or a `commit` after `abandon`, raises `RuntimeError`.
- **`PreparedCall.abandon()`** — releases without recording (body raised, or result uncacheable). Idempotent no-op by default; stored argument values are content-addressed orphans for GC (#826 tracks the gc-vs-in-flight race). `PreparedCall` is also a context manager that abandons when the block exits uncommitted (sync convenience; the wrapper's `Future` path calls the methods directly).

`Call → PreparedCall → DigestedCall` is one state progression in one module, and `DigestedCall` only ever holds digests: the pending result rides privately on `PreparedCall` until the cache's `save` stores it via `PreparedCall.resolve(values)` — the mirror of `Call.stash(values)`.

## Interface changes

- **`BaseCache.save` takes `PreparedCall | Call`.** A live `Call` is the degenerate one-shot form (values static, nothing to drift), so post-hoc saving — transfers, `redigest`, external callers, existing tests — keeps working unchanged; a bare `DigestedCall` still files as-is.
- **The wrapper degrades to uncached execution when `prepare` fails**: `Rejected` logs a policy warning, any other exception logs with traceback — the body runs either way, matching the existing save-side handling.
- **Remote protocol**: a `PreparedCall` never crosses the wire. `SshCache.prepare` is one round trip (the server stashes the arguments and returns the sealed record); a commit is two — a wire-only, write-gated `save_value` RPC stores the result value (strict: a storage-refused value rejects the commit, as locally), then the plain `DigestedCall` is filed. Read-only remotes admit digest-only locally and reject the commit without a round trip. `Path` values over SSH remain unsupported. A uniform `.values` facade was considered and deliberately deferred; `_RemoteValues` is its seed.
- **Read-only caches admit digest-only**: `ReadOnlyMixin.prepare` seals the key without writing anything; the commit after the body is rejected, so the call runs and returns uncached — the same observable behavior as the previous post-body save rejection.

## Behavior change

**BREAKING**: calls of argument-mutating functions are now recorded under the pre-call argument state. Such functions start hitting on honest repeats — and their side effect on the input consequently happens on cold calls only (it is neither recorded nor replayed). Caches populated under the old keying re-execute once and re-record correctly.

## Testing

- `tests/unit/call/test_prepared_call.py`: prepare/commit/abandon lifecycle, key sealing, single-shot commit guards, context-manager semantics, read-only rejection, one-shot-save key equality, prepare-failure degradation, and end-to-end mutation coherence (mutating consumer hits; mutated state is a distinct call; body exceptions leave no record).
- `tests/unit/test_remote.py` additions: prepare/commit wire round trip (sealed key survives argument mutation; nothing `PreparedCall`-shaped crosses the wire), read-only local admission with zero RPCs, unstorable-result rejection, `None`-result round trip.
- Full suite: 1661 passed, 11 skipped. `ty check src/` clean.
- Follow-up on the `temppath` branch (stacked on this PR): path-level integration tests and the `file_semantics.rst` rewrite for the new mutation semantics.

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -48,46 +48,37 @@ class BaseCache(OperationContext):
         from . import config as _config
         return _config.cache_from_config(config)
 
-    @abstractmethod
-    def save_value(self, value: Any) -> "Digest":
-        """Store one value, returning its content digest.
-
-        The write-side counterpart of :meth:`load_value`; also the primitive
-        :meth:`prepare` runs argument and result values through.
-        """
-        ...
-
     def prepare(self, call: Call) -> PreparedCall:
-        """Admit *call* to this cache: store its arguments, seal its lookup key.
+        """Admit *call* to this cache: seal its lookup key before the body runs.
 
-        The first half of the two-phase save protocol.  Argument values go
-        through :meth:`save_value` *now* — before the function body runs — so
-        the recorded identity always describes the arguments as they were at
-        call time, even if the body later mutates them.  Because ``digest(x)
-        == save_value(x)`` for every storable value, the sealed key equals
-        ``call.to_lookup_key()``.
+        The first half of the two-phase save protocol.  Caches that own a
+        value storage (:class:`Cache`) override this to stash the argument
+        values *now* — before the function body runs — so the recorded
+        identity describes the arguments as they were at call time, even if
+        the body later mutates them.
+
+        This base implementation is the digest-only admission used by caches
+        that cannot (or must not) write ahead of the body — read-only views,
+        aggregates without their own value storage: the key is sealed, but
+        nothing is written and whether a record survives is decided by
+        :meth:`save` at commit time.  Because ``digest(x) == values.save(x)``
+        for every storable value, both forms seal the same key.
 
         Finish the returned :class:`~fleche.call.PreparedCall` with exactly
         one of :meth:`~fleche.call.PreparedCall.commit` (store the result,
         file the record) or :meth:`~fleche.call.PreparedCall.abandon`.
-
-        Argument values the storage refuses (``SaveError``) fall back to a
-        digest-only reference, as in :meth:`fleche.call.Call.stash`.
         """
-        digested = call._to_digested(self.save_value)
-        return PreparedCall(
-            call=call, digested=digested, key=digested.to_lookup_key(), cache=self
-        )
+        return PreparedCall(digested=call.digest(), cache=self)
 
     @abstractmethod
     def save(self, call: DigestedCall | Call) -> str:
         """File a call record.
 
-        The primary form takes a fully digested record whose values were
-        already stored — see :meth:`fleche.call.Call.prepare` for the
-        two-phase protocol that does both in the right order.  A live
-        :class:`Call` is also accepted as the degenerate one-shot form
-        (values static, stored on the spot)."""
+        Takes either a live :class:`Call` — values are stored on the spot, the
+        one-shot form — or a :class:`DigestedCall` whose argument values were
+        already stored by :meth:`prepare`, in which case only its result (if
+        any) still needs storing.  See :meth:`prepare` for the two-phase
+        protocol that does both in the right order."""
         ...
 
     @abstractmethod
@@ -332,37 +323,35 @@ class Cache(PerKeyLockMixin, BaseCache):
         with self._operation_context(key):
             return self.values.load(key)
 
-    def save_value(self, value: Any) -> Digest:
-        # No cache-level lock: the key is only known once the value storage
-        # has digested the value, and value storages carry their own per-key
-        # locking (PerKeyLockMixin) where they need it.
-        return self.values.save(value)
+    def prepare(self, call: Call) -> PreparedCall:
+        # Stash the arguments *now*, before the function body runs, so the
+        # record cannot end up keyed on post-mutation content.  ``stash``
+        # leaves the still-unknown result as ``None``; ``save`` below stores
+        # it once ``commit`` fills it in.  No cache-level lock: the keys are
+        # only known once the value storage has digested each value, and value
+        # storages carry their own per-key locking where they need it.
+        return PreparedCall(digested=call.stash(self.values), cache=self)
 
     def save(self, call: DigestedCall | Call) -> str:
-        # Record-only: argument and result values were already written to
-        # ``self.values`` by Call.prepare / PreparedCall.commit, whose digests
-        # this record carries.  Writing them here instead would re-read mutable
-        # values (e.g. Path contents) *after* the function body ran and file
-        # the record under post-mutation content — the incoherence the
-        # two-phase protocol exists to prevent.
-        #
-        # A live Call (values not yet stored) is still accepted as the
-        # degenerate one-shot form: with no function body between digesting
-        # and filing there is nothing to drift, so stash-then-file is
-        # equivalent to prepare/commit here.  Inlined rather than routed
-        # through prepare().commit() so one logical save does not re-enter
-        # subclass save() overrides a second time.  Wrappers and stacks need
-        # no own shim — every save path lands here.
-        if isinstance(call, Call):
-            key = call.to_lookup_key()
-            with self._operation_context(key):
-                try:
-                    digested = call.stash(self.values)
-                except storage.SaveError as e:
-                    raise Rejected(e)
-                return self.calls.save(digested)
         key = call.to_lookup_key()
         with self._operation_context(key):
+            try:
+                if isinstance(call, Call):
+                    # One-shot form: nothing was stored ahead of time.  With no
+                    # function body between digesting and filing there is
+                    # nothing to drift, so stash-then-file is equivalent to
+                    # prepare/commit here.
+                    call = call.stash(self.values)
+                elif call.result is not None and not isinstance(call.result, Digest):
+                    # Second half of the two-phase protocol: the arguments are
+                    # already digests (stored by ``prepare``), only the result
+                    # arrives live from ``PreparedCall.commit``.  Not re-saving
+                    # the arguments here is the point — reading them again
+                    # *after* the body ran would file the record under
+                    # post-mutation content.
+                    call = replace(call, result=self.values.save(call.result))
+            except storage.SaveError as e:
+                raise Rejected(e)
             return self.calls.save(call)
 
     def load(self, key: str) -> LazyCall:
@@ -530,8 +519,11 @@ class CacheWrapper(BaseCache):
 
     cache: BaseCache
 
-    def save_value(self, value: Any) -> Digest:
-        return self.cache.save_value(value)
+    def prepare(self, call: Call) -> PreparedCall:
+        # The inner cache decides how the arguments are stored; rebinding the
+        # cache makes the eventual commit go through *this* wrapper's ``save``,
+        # so wrapper policy (read-only, filtering, size limits) still applies.
+        return replace(self.cache.prepare(call), cache=self)
 
     def save(self, call: DigestedCall | Call) -> str:
         return self.cache.save(call)
@@ -589,18 +581,14 @@ class ReadOnlyMixin:
     def evict(self, key: str | Digest) -> None:
         raise Rejected("Cannot evict from a read-only cache", self, key)
 
-    def save_value(self, value: Any) -> Digest:
-        raise Rejected("Cannot save values to a read-only cache", self)
-
     def prepare(self, call: Call) -> "PreparedCall":
-        # Digest-only admission: a read-only cache stashes nothing, so the
-        # function body still runs with a correctly sealed key, and the
-        # eventual commit is rejected (save_value raises) — matching the
-        # behavior save() rejection produced before the two-phase protocol.
-        digested = call.digest()
-        return PreparedCall(
-            call=call, digested=digested, key=digested.to_lookup_key(), cache=self
-        )
+        # Digest-only admission (the BaseCache default, restated because this
+        # mixin is base-free and must beat CacheWrapper.prepare in the MRO): a
+        # read-only cache stashes nothing, so the function body still runs with
+        # a correctly sealed key and the eventual commit is rejected by
+        # ``save`` above — the behavior save() rejection produced before the
+        # two-phase protocol, minus the wasted writes.
+        return PreparedCall(digested=call.digest(), cache=self)
 
 
 @dataclass(frozen=True)
@@ -846,14 +834,16 @@ class CacheStack(PerKeyLockMixin, _MultiCache):
             if isinstance(c, CacheStack):
                 raise ValueError("CacheStack cannot be nested inside another CacheStack")
 
-    def save(self, call: DigestedCall | Call):
-        self.stack[0].save(call)
+    def save(self, call: DigestedCall | Call) -> str:
+        # Returning the key (rather than dropping it, as this used to) keeps
+        # the BaseCache contract that PreparedCall.commit hands back to callers.
+        return self.stack[0].save(call)
 
-    def save_value(self, value: Any) -> Digest:
-        # Writes always land on stack[0] (matching save); reads that need the
-        # full fan-out go through load_value, which _MultiCache spreads over
-        # every member.
-        return self.stack[0].save_value(value)
+    def prepare(self, call: Call) -> PreparedCall:
+        # Writes always land on stack[0] (matching save), so that is where the
+        # arguments are stashed; rebinding the cache routes the commit back
+        # through this stack's ``save``.
+        return replace(self.stack[0].prepare(call), cache=self)
 
     @contextlib.contextmanager
     def _operation_context(self, key, *, intent: Intent = Intent.WRITE):
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -13,7 +13,7 @@ from . import storage
 from .storage.base import _apply_shrink, _resolve_prefix, Intent, OperationContext
 from .storage.destructuring import HasChildDigests
 from .storage.thread_safe import PerKeyLockMixin, _PicklableRLock
-from .call import Call, DigestedCall, LazyCall, QueryCall
+from .call import Call, DigestedCall, LazyCall, PreparedCall, QueryCall
 from . import call
 from . import query
 
@@ -49,7 +49,45 @@ class BaseCache(OperationContext):
         return _config.cache_from_config(config)
 
     @abstractmethod
-    def save(self, call: Call) -> str:
+    def save_value(self, value: Any) -> "Digest":
+        """Store one value, returning its content digest.
+
+        The write-side counterpart of :meth:`load_value`; also the primitive
+        :meth:`prepare` runs argument and result values through.
+        """
+        ...
+
+    def prepare(self, call: Call) -> PreparedCall:
+        """Admit *call* to this cache: store its arguments, seal its lookup key.
+
+        The first half of the two-phase save protocol.  Argument values go
+        through :meth:`save_value` *now* — before the function body runs — so
+        the recorded identity always describes the arguments as they were at
+        call time, even if the body later mutates them.  Because ``digest(x)
+        == save_value(x)`` for every storable value, the sealed key equals
+        ``call.to_lookup_key()``.
+
+        Finish the returned :class:`~fleche.call.PreparedCall` with exactly
+        one of :meth:`~fleche.call.PreparedCall.commit` (store the result,
+        file the record) or :meth:`~fleche.call.PreparedCall.abandon`.
+
+        Argument values the storage refuses (``SaveError``) fall back to a
+        digest-only reference, as in :meth:`fleche.call.Call.stash`.
+        """
+        digested = call._to_digested(self.save_value)
+        return PreparedCall(
+            call=call, digested=digested, key=digested.to_lookup_key(), cache=self
+        )
+
+    @abstractmethod
+    def save(self, call: DigestedCall | Call) -> str:
+        """File a call record.
+
+        The primary form takes a fully digested record whose values were
+        already stored — see :meth:`fleche.call.Call.prepare` for the
+        two-phase protocol that does both in the right order.  A live
+        :class:`Call` is also accepted as the degenerate one-shot form
+        (values static, stored on the spot)."""
         ...
 
     @abstractmethod
@@ -294,14 +332,38 @@ class Cache(PerKeyLockMixin, BaseCache):
         with self._operation_context(key):
             return self.values.load(key)
 
-    def save(self, call: Call) -> str:
+    def save_value(self, value: Any) -> Digest:
+        # No cache-level lock: the key is only known once the value storage
+        # has digested the value, and value storages carry their own per-key
+        # locking (PerKeyLockMixin) where they need it.
+        return self.values.save(value)
+
+    def save(self, call: DigestedCall | Call) -> str:
+        # Record-only: argument and result values were already written to
+        # ``self.values`` by Call.prepare / PreparedCall.commit, whose digests
+        # this record carries.  Writing them here instead would re-read mutable
+        # values (e.g. Path contents) *after* the function body ran and file
+        # the record under post-mutation content — the incoherence the
+        # two-phase protocol exists to prevent.
+        #
+        # A live Call (values not yet stored) is still accepted as the
+        # degenerate one-shot form: with no function body between digesting
+        # and filing there is nothing to drift, so stash-then-file is
+        # equivalent to prepare/commit here.  Inlined rather than routed
+        # through prepare().commit() so one logical save does not re-enter
+        # subclass save() overrides a second time.  Wrappers and stacks need
+        # no own shim — every save path lands here.
+        if isinstance(call, Call):
+            key = call.to_lookup_key()
+            with self._operation_context(key):
+                try:
+                    digested = call.stash(self.values)
+                except storage.SaveError as e:
+                    raise Rejected(e)
+                return self.calls.save(digested)
         key = call.to_lookup_key()
         with self._operation_context(key):
-            try:
-                digested = call.stash(self.values)
-            except storage.SaveError as e:
-                raise Rejected(e)
-            return self.calls.save(digested)
+            return self.calls.save(call)
 
     def load(self, key: str) -> LazyCall:
         with self._operation_context(key):
@@ -392,8 +454,8 @@ class Cache(PerKeyLockMixin, BaseCache):
 
         This may take time depending on cache size."""
         for key in self.calls.list():
-            call = self.load(key).fetch()
-            new_key = call.to_lookup_key()
+            loaded = self.load(key).fetch()
+            new_key = loaded.to_lookup_key()
             if new_key == key:
                 continue
             # Hold the per-key locks for both the old and the new key so the
@@ -407,7 +469,7 @@ class Cache(PerKeyLockMixin, BaseCache):
             first, second = sorted((key, new_key))
             with self._operation_context(first), self._operation_context(second):
                 # instantiate values too
-                self.save(call)
+                self.save(loaded)
                 self.evict(key)
 
     def gc(self) -> set[Digest]:
@@ -468,7 +530,10 @@ class CacheWrapper(BaseCache):
 
     cache: BaseCache
 
-    def save(self, call: Call) -> str:
+    def save_value(self, value: Any) -> Digest:
+        return self.cache.save_value(value)
+
+    def save(self, call: DigestedCall | Call) -> str:
         return self.cache.save(call)
 
     def load(self, key: str) -> LazyCall:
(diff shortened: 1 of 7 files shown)


src/fleche/caches.py
```diff
@@ -468,7 +496,11 @@ class CacheWrapper(BaseCache):
 
     cache: BaseCache
 
-    def save(self, call: Call) -> str:
+    @property
+    def values(self) -> "storage.ValueStorage":
+        return self.cache.values
+
```
pmrv: let's keep this for the future, you may add a save_value method for now

src/fleche/call.py
```diff
@@ -214,6 +214,12 @@ def _to_digested(self, save_fn: Callable[[Any], Digest]) -> "DigestedCall":
                         raise
                     logger.warning("Failed to save argument %r: falling back to digest-only reference", k)
                     arguments[k] = digest.digest(v)
+        return arguments
+
+    def _to_digested(self, save_fn: Callable[[Any], Digest]) -> "DigestedCall":
```
pmrv: since we're anyway building an 'incomplete' digestedcall, I don't want to have an extra method for digesting the arguments.  just update the existing one to handle None result.

src/fleche/caches.py
```diff
@@ -48,8 +48,23 @@ def from_config(cls, config: "dict[str, Any] | list[dict[str, Any]]") -> "BaseCa
         from . import config as _config
         return _config.cache_from_config(config)
 
+    # Every concrete cache exposes ``values``: a value-storage view with plain
+    # verbs (``values.save(v) -> Digest``, ``values.load(key)``).  Declared as
+    # an annotation only — :class:`Cache` satisfies it with a dataclass *field*
+    # of the same name, which a base-class property (a data descriptor) would
+    # shadow at runtime.  ``load_value`` remains as the locking convenience
+    # wrapper around it.
+    values: "storage.ValueStorage"
```
pmrv: actually, I think I'd like a 'hybrid' interface more.  instead of having all the new methods on call, let the cache also grow some and have these inject the correct value storage to the preparedcall counterpart, so we don't need to expose the values attribute.  this also means readonlycache can override and just not stash the arguments

pmrv (commented):
revise.

src/fleche/caches.py
pmrv: Done — the `values` facade is withdrawn (kept in view for the future) and replaced by a `save_value` method, the write-side counterpart of `load_value`: implemented on `Cache`, forwarded by wrappers, routed to `stack[0]` by `CacheStack`, rejected by `ReadOnlyMixin`, sent over the wire by `SshCache`.

src/fleche/call.py
pmrv: Done — `_digest_arguments` is folded back into `_to_digested`, which now leaves a `None` result as `None` (the incomplete record awaiting its result) instead of running it through `save_fn`. `prepare` just calls `_to_digested(self.save_value)`.

src/fleche/caches.py
pmrv: Done — `prepare` moved onto the cache: implemented once on `BaseCache` over `save_value`, so with `self` being a wrapper/stack the right storage and save policy are injected automatically, and `Call` grows no new public methods. `ReadOnlyMixin.prepare` overrides to a digest-only admission: nothing is stashed, the key is still sealed correctly, and the commit after the body is rejected — same observable behavior as the old post-body save rejection, but with zero writes. This also deleted the wrapper's `prepared = None` special-casing; net −43 lines.

src/fleche/caches.py
```diff
@@ -49,7 +49,45 @@ def from_config(cls, config: "dict[str, Any] | list[dict[str, Any]]") -> "BaseCa
...
+        ``call.to_lookup_key()``.
+
+        Finish the returned :class:`~fleche.call.PreparedCall` with exactly
+        one of :meth:`~fleche.call.PreparedCall.commit` (store the result,
+        file the record) or :meth:`~fleche.call.PreparedCall.abandon`.
+
+        Argument values the storage refuses (``SaveError``) fall back to a
+        digest-only reference, as in :meth:`fleche.call.Call.stash`.
+        """
+        digested = call._to_digested(self.save_value)
```
pmrv: break's api

pmrv: Just drop the generic implementation here and pass (as instructed before) the entire value storage.  Might even just re use .stash itself after you make sure it can deal with None result.

src/fleche/call.py
```diff
@@ -324,6 +329,80 @@ def fetch(self, cache) -> "LazyCall":
...
+
+    .. code-block:: python
+
+        with cache.prepare(call) as prepared:
+            prepared.commit(func(...))       # skipped on exception -> abandoned
+    """
+
+    call: Call
+    digested: DigestedCall
+    key: Digest
```
pmrv: unnecessary, since you have *both* call and digested around that also know the key.

src/fleche/call.py
```diff
@@ -324,6 +329,80 @@ def fetch(self, cache) -> "LazyCall":
...
+    Also usable as a context manager in synchronous code: leaving the block
+    without having committed abandons the call.
+
+    .. code-block:: python
+
+        with cache.prepare(call) as prepared:
+            prepared.commit(func(...))       # skipped on exception -> abandoned
+    """
+
+    call: Call
```
pmrv: not used afaict.

src/fleche/call.py
```diff
@@ -324,6 +329,80 @@ def fetch(self, cache) -> "LazyCall":
...
+        Returns:
+            Digest: the record key (equal to :attr:`key`).
+
+        Raises:
+            fleche.caches.Rejected: if the result cannot be stored or the cache
+                refuses the record.
+        """
+        from .storage.base import SaveError  # lazy import: storage.base depends on call
+        try:
+            self.digested.result = self.cache.save_value(result)
```
pmrv: let this be done by Cache.save at the end of the method here, and you can drop save_value.

pmrv (commented):
revise

pmrv:
@claude [opus] ^

pmrv:
> SSH: save_value (one round trip per value) is replaced by a single prepare RPC — the remote stashes the arguments and returns the sealed DigestedCall; the client completes with the existing save. Read-only remotes short-circuit to the local digest-only admission instead of raising at prepare time, so the body still runs and the rejection lands at commit as before.

I think we just stick to the default impl of prepare, at least calls with paths would get the correct content hash, but it would still break on shipping the whole thing over with `save()`.

src/fleche/call.py
```diff
@@ -261,7 +266,11 @@ class DigestedCall:
 
     name: str
     arguments: dict[str, "digest.Digest"]
-    result: "digest.Digest | None" = None
+    # Normally a Digest pointer, or None while the result is still unknown
+    # (a record prepared before its function body ran).  Between
+    # :meth:`PreparedCall.commit` and the cache's ``save`` it briefly holds the
+    # live return value, which ``save`` stores and replaces with its Digest.
+    result: "digest.Digest | Any" = None
```
pmrv: This is the center of the "result parked on `DigestedCall`" wart. Three concrete costs of smuggling the live value through this field:

1. `Digest | Any` is just `Any` to a type checker, so the class's core invariant ("arguments and result are Digest pointers") is now unenforceable and contradicted by its own comment.
2. `Cache.save` has to sniff the state back out with `result is not None and not isinstance(result, Digest)`, which misfires when a function legitimately *returns* a `Digest` (see comment there — it files a dangling record).
3. `commit` mutates the shared `digested` in place and leaves the live result parked on it after save (see comment on `commit`).

A minimal fix that keeps `save` as the single entry point (i.e. doesn't resurrect `save_value`) is to make the transient state an explicit type instead of a smuggled `Any`:

```python
@dataclass(frozen=True)
class PendingResult:
    """A result captured at commit time, not yet in value storage."""
    value: Any
```

- `PreparedCall.commit`: `digested = replace(self.digested, result=PendingResult(result), metadata=...)` — no in-place mutation, nothing parked afterwards.
- `Cache.save`: `elif isinstance(call.result, PendingResult): call = replace(call, result=self.values.save(call.result.value))` — exact type test, no sniffing.
- This field: `result: Digest | PendingResult | None = None`, and the four-line comment shrinks to one: `# PendingResult only between PreparedCall.commit and the cache's save.`

It also fixes the Digest-result bug for free, since a Digest return value arrives wrapped.

---

src/fleche/caches.py
```diff
@@ -294,14 +323,36 @@ def load_value(self, key):
...
+                    # prepare/commit here.
+                    call = call.stash(self.values)
+                elif call.result is not None and not isinstance(call.result, Digest):
+                    # Second half of the two-phase protocol: the arguments are
+                    # already digests (stored by ``prepare``), only the result
+                    # arrives live from ``PreparedCall.commit``.  Not re-saving
+                    # the arguments here is the point — reading them again
+                    # *after* the body ran would file the record under
+                    # post-mutation content.
+                    call = replace(call, result=self.values.save(call.result))
```
pmrv: The `not isinstance(call.result, Digest)` sniff conflates "already-stored pointer" with "user data that happens to be a `Digest`". A function that returns a `Digest` now files a dangling record — verified against this branch:

```python
cache = Cache(ValueMemory({}), CallMemory({}))
p = cache.prepare(Call(name="f", arguments={"x": 1}, module="m", version="1"))
key = p.commit(digest([1, 2, 3]))   # result IS a Digest; [1,2,3] never stored
cache.load(key).result              # KeyError
```

The one-shot path (`save(Call)`) runs the same value through `values.save` and it loads back, so the two save forms diverge on this input. Returning cache keys from cached functions is not exotic usage for this library, so I'd count it a real (if narrow) regression rather than a theoretical one. The `PendingResult` marker suggested on the `DigestedCall.result` field makes this branch an exact type test and fixes it.

On the comment itself: the second sentence ("Not re-saving the arguments here is the point — …") is the one worth keeping; the first sentence restates the protocol already documented on `prepare`. Suggest:

```python
# Result committed live by PreparedCall.commit; the arguments must NOT be
# re-saved here — reading them after the body ran is exactly the
# post-mutation keying this protocol exists to prevent.
```

---

src/fleche/caches.py
```diff
@@ -294,14 +323,36 @@ def load_value(self, key):
...
 
-    def save(self, call: Call) -> str:
+    def prepare(self, call: Call) -> PreparedCall:
+        # Stash the arguments *now*, before the function body runs, so the
+        # record cannot end up keyed on post-mutation content.  ``stash``
+        # leaves the still-unknown result as ``None``; ``save`` below stores
+        # it once ``commit`` fills it in.  No cache-level lock: the keys are
+        # only known once the value storage has digested each value, and value
+        # storages carry their own per-key locking where they need it.
+        return PreparedCall(digested=call.stash(self.values), cache=self)
```
pmrv, line 326: This creates a new data-loss race with `gc()`: the stashed argument values are unreferenced by any call record for the entire duration of the function body, and `gc()` marks reachability from call records only. Verified against this branch:

```python
p = cache.prepare(Call(name="g", arguments={"xs": [1, 2]}, module="m", version="1"))
cache.gc()                      # runs while the "body" executes → evicts the stash
key = p.commit("res")
cache.load(key).arguments["xs"] # silently the bare digest string, value gone
```

(`LazyArguments` masks the `KeyError` by returning the digest, so it degrades silently.) The old race — gc between `values.save` and `calls.save` inside `save()` — was microseconds; this one is minutes-to-hours for the workloads fleche targets. The PR turns an implicit "don't gc while calls are in flight" assumption into a load-bearing one, but nothing documents it.

In ascending order of effort: (a) say it loudly on `gc()` and `prepare()`; (b) give `gc()` a grace period — skip values younger than some cutoff, the standard content-addressed-store answer (git's `--prune=<date>`), which also fixes the old narrow race; (c) pin in-flight digests somewhere `gc()` can see them. I'd do (a) in this PR and file (b) as a follow-up.

---

src/fleche/call.py
```diff
@@ -324,6 +333,74 @@ def fetch(self, cache) -> "LazyCall":
...
+
+        Raises:
+            fleche.caches.Rejected: if the result cannot be stored or the cache
+                refuses the record.
+        """
+        self.digested.result = result
+        if metadata is not None:
+            self.digested.metadata = metadata
+        self._finished = True
+        return self.cache.save(self.digested)
```
pmrv: Two things here beyond the parking issue itself:

1. After `Cache.save` does its `replace(...)`, the live result stays on `self.digested.result` for as long as the `PreparedCall` lives (in the wrapper's future path that's the closure attached to the future). With `replace` instead of in-place assignment (per the `PendingResult` suggestion) the committed record is built fresh and nothing lingers.
2. The docstring promises "exactly one of `commit` or `abandon`", but nothing enforces it — `_finished` is only consulted by `__exit__`. A second `commit`, or `commit` after `abandon`, silently files a record. A two-line guard (`if self._finished: raise RuntimeError(...)`) makes misuse of the new public protocol fail loudly instead.

---

src/fleche/wrapper.py
```diff
@@ -235,7 +235,20 @@ def _run_and_cache():
...
 
-            result: _T = func(*args, **kwargs)
+            # Two-phase save: argument values are stored and the record's key
+            # sealed *before* the body runs, so the recorded identity is the
+            # arguments as passed — even if the body mutates them (e.g. writes
+            # into a directory it received).  A read-only cache stashes
+            # nothing here (digest-only prepare) and rejects at commit time,
+            # so the call still runs and returns uncached, as it previously
+            # did when the post-body save was rejected.
+            prepared = cache.prepare(call)
```
pmrv: Robustness: `cache.prepare(call)` is the only cache interaction on this path without a fallback — lookups get `except Indigestible`, commits get `except Rejected`, but a `prepare` that raises aborts the call *before the body runs*. Today's built-in caches never raise `Rejected` from `prepare` (argument `SaveError`s fall back to digest-only inside `stash`), but `SshCache.prepare` is a network round trip, and any third-party cache that rejects at admission takes the whole call down. Mirroring the existing pattern —

```python
try:
    prepared = cache.prepare(call)
except Rejected as e:
    logger.warning("Cache rejected prepare: %s", e.args)
    return func(*args, **kwargs)
```

— preserves the library's "caching failures degrade, they don't break the call" stance at the one spot this PR adds.

On the comment: eight lines, and half is history ("as it previously did when the post-body save was rejected"). Both load-bearing points survive at half the length:

```python
# Seal the call's identity (and stash its argument values) before the body
# runs, so mutations the body makes to its arguments cannot leak into the
# recorded key.  A read-only cache seals without writing and rejects at
# commit; the call still runs and returns uncached.
```

Also: line 284's `call.metadata = metadata` is dead now that `commit` takes metadata explicitly — `m.post` only ever sees `replace(call, metadata={})`.

---

src/fleche/caches.py
```diff
@@ -768,8 +834,16 @@ def __post_init__(self):
             if isinstance(c, CacheStack):
                 raise ValueError("CacheStack cannot be nested inside another CacheStack")
 
-    def save(self, call: Call):
-        self.stack[0].save(call)
+    def save(self, call: DigestedCall | Call) -> str:
+        # Returning the key (rather than dropping it, as this used to) keeps
+        # the BaseCache contract that PreparedCall.commit hands back to callers.
+        return self.stack[0].save(call)
```
pmrv: Example of the pattern to cut throughout: "rather than dropping it, as this used to" is commit-message material, not code — the `-> str` annotation plus the `BaseCache.save` contract already say everything the next reader needs. Suggest deleting the comment entirely. (Same treatment for the tail of `ReadOnlyMixin.prepare`'s comment — "the behavior save() rejection produced before the two-phase protocol, minus the wasted writes" — while its MRO sentence is genuinely load-bearing and should stay.)

---

src/fleche/remote.py
```diff
@@ -792,6 +799,19 @@ def load(self, key: str) -> LazyCall:
...
+        # argument values before the body runs and seals the record, which
+        # comes back for the client to complete with ``save``.  Values travel
+        # by cloudpickle, so a Path argument ships its path *string*, not its
+        # content — paths over SSH remain unsupported, as before.
+        self._ensure_handshake()
+        if self._info_cache and self._info_cache.get("read_only", False):
+            # Digest-only admission, as BaseCache: the body still runs, and the
+            # commit's ``save`` is rejected locally without a round trip.
+            return super().prepare(call)
+        return PreparedCall(digested=self._rpc("prepare", call), cache=self)
```
pmrv, line 870: The routing here is sound (read-only short-circuit before dispatch, sealed record back, client finishes with plain `save`), but the only test touching it is the dispatch-coverage list entry in `test_remote.py`. Neither behavior this method adds is exercised: the read-only branch falling back to digest-only admission, nor a live `prepare` RPC followed by `commit` shipping the `DigestedCall`-with-live-result over the wire. The existing `remote`/`server_cache` fixtures make both cheap — an end-to-end mutating-argument test against the remote would pin down the one genuinely new wire shape this PR introduces (a `DigestedCall` whose `result` is a live object).

---

pmrv (commented):
Thorough review of the two-phase save protocol. Verified locally on `b8b3865`: full suite green (1652 passed, 11 skipped), `ty check src/` clean. All claims below were reproduced with probes against this branch, not just read off the diff.

## Overview

The design is right and the routing hangs together well: identity sealed at `prepare`, result stored as returned at `commit`, storage supplied by the inner cache while `save` policy stays with the outer one via the `cache` rebind (checked the MRO/rebind logic for `CacheWrapper`, `ReadOnlyMixin`, `FilteredCache`, `CachePool`, `CacheStack`, `SizeLimitedCache`, and `SshCache` — all route correctly, including read-only-bottom stacks). The test file covers the lifecycle, both wrapper routings, and the end-to-end mutation-coherence claims properly.

Two substantive problems, one design wart, and the comment style — details anchored inline.

## 1. Result parked on `DigestedCall` (design wart, with a real bug attached)

`commit` smuggles the live return value through `DigestedCall.result` (annotation widened to `Digest | Any`, i.e. `Any`), and `Cache.save` sniffs it back out with `not isinstance(call.result, Digest)`. That sniff misfires when a function legitimately returns a `Digest`: the record is filed with a result digest pointing at nothing, and `load(key).result` raises `KeyError` (probe in the inline comment; the one-shot path handles the same input, so the two save forms diverge). It also means `commit` mutates the shared `digested` in place and leaves the live result parked on it after save.

Suggested fix (inline on `call.py`): a small frozen `PendingResult` marker type. `commit` builds the record with `replace(...)`, `Cache.save` unwraps on an exact `isinstance(call.result, PendingResult)` test, and the field goes back to a checkable `Digest | PendingResult | None`. Keeps `save` as the single entry point — no `save_value` resurrection — and fixes the Digest-result bug for free.

## 2. `gc()` vs. in-flight calls — new silent data-loss race

Stashed argument values are unreachable from any call record for the whole body duration, and `gc()` marks from call records only. Probe: `prepare` → `gc()` → `commit` yields a record whose arguments load as bare digest strings (the `LazyArguments` KeyError fallback masks it, so it degrades silently). The old race window was microseconds inside `save()`; now it's the length of the function body — for fleche's workloads, potentially hours. Minimum for this PR: document it on `gc()` and `prepare()`. Better follow-up: a gc grace period for young values (the git `--prune=<date>` pattern), which also closes the pre-existing narrow race.

## 3. Comments

Agreed they need a pass: too long, and several narrate history ("as this used to", "the behavior save() rejection produced before the two-phase protocol, minus the wasted writes") that belongs to the commit message and PR description, which already carry it. The good points inside them — *why* arguments must not be re-saved at commit, *why* `ReadOnlyMixin` restates `prepare` (MRO) — are worth keeping and survive at half the length. Concrete trimmed versions suggested inline on the worst offenders (`wrapper.py`, `Cache.save`, `CacheStack.save`).

## Smaller points

- **wrapper**: `cache.prepare(call)` has no `Rejected` fallback, unlike every other cache interaction on that path — a prepare-raising cache (or a dropped SSH connection at admission) aborts the call before the body runs instead of running it uncached. Suggested guard inline.
- **wrapper.py:284**: `call.metadata = metadata` is dead now that `commit` takes metadata explicitly.
- **remote**: no behavioral test for `SshCache.prepare` — only the dispatch-coverage list entry; the new wire shape (live result on a `DigestedCall`) is untested end-to-end.
- **`PreparedCall`**: `cache: Any` could be `"BaseCache"` under `TYPE_CHECKING`; and nothing enforces the documented "exactly one of commit/abandon" (`_finished` is only read by `__exit__`).
- **PR description is stale**: it still specs `save_value` (interface-changes section, SSH "one round trip per value"), which `b8b3865` removed. Worth updating before merge since it's the design record for a breaking change.

---

src/fleche/caches.py
pmrv: let's park this in an issue for now.  Nothing gc()'s automatically, so for now the race is dormant.  Mid-term I think the solution is to register pending calls in the call storage, though that comes with it's own problems.

src/fleche/call.py
pmrv: sounds good.

src/fleche/wrapper.py
pmrv: logging warning is a go

src/fleche/remote.py
pmrv: add those tests

src/fleche/caches.py
pmrv: `Digest` should be a return value from a user function ever, but I agree the smell is there.

src/fleche/caches.py
pmrv: Filed as #826, including the reproducer, the dormant status, and the pending-records-in-call-storage direction (with its expiry/read-only open problems noted).

---

src/fleche/call.py
pmrv: Done in #825 (stacked onto this branch): `commit` builds the record with `replace` — nothing parked on `digested` afterwards — and raises `RuntimeError` on a second `commit` or a `commit` after `abandon`; `abandon` stays idempotent so cleanup paths can't trip it. Tests included.

---

src/fleche/wrapper.py
pmrv: Done in #825: `except Rejected` → warning → run uncached, mirroring the save-side handling; the comment block above is trimmed to the invariant while I was there.

---

src/fleche/remote.py
pmrv: Added in #825: a live wire round trip (arguments stashed server-side before the body runs, key sealed against argument mutation, commit ships the live result) and the read-only branch (local digest-only admission with zero RPCs, commit rejected locally). Adding them also surfaced a test-ordering leak in `test_prepared_call.py`'s e2e tests — sticky `fl.cache("memory")` populates the `_live_caches` singleton and crashes `test_run_server_serves_active_cache_over_stdio_until_eof[memory]` when the files run back to back — fixed in the same PR by scoping a private `Cache` instance.

---

src/fleche/caches.py
pmrv: Remaining trims landed in #830 (the `CacheStack.save` comment was already deleted with #825): `BaseCache.prepare` docstring compressed, `Cache.prepare` narration dropped, `ReadOnlyMixin.prepare` historical tail cut (MRO sentence kept), `SshCache.prepare` ", as before" removed. Comment-only diff.

---

src/fleche/caches.py
```diff
@@ -294,11 +318,31 @@ def load_value(self, key):
...
+                    # One-shot form: nothing was stored ahead of time, and with
+                    # no function body between digesting and filing there is
+                    # nothing to drift.
+                    digested = call.stash(self.values)
+                elif isinstance(call, PreparedCall):
+                    # Committed result, stored as returned; the arguments must
+                    # NOT be re-saved here — reading them after the body ran is
+                    # exactly the post-mutation keying prepare exists to
+                    # prevent.
+                    digested = call.resolve(self.values.save(call._result))
```
pmrv:
```suggestion
                    digested = call.resolve(self.values.save)
```
Perfection.

pmrv: maybe call.resolve(self.values) for parity with call.stash

src/fleche/caches.py
```diff
@@ -768,8 +825,14 @@ def __post_init__(self):
             if isinstance(c, CacheStack):
                 raise ValueError("CacheStack cannot be nested inside another CacheStack")
 
-    def save(self, call: Call):
-        self.stack[0].save(call)
+    def save(self, call: PreparedCall | Call) -> str:
+        return self.stack[0].save(call)
+
+    def prepare(self, call: Call) -> PreparedCall:
```
pmrv, line 831: Isn't this actually just fine to not overload, since our save overload just forwards to self.stack[0] again?

src/fleche/caches.py
pmrv: Applied in #836 as the storage-parity form: `resolve(values)` saves the pending result itself, `Cache.save` passes `self.values`, and `SshCache.save` passes a minimal `_RemoteValues` shim whose `save` is one `save_value` round trip — which doubles as a seed for the values facade the PR description defers.

---

src/fleche/caches.py
pmrv: Half yes — it depends which part you drop:

- **The override itself is load-bearing.** Without it, `CacheStack` inherits `BaseCache.prepare`, which is the *digest-only* admission — the argument values would never be stashed, and the commit deliberately doesn't re-save them (`Cache.save` only stores the result), so stack records would file with dangling argument references. The routing to `stack[0].prepare` is what gets the stash.
- **The `cache=self` rebind is droppable today.** `return self.stack[0].prepare(call)` commits straight to `stack[0].save`, and since `CacheStack.save` is a pure one-line forward to the same place, that's behaviorally identical. The rebind only pays off if the stack's `save` ever grows policy of its own; keeping it preserves "you commit through the cache you prepared on" for the cost of one `replace`. I lean keep-for-invariant, but dropping it is fine — say the word and I'll fold it into #836 (one test asserts `prepared.cache is stack` and would follow).

---

pmrv:
## Final read-through (head `ae092dd`, post-#825/#830)

Re-read the complete diff against `main` fresh. Verdict: the design is coherent and merge-ready, modulo the two items below. Lifecycle states live in `call.py` with the cache as entry point; routing checks out across every cache type (`Cache`, wrappers, `ReadOnlyMixin`/`FilteredCache`/`CachePool` digest-only admission, `CacheStack` stash-on-`stack[0]`, `SizeLimitedCache` via MRO, `SshCache` two-trip); the wrapper degrades to uncached on every failure path; only `Call` and `DigestedCall` ever cross the wire. Verified on this head: full suite 1661 passed / 11 skipped, `ty check src/` clean, no stale references in `agents/` or `docs/` (the docs rewrite is deferred to `temppath` per the description).

**One bug found, fixed in #836**: a result the server's value storage refuses (`SaveError`) diverged between local and SSH. Locally the commit raises `Rejected` and files nothing; over the wire `_save_value` rode the value through the synthetic call's *argument* position, whose fallback is lenient — the client filed a record with a dangling result digest, which loads as a permanent phantom miss. #836 moves the value to the strict *result* position (with `None` special-cased) and adds regression tests. The `resolve(values)` suggestion from the review threads is folded into the same PR.

**Remaining nit — the PR description**: the Interface-changes section still specs `save_value` as a public `BaseCache` method ("wrappers forward it, CacheStack routes it…", "one round trip per value") and the old test counts. What landed: `save` takes `PreparedCall | Call`; `save_value` exists only in the SSH wire protocol (write-gated client-side, routed through `prepare` server-side); a remote commit is two trips (result value, then the digested record); `PreparedCall` never crosses the wire. Happy to draft the updated text if useful.

Parked separately: the gc/in-flight race is #826. The `CacheStack.prepare` question is answered in-thread (override is load-bearing; the `cache=self` rebind is the droppable half).

---

src/fleche/caches.py
pmrv: Rebind dropped in #836 (merged): `CacheStack.prepare` now returns `stack[0]`'s admission directly and the commit files there; the test follows.

---

pmrv (commented):
lgtm, merge!

Issue #517, opened 2026-05-11
What's the digest of a non existing file?


pmrv:
Files naturally hash by their content, but what about the empty file?

pmrv:
This comes up when a function takes a path as argument.  There's at least two intents I can think of, either the function is supposed to write some there.  In that case we'd want to hash by path, so that it re-runs if a different path is supplied.  OTOH, if it's supposed to be an input than hashing by content (empty) is correct, function doesn't care where nothing is read from.

pmrv:
@claude Play through some other plausible scenarios.  Do you see a unifying option?  Make it Indigestible and bail on such function is an option too.

pmrv:
**Claude finished @pmrv's task in 1m 56s** —— [View job](https://github.com/pmrv/fleche/actions/runs/25768098721)

---
### Scenario analysis

- [x] Gather context
- [x] Play through plausible scenarios and identify unifying option

---

Here are the distinct scenarios I can think of for a `Path` argument to a fleche-decorated function:

**A. Input file (read intent)**
```python
@fleche()
def parse(path: Path) -> dict:
    return json.loads(path.read_text())
```
Hash by content is exactly right here. If the file doesn't exist yet, the function would raise `FileNotFoundError` before a result could be saved anyway, so `Indigestible` → run-uncached is harmless and honest.

**B. Output file (write intent)**
```python
@fleche()
def render(data: dict, out: Path) -> None:
    out.write_text(json.dumps(data))
```
The path is a destination. You want the cache key to track *where* the result goes, not what's there now (the file doesn't exist on first call). Hash-by-content is actively wrong here — it would give the same key for any two non-existing output paths, so `out=Path("a.json")` and `out=Path("b.json")` both return the same cached `None` after the first run. The right answer here is `Ignored[Path]` or hash-by-path-string, not hash-by-content.

**C. Marker / sentinel file (existence check)**
```python
@fleche()
def setup(done_marker: Path) -> None:
    done_marker.touch()
```
The file is a done-flag. You want to re-run if the path changes. Whether the file exists is almost irrelevant to the cache key (the user is probably using `Required` or `Ignored` here). Hash-by-content doesn't serve this use case either.

**D. Computed output path (pure function over names)**
```python
@fleche()
def output_path(base: Path, stem: str) -> Path:
    return base / f"{stem}.h5"
```
The `Path` arguments are pure labels; they never refer to real files. Hash-by-path-string is the only sensible option; existence is irrelevant.

[...]

Issue #516, opened 2026-05-11
First class support for Path, files and directories


pmrv:
flechette's should be able to operate one files transparently, ie. they would be saved in caches and restored to filesystem as needed

PR #797, opened 2026-07-27
feat(storage): cache filesystem paths by content

pmrv:
**Stacked on #793** (`prepared-call`) — merge that first; this PR shows only the path-caching commits on top of it.

## What

First-class support for `Path` values in cached functions (#516, #33, #517): files and directories are stored by **content** and rematerialized under temporary paths on cache hits.

- **Identity**: a file is `(basename, content)`; a directory is its tree alone (child names in, root name out); plain `bytes` is the content-only escape hatch. Nonexistent paths warn and run uncached (#517).
- **Storage**: `PathValueMixin` converts paths to `FileBlob`/`DirectoryBlob` records referencing content-addressed `bytes` blobs (identical bodies deduplicate across names, directories, and bytes values), wired into every default value storage. Digest salts mirror storage exactly, preserving `digest(path) == save(path)`.
- **Materialization**: hits return `TempPath`s — `Path` subclasses whose backing temp tree lives while any derived path is referenced (3.12+ via `with_segments`; 3.11 via a weak registry of live roots).
- **Destructuring**: paths are found inside dicts / lists / tuples / dataclasses / attrs, as values or keys, at any depth. `DigestedDict` is generalized to a type-preserving `DigestedMapping`; the destructurer predicate is an exact-type allowlist (`dict`, `OrderedDict`, `list`, `tuple`) since subclasses may repurpose the `type(value)(pairs)` reconstruction contract (`defaultdict`, `Counter`).

## Docs

- `docs/usage/file_semantics.rst` — the full user-facing contract (identity, materialization, lifetime, nesting, opaque containers, argument mutation, fidelity limits). Black-box validated: agents restricted to reading only this page wrote assertion scripts against it over three rounds; every failed or ambiguous expectation was folded back into the text.
- Recipes page, dev-internals page, and a runnable `notebooks/PathsInContainers.ipynb` exploring the container-nesting edge cases.

## Interplay with #793

Path arguments make argument mutation easy (writing into a received directory), so this branch is where the two-phase save protocol pays off visibly: `tests/integration/test_path_mutation.py` pins that mutating consumers hit on honest repeats, the mutated tree is a distinct call, and a mutated argument passed back out is captured in its final state.

Full suite green on Python 3.13 and 3.11.


docs/usage/file_semantics.rst
```diff
@@ -0,0 +1,222 @@
...
+most commonly, writing an output file *into* a directory it received — is
+still recorded under the pre-call content, so honest repeat calls hit.  The
+mutation itself, however, is neither recorded nor replayed: a cache hit
+leaves the argument untouched, so a side effect on the input happens on cold
+calls only.
+
+fleche caches *pure* functions.  What a function does to its arguments
+without passing it back out is invisible to the cache — treat received paths
+as read-only and write outputs to a fresh directory (``tempfile.mkdtemp``).
+A mutated argument that *is* returned is captured faithfully in its final,
```
pmrv: feels almost more like it should go into a more generic section rather than file specific.

docs/usage/file_semantics.rst
```diff
@@ -0,0 +1,222 @@
...
+   Path(loc).exists()       # ... False: the temp tree was deleted
+
+Keep the ``Path`` object (or the container holding it) for as long as you need
+the file.  If you need the file at a stable location, copy it out:
+``shutil.copy(p, dest)``.
+
+Paths nested inside containers
+------------------------------
+
+Paths are found and content-stored inside the containers fleche takes apart:
```
pmrv: cross ref destructuring here, since that's the key

notebooks/Files.ipynb
pmrv, line 1: save executed version

pmrv, line 1: move the completedprocess digester into fleche itself

pmrv (commented):
revise

pmrv:
Interactions with ssh cache unclear

pmrv:
Pushed `4e2083e` here as you asked (from the [#831 thread](https://github.com/pmrv/fleche/pull/831#issuecomment-5208623452)) — `ty check src/` is green on `main` but was reporting three diagnostics on this branch, so the ty workflow was failing for every PR into it, and after #831 landed, for the branch itself.

Three suppression comments, no behaviour change:

```diff
-            (self.get(storage, k), self.get(storage, v))
+            (self.get(storage, k), self.get(storage, v))  # ty: ignore[too-many-positional-arguments]

-        return type(value)(zip(children[:n], children[n:]))
+        return type(value)(zip(children[:n], children[n:]))  # ty: ignore[too-many-positional-arguments]

-class TempPath(type(Path())):
+class TempPath(type(Path())):  # ty: ignore[unsupported-base]
```

Both are false positives on code that's correct at runtime. In `DigestedMapping`, ty narrows `type(value)` to `type[Mapping]` — an ABC with no `__init__` — so the call resolves to `object.__init__` and the argument reads as one positional too many; the runtime type is always a concrete `dict`/`OrderedDict`. For `TempPath`, ty can't resolve an MRO through a dynamic base, and that indirection has to stay while 3.11 is supported, since bare `Path` only became subclassable in 3.12.

I suppressed rather than reworked: a `cast` on the two constructors trades a checker complaint for an indirection that earns nothing, and the `TempPath` base can't change until the 3.11 floor moves. Say the word if you'd rather have the `cast` version in `destructuring.py`. Placement matters — ty binds a suppression to the line it *reports*, which for the first one is the generator expression, not the `return type(self.items)(` above it.

Verified with the pinned `ty == 0.0.62` CI uses: **All checks passed!** Suite unchanged at 1763 passed / 11 skipped.

#828 is rebased on top of this and green locally (1788 passed / 11 skipped).

---

pmrv:
Conflict resolved in `fa5d2d2` — but it wasn't only textual, and your instinct about the path/destructuring interaction was right, just in a different place than depth. Two real bugs came out of it.

## 1. `prepare` is a fourth values-over-the-wire route

`prepared-call` adds `SshCache.prepare`, which ships the **whole live call** so the remote stashes argument values before the body runs. #828's guard covered `save_value`, `save`, and `load_value` — not this. So a path argument would have gone over as a bare string for the server to resolve against its own filesystem: the same silent `digest(x) == save_value(x)` break, via a new door. (The old comment there even said "paths over SSH are unsupported" — again, nothing enforced it.)

Resolved by keeping both sides and closing the route:

- `save` now handles `PreparedCall` as well as the live-`Call` and `DigestedCall` forms, and routes the result through `save_value` rather than `_rpc` so the guard applies to it.
- `prepare` falls back to the local `BaseCache` two-phase prepare when any argument carries a path. Each argument then goes through `save_value`, where a path degrades to a digest-only reference computed **locally** — seal intact, lookups still hit. Same degradation #828 established elsewhere, no new concept. The fallback is narrow: ordinary calls keep the one-round-trip path, pinned by a test.

## 2. The guard was mirroring the wrong walk

Chasing your concern I checked `remaining_depth` first, and it's clean — 36/36 across `remaining_depth` ∈ {0,1,2,3,5,10} × nesting depth ∈ {0,1,2,3,5,8}, with the original file deleted before load, and with the setting verified to actually change storage layout (17 → 10 → 3 → 1 entries). A `Path` matches no destructurer, so its depth stays `inf`, it's never inlined, and `PathValueMixin` always gets it.

But the check turned up something worse. `find_path` mirrored **destructuring** — and the invariant at stake is `digest(x) == save_value(x)`, while `digest` recurses **further** than destructuring does:

| value | destructuring | `digest` reads the file | guard saw it (before) |
|---|---|---|---|
| `[path]` | descends | yes | yes |
| `Bundle(path, 0.5)` (namedtuple) | **opaque** | **yes** | **no** |
| `{path}` (set) | **opaque** | **yes** | **no** |
| `frozenset({path})` | **opaque** | **yes** | **no** |

A path hidden in any of those still decides the key, so it crossed the wire and the server digested that name against its own disk. Locally this is harmless — same process, same filesystem — which is exactly why it wasn't visible before.

`find_path` now follows `digest`, stopping where `digest` stops (an arbitrary object is `Indigestible`, so no path inside it can decide a key) and declining to consume one-shot iterables rather than exhausting them as a side effect.

## Verification

1807 passed / 11 skipped, `ty check src/` clean. New tests pin: `prepare` not shipping a path-carrying call while still using the remote for ordinary ones, `commit` rejecting a path result, and namedtuple/set paths being refused at the boundary.

Two API changes on your branch that my #828 tests had to follow: `PreparedCall.key` → `to_lookup_key()`, and `Cache.save_value` is gone in favour of the two-phase protocol.

---

pmrv:
Retarget to `main` conflicted; resolved in `add1dea`. Nine files, and one of them was a trap worth reading about.

**Where main won.** Everything the two-phase-save protocol evolved into during #793's review is taken from main as-is: `PreparedCall.resolve` now takes the value storage and saves internally instead of accepting a pre-computed digest, `CacheStack.prepare` binds `stack[0]` directly, and the server-side `_save_value` rides the *result* position for strict save semantics. `caches.py`, `call.py`, and `test_prepared_call.py` are main's wholesale.

**Where both applied.** `storage/{void,pickle_file,bagofholding_file}.py` — main added `@register_storage` + `to_config`, this branch added `PathValueMixin` to the value-storage MRO. Kept both.

**`remote.py`.** Main's `_RemoteValues` adapter is a better hook for the path guard than what I had, so the guard moved there: `_RemoteValues.save` now routes through `SshCache.save_value` rather than `_rpc`, which means a committed path result is refused at the single existing guard site instead of being resolved against the server's filesystem. One guard site, not two.

## The trap

Git produced **two** `SshCache.prepare` definitions — this branch's guarded one and main's unguarded one — as separate additions, with **no conflict marker**. The file parsed fine. The later definition silently shadowed the guard, so a path argument would once again have been shipped for the server to resolve against its own disk: precisely the bug #828 closed, reintroduced by a clean merge.

`test_prepare_does_not_ship_a_call_carrying_a_path` caught it — the server's value store had the file's bytes in it when the test asserted it should be empty. Main's copy is removed. I also ran an AST duplicate-definition scan over the merged modules to check for others; the only other hit is `BaseCache.shrink` x3, which is its `@overload` set and identical to main.

Worth flagging generally: a silently-shadowed method is a merge failure mode that neither conflict markers nor a syntax check will catch, and it was only visible because a behavioural test existed for the guard.

## Verification

1860 passed / 11 skipped, `ty check src/` clean (pinned 0.0.62), docs build clean.

Still parked on your review: the `find_path` widening from `fa5d2d2` — whether the guard should follow `digest` (catching namedtuple/set-hidden paths) or stay mirrored to destructuring. Unchanged by this merge, and reverting remains contained.

---

pmrv:
The 192 red rows in the benchmark comment above are **runner variance, not a regression**. Don't act on them.

The decisive evidence is the previous run on this same PR, ~40 minutes earlier:

| run | head | baseline | significant changes |
|---|---|---|---|
| 17:03Z | `add1dea` | `8d7dc91` | **0** |
| 17:46Z | `4336827` | `ecb3f4e` | **192**, all red, +10% → +43% |

`add1dea` → `4336827` is the merge of main's #839. I AST-compared every `src/` file between the two heads with docstrings stripped, and the *only* executable difference is `call.QueryCall` → `QueryCall` in one abstract method's annotation (same object, different AST node) plus the added imports and `__all__` entries in `storage/__init__.py`. Everything else `git diff` shows in `src/` is #839's Sphinx cross-reference pass rewriting docstrings. No hot-path code changed at all.

The shape confirms it independently — the slowdown is uniform across things this branch structurally cannot touch:

- `Memory(Raw)` `numpy_arrays` `contains_hit`: 1.7 µs → 1.9 µs (+11.8%) — the raw backend, no destructuring, no `PathValueMixin`.
- `Digest` / `Integer`: 120 µs → 140 µs (+16.7%) — pure hashing, never reaches storage.
- `SqlFile` `calls` `contains_miss`: 200 µs → 240 µs (+20%) — call storage, untouched by this branch.

A change that made *every* measurement 10–40% slower, including paths with no code between them and the branch, is a slower machine. If the path work were genuinely carrying a 30% cost it would also have shown up in the 17:03Z run against the same code.

There's precedent in the tracker for exactly this: the 2026-06-18 ×13 `H5+Sql data_heavy miss` spike that didn't recur on 06-25 and was recorded as I/O jitter.

Worth re-running against a fresh baseline before anyone reads these numbers as real. I'd suggest the delta table stay collapsed-by-default when the *entire* row set moves one direction — that pattern is nearly always the runner, and it buries any genuine single-row regression in noise.

---

pmrv:
Did a review pass over this PR (head `add1dea`). Overall the design holds together well — I checked several invariants that all came out clean: `digest(path) == values.save(path)` holds for files and directories, directory digests are iteration-order independent, symlink cycles terminate via ELOOP, the `DigestedDict` alias keeps old pickles loadable, and wrapper/stack `prepare` rebinding routes commits through the wrapper's save policy correctly. The config rebase logic also correctly skips `~`, absolute paths, and ssh `workdir`.

That said, there are a few real problems, roughly in severity order. The top two and #4 were verified by running reproduction scripts against the PR head.

### 1. `Cache.gc()` permanently destroys stored Path values (verified)

`src/fleche/storage/paths.py:244` — `FileBlob`/`DirectoryBlob` child references are invisible to the reachability walk. `DestructuringMixin.child_digests(key)` calls `super().load(key)`, which `PathValueMixin.load` intercepts and materializes into a `TempPath`, so `_raw_sub_digests` returns an empty set and the content blobs backing the path are never marked reachable. `gc()` then evicts them, and the next `load(key).result` raises `KeyError` — which the wrapper silently treats as a cache miss, so every path-valued entry is lost after any maintenance gc. This is especially bad because `PreparedCall.abandon()`'s docstring explicitly directs users to run gc to reclaim orphans. Side effect of the same wiring: `gc`/`count_reuses` materialize every stored path entry into temp dirs just to inspect it.

### 2. `gc()` during the prepare→commit window evicts just-stashed arguments (verified)

`src/fleche/caches.py:321` — if `gc()` runs while a two-phase call is between `prepare` and `commit` (i.e. while the function body executes), the stashed argument values aren't referenced by any record yet and get evicted as orphans. The later `commit(result)` then files a record with dangling argument references — `lc.arguments[...]` returns a bare `Digest`. Before this PR, values and record were written together inside `save()`, so the unreachable window was microseconds; the two-phase protocol widens it to the whole body duration with no guard or documentation. Even a note in the gc docstring ("don't run gc while calls are in flight") would help, but a real fix probably needs prepared-but-uncommitted values to count as reachable.

### 3. `SshCache.prepare`'s path fallback silently degrades *all* arguments, not just paths

`src/fleche/remote.py:964` — the fallback contradicts its own comment ("each argument goes through save_value") and `RemotePathUnsupported`'s documented degradation ("only the [path] bytes are unavailable"). When `find_path` triggers the fallback, `super().prepare()` digests everything locally and stores nothing, so for `f(p: Path, df: DataFrame)` the DataFrame is never stored either — `load(key).arguments['df']` later returns a bare `Digest`. The intended behavior per the comment looks like `call.stash(_RemoteValues(self))` for non-path arguments, with only the path arguments falling back to digest-only.

### 4. `find_path` misses re-iterable containers that `digest` walks (verified)

`src/fleche/storage/paths.py:64` — `find_path` only descends into exactly `list`/`tuple`/`set`/`frozenset`, but `digest`'s `Iterable` arm walks any re-iterable container. A `Path` inside e.g. a `deque` bypasses the SshCache guard: the digest reads the file locally, but the live value ships over the wire and the server re-digests the path name against *its* filesystem — exactly the `digest(x) == save_value(x)` seal break that `RemotePathUnsupported` exists to prevent. The generator concern in the comment doesn't apply to re-iterable containers, which could be walked safely.

### 5. Unreadable paths raise instead of degrading to `Indigestible`

`src/fleche/digest.py:171` — the new Path arm maps not-exists and non-file/dir to `Indigestible`, but a `PermissionError`/`OSError` from actually reading the content escapes raw. A wrapped call `f(path)` on an unreadable file now crashes before the body runs, where before this PR (and for every other undigestable argument) it would gracefully run uncached. Read errors should degrade the same way as the other cases in that arm.

### 6. Remote cache miss now costs three sequential round trips

`src/fleche/remote.py:902` — a miss pays `prepare` (ship arguments), `save_value` (ship result), then `save` (file record), versus one `save` before. On a high-latency link that's meaningful per-call overhead. The prepare trip is inherent to sealing before the body, but the commit's two trips could plausibly be a single `commit` RPC carrying `(digested_record, result_value)` that the server resolves and files atomically — which would also close the partial-failure window the comment on `save` already acknowledges (result stored, record never filed). Not a blocker, but worth considering while the protocol is fresh.

### 7. Test leaks the sticky process-wide cache

`tests/integration/test_path_mutation.py:41` — the tests call `fl.cache('memory')` without entering the returned context manager, so the sticky ContextVar is never restored and leaks into every later test in the session (order-dependent flakiness; runs-counting tests can see stale hits). The neighboring `test_paths_workflow.py` / `test_prepared_call.py` use the `with` form — same pattern here would fix it.

---

Findings 1 and 2 are the ones I'd consider blocking, since they turn routine gc into data loss. 3–5 are correctness gaps in the degradation story that the PR's own comments/docs promise; 6–7 are take-or-leave.

---

pmrv:
Start with issue 1.

pmrv:
Finding 1 fixed in `28df4cd`. Reproduced first, and it's as bad as you describe — one `gc()` and every path-valued entry is gone, reported as an ordinary cache miss:

```
value entries before gc: 6
child_digests of the record's result: 2 children reported
gc evicted: 2 entries; 4 remain
AFTER GC — KeyError: 'e5e95d01...'
```

Two causes, and the second is why declaring the references alone wouldn't have been enough.

**The blobs declared no children.** `_raw_sub_digests` is now a cooperative chain instead of a single closed `match`: `DestructuringMixin` reports its `Digested` wrappers and delegates anything it doesn't recognise down the MRO, `PathValueMixin` reports a `FileBlob`'s content and a `DirectoryBlob`'s entries, and `ValueStorage` terminates it with the empty set. So a storage's reachable set is the union over its layers rather than whichever layer answers first — which is what you'd want anyway the next time a layer wraps values in a record of its own.

**The walk read through `load`, which mends.** This is the part I'd have missed if I'd only patched the blob case: mending *resolves child references away*, so once `PathValueMixin.load` has materialized a `TempPath`, the reference is simply not in the object any more. Declaring children on the blob wouldn't help, because the walk never sees a blob. It now reads a new `ValueStorage.load_raw` — the same read with every mending layer skipped, which works because mending layers only ever override `load`. Locking is unaffected; `load_raw` still goes through `_operation_context`.

That also closes the side effect you noted: `gc` and `count_reuses` materialized every stored path into a temp tree just to inspect it. Verified now at zero:

```
gc evicted: 0 | temp dirs materialized during gc: 0
temp dirs materialized during count_reuses: 0
entries intact after gc: alphaalpha | ['a.txt', 'b.bin', 'sub']
```

**Tests** (`tests/unit/caches/test_gc.py`): content survives gc for a file, a directory tree, and both nested in a dict; an orphaned path is *still* evicted (the fix must not make path content unconditionally reachable); blobs report their children; and the walk materializes nothing. Five of the six fail against the previous source.

1869 passed / 11 skipped, `ty` clean, docs clean. Wrote the invariant up in `dev/path_storage` — "Blobs must declare their references" — since the next mixin to add a record type will need to know it.

Moving on to finding 2 (the prepare→commit gc window) unless you'd rather I took a different one next.

---

pmrv:
Same artifact as the 192-row run — see [the analysis above](https://github.com/pmrv/fleche/pull/797#issuecomment-5220343117); still uniform across paths this branch cannot touch (`digest(Integer)` +16.7% never reaches storage, `SqlFile calls` +20% doesn't use `ValueMixin.load`).

Flagging it separately only because this is the first run containing a change of mine to the **load path** — `28df4cd` made `ValueMixin.load` delegate to the new `load_raw` — so "fix pushed, 182 regressions" invites exactly the wrong inference. Measured it rather than assuming:

```
load()      11.717 µs/entry   (full mending chain)
load_raw()  10.733 µs/entry
one bare Python call ≈ 0.0339 µs  →  ~0.29% of a load
```

The added delegation is one Python call. The ~1 µs gap between the two is the *mending* work, which was there before. Nothing near the 12–30% being reported, and the same uniform pattern appeared on `4336827`, which had no load-path change at all.

---

pmrv:
next issue 5

pmrv:
Finding 5 fixed in `59f17e6`. Also subscribed to this PR's activity, so I'll see comments here directly — sorry for missing that you'd picked 5 next.

Reproduced first. `chmod 000` is useless here since the container runs as root, so I used a file that errors on read regardless of privilege:

```
exists: True | is_file: True
raw read_bytes()  -> OSError: [Errno 5] Input/output error
digest(p)         -> OSError: [Errno 5] Input/output error   <-- escaped raw
```

Now:

```
[fleche] No hash for argument: Could not read /proc/self/mem: [Errno 5] Input/output error
call 1 -> body ran
call 2 -> body ran
body executed 2x (uncached), records stored: 0
```

The wrap covers the directory walk too, so one unreadable child no longer makes the enclosing tree raise — that seemed clearly within the same defect rather than a separate one. `Indigestible` isn't an `OSError`, so the deliberate raises inside the block pass through untouched.

**Tests** patch the read instead of chmod-ing, so they stay meaningful when the suite runs as root and mode bits aren't enforced: unreadable file, unreadable directory, one unreadable child of an otherwise readable tree, the end-to-end uncached run, and a guard that readable paths still digest normally. Four of the five fail against the previous source.

I also renamed the `file_semantics` section from "Nonexistent paths" to "Paths with no readable content" and stated the widened contract, since "does not exist" was doing the work of "has no readable content" and that's what the arm actually means now.

1874 passed / 11 skipped, `ty` clean, docs clean.

Findings 1 and 5 done. Remaining: 2 (blocking — the prepare→commit gc window), 3 and 4 (both in code I added, happy to take those next), 6 and 7. Say which.

---

pmrv:
Worth noting because it settles the earlier benchmark noise: this run used a **recomputed** baseline (`43b33b1`) instead of a cached one, and the count went **192 → 3**. Same branch, same code. That's the cached-baseline artifact confirmed rather than argued.

The 3 that remain are all `H5+Sql` `miss` (+17.8% / +19.2% / +27.5%). I don't think they're this PR either, though with less certainty than before:

- **Only H5 moved.** If the path work added per-save cost, `Pickle+Sql` and `Memory` misses would move with it — they're all green. `PathValueMixin.save` is `if isinstance(value, Path): ... ; return super().save(value, key)`, so for the non-path values these benchmarks use it's one isinstance and one extra frame, against a ~5 ms measurement.
- **No extra puts.** The only way to move an H5 number meaningfully is to change the number of `put`s, since each is a full file open/close. The path layer adds puts for `Path` values only, and there are none in these workloads.
- **This exact metric has form.** `agents/DEVELOPING.md` records the 2026-06-18 ×13 `H5+Sql data_heavy miss` spike that didn't recur on 06-25 and was concluded to be I/O jitter, plus the standing note that `BagOfHoldingH5File` pays a structural ×5 on save from per-`put` file open/close — the row most exposed to disk variance on a shared runner.

So: same family, same direction, no plausible mechanism, and a documented history of jitter in precisely that row. If you'd like certainty rather than inference I can re-run the benchmark job — cheap now that a recomputed baseline gives a clean signal.

---

pmrv:
Finding 2 fixed in `a9e1a13`. Reproduced first — a `gc()` from inside the function body is enough:

```python
@fl.fleche
def f(x):
    cache.gc()          # routine maintenance, mid-body
    return len(x)

f("x" * 32)
cache.load(key).arguments["x"]
# before: 'd5b3cf31a53e6f64bc02c662e36f48fe87949e4def8fef302f16c5a4a81679f9'
#         a bare Digest — the sealed argument was swept mid-body
# after:  'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
```

**Prepared-but-unfinished calls are now gc roots.** `PreparedCall` registers itself in a process-wide weak registry on construction and drops out on `commit` (by then the record references the values — or the save was rejected and they *are* genuine orphans) or on `abandon`. `Cache.gc` seeds its reachable set from `call.in_flight_digests()`, so the existing transitive walk picks up nested arguments too, not just the top-level keys.

Two details that decided the shape:

**Registration lives in `__post_init__`, not in `prepare`.** `CacheWrapper.prepare` is `replace(self.cache.prepare(call), cache=self)` — a *new* object. Registering in `prepare` would leave only the inner one registered, and that one is garbage the moment `replace` returns, so every wrapped cache would have kept the bug. `test_gc_keeps_arguments_of_a_call_in_flight_through_a_wrapper` forces a `gc.collect()` between prepare and sweep to pin that down.

**The registry is weak.** A prepared call dropped without `commit` or `abandon` frees its roots on collection instead of pinning values for the life of the process. (Keyed by `id` because `PreparedCall` is an unhashable dataclass; `WeakValueDictionary` only drops a key whose stored reference is still the dead one, so a recycled `id` can't evict a live entry.)

**What this does not close**, both now stated in `gc`'s docstring rather than left implicit:

- The registry is **per-process**. The concrete case is `SshCache`: the arguments are stashed by the *server's* `prepare`, and only the sealed record travels back, so a sweep run on the server has nothing to key on while the client's body runs.
- Within a process there is still a window between storing a value and registering the call that owns it. I narrowed it by reading the roots *after* listing the eviction candidates — a value stored after that snapshot isn't a candidate at all — but not to zero. That residue is the same window the one-shot `save()` has always had between `stash` and `calls.save`: bounded by a storage write, not by a function body. Closing it properly needs a lock shared by every writer, which felt like the wrong trade for this PR; say the word if you'd rather have it.

Seven tests in `tests/unit/caches/test_gc.py`, including the negatives that keep the fix from degenerating into "never collect anything" — an abandoned call's arguments are still swept, and so are a committed call's once its record is evicted — plus a threaded one that sweeps while four threads admit calls.

Full suite green (1887 passed, 11 skipped), `ty` 0.0.65 clean.

Remaining: 3 and 4 (both in code I added), 6 and 7. Happy to keep going in that order unless you'd rather reorder.

---

pmrv:
CI on `a9e1a13` went red on my own new threaded test, and it was right to. Fixed in `ff168a0`.

Not the cause the bot summary suggested (registry locking) — the registry read was already a snapshot taken under a lock. The failure was a **second, older race in `gc` itself**, which the in-flight roots made visible rather than introduced:

```
gc: read call records          <- call X's record not filed yet
      ...X commits: files record, then deregisters...
    read in-flight registry    <- X already gone
    evict                      <- X's arguments swept, under a record that references them
```

A call that committed *between* those two reads was covered by neither. Same dangling-argument symptom as finding 2, through a narrower door, and it predates this PR in the sense that the record read has always come first — there was just no registry for it to race against before.

**The fix is an ordering, not a lock.** `commit` deregisters only *after* `save` has filed the record, so the two protections overlap in time. Read the registry *before* the records and a call finishing mid-sweep is always seen by one of them. Candidates are listed first, so a value stored later in the sweep is not a candidate at all:

```python
candidates = list(self.values.list())
reachable = call.in_flight_digests()
for key in self.calls.list(): ...
```

That is also why I didn't reach for the suggested global lock: it would have to be held by every writer for the whole sweep, and the overlap already gives the same guarantee for free.

Confirmed both directions locally rather than assuming — with the old order the test fails ~3 of 8 runs; with the new order 15 of 15 pass, and I raised the test's sweep count from 50 to 200 so it has more power on a fast runner than it did on CI. `tests/unit` green (1796), `ty` 0.0.65 clean.

The residual window is now only between storing a value and registering the call that owns it, and only if it spans the first two reads — the same window the one-shot `save()` has always had between `stash` and `calls.save`. `gc`'s docstring spells out the ordering and what it does and does not cover.

---

pmrv:
Finding 3 fixed in `4c10b92`. CI was green on `ff168a0` and you hadn't picked a next one, so I took this rather than idle — it's my code from `fa5d2d2`. Say the word and I'll reorder.

You were right that the fallback contradicted its own comment. `super().prepare(call)` is `BaseCache.prepare` — `call.digest()`, which digests everything and stores *nothing*. So one path argument cost every sibling:

```python
remote.prepare(Call(name="f", arguments={"p": a_file, "payload": {"rows": [1, 2, 3]}}))
# before: arguments["payload"] -> '65d52a82…'   (never stored, for no reason)
# after:  arguments["payload"] -> {'rows': [1, 2, 3]}
```

**The fix is the shape you sketched**, and it turned out to need no new degradation logic. `call.stash(_RemoteValues(self))` routes each argument through `save_value` one at a time; only the path arguments raise `RemotePathUnsupported`, and `Call.stash`'s existing per-argument `SaveError` fallback already does exactly the right thing with them — `digest(v)` computed locally, so the `digest(x) == save_value(x)` seal holds and lookups still hit. The siblings are stored normally and stay loadable off the record.

```python
return PreparedCall(digested=call.stash(_RemoteValues(self)), cache=self)
```

Worth naming the cost explicitly since it's the one real trade: this branch now pays **one round trip per argument** instead of one for the whole call. It only applies to calls that carry a path — the no-path path still gets the single `prepare` RPC, which `test_prepare_still_uses_the_remote_when_no_path_is_involved` pins. Given finding 6 is about trip count, tell me if you'd rather have batching first and this second.

One existing test had to change and I want to flag it rather than have it look like drift: `test_prepare_does_not_ship_a_call_carrying_a_path` asserted `not server_cache.values.storage`, which is now false by design — the sibling *does* reach the server. It now asserts the narrower true thing, that neither the path's digest nor its content bytes are there. Plus a new `test_prepare_degrades_only_the_path_argument_not_its_siblings` covering prepare → commit → load end to end.

Docs updated in both registers: `file_semantics.rst` (the degradation is per argument) and `path_storage.rst` (why per-argument rather than per-call).

`tests/unit` green (1797), integration green (45 passed, 4 skipped), `ty` 0.0.65 clean, Sphinx builds with no warnings.

Remaining: 4, 6, 7. Taking 4 next unless you say otherwise.

---

pmrv:
Finding 4 fixed in `4d6ded1`. Confirmed the bypass first — `deque` walks straight past the old allowlist:

```python
find_path(collections.deque([path]))
# before: None   (guard never fires; the value ships and the server re-digests the name)
# after:  PosixPath(...)
```

You were right that the generator concern didn't justify the allowlist. The iterable arm now walks whatever `digest` walks, and skips only what `digest` itself never looks inside:

- **`OPAQUE_ITERABLES`** — `ndarray`, `DataFrame`, `Series`, `Index`. These match *above* `digest`'s `Iterable` arm and hash their own buffer, so no element inside one can influence a digest. I put the tuple in `fleche.digest` rather than re-deriving it in `paths.py`, since the whole failure mode here is a mirror drifting from the thing it mirrors.
- **`range`** — elements are `int` by construction. This one is about the walk pushing onto a list rather than streaming: `range(10**9)` would exhaust memory to learn nothing.
- **one-shot iterators**, via `iter(x) is x`. Walking a generator consumes it — and `digest` has already drained one by the time a value could ship, so there is nothing left in it to find.

Tests: seven containers no allowlist would have covered (`deque`, `OrderedDict`, `defaultdict`, `Counter` keys, `ChainMap`, a nested deque, and a hand-rolled `__iter__` class), plus the exclusions asserted for their stated reasons rather than just their outcomes.

**Two things I found while verifying, neither fixed here:**

**1. `digest` of an object-dtype ndarray is address-based.** Verifying "elements never reach a digest" turned up that the ndarray arm's `value.tobytes()` hashes element *pointers* for `dtype=object`:

```python
mk = lambda: "".join(["x", "y", str(9)])
a, b = np.array([mk()], dtype=object), np.array([mk()], dtype=object)
list(a) == list(b)              # True
digest(a) == digest(b)          # False
```

Equal arrays, different keys — so an object array argument never hits across processes, and a path inside one is invisible to both `digest` and this guard (which is why skipping it is faithful, not a shortcut). Pre-existing and unrelated to this PR, and fixing it changes digests, so I left it alone. Want an issue?

**2. `child_slots` is now dead.** I added it in #828 purely to back the old destructuring-shaped `find_path`; nothing calls it now and nothing tests it, but it is exported in `fleche.storage.__all__`. Happy to drop it — say the word, or keep it if you want it as a general helper. I did fix the `path_storage.rst` line that still claimed `find_path` walks the way a destructuring save does.

`tests/unit` + integration green (1851 passed, 4 skipped), `ty` 0.0.65 clean, Sphinx builds with no warnings.

Remaining: 7 (the sticky-cache leak) and 6 (the three round trips — which my finding-3 fix made more relevant, since a path-carrying `prepare` now costs a trip per argument). Taking 7 next unless you'd rather have 6.

---

pmrv:
Finding 7 fixed in `58a0cfe`. Proved the leak before touching it, with a probe test either side of the module:

```python
def test_a_records_the_cache_before(request):
    request.config._before = fl.cache()

def test_z_cache_is_unchanged_after(request):
    assert fl.cache() is request.config._before
```

```
FAILED test_z_cache_is_unchanged_after
  Cache(values=ValueMemory(storage={}, remaining_depth=1), calls=CallMemory(storage={}))
```

So it wasn't hypothetical — the memory cache stayed active for the rest of the session. All three tests now use the `with` form, matching `test_paths_workflow.py` and `test_prepared_call.py`. Nothing depended on the leak: 1851 passed, 4 skipped.

That's **1, 2, 3, 4, 5, 7 done**. Only **6** is left, and I'd rather you decide it than guess — it's the one you called take-or-leave, and it's a protocol change:

- The commit's two trips (`save_value` then `save`) could become a single `commit` RPC carrying `(digested_record, result_value)`, which also closes the partial-failure window your comment on `save` already flags (result stored, record never filed).
- The `prepare` trip is inherent to sealing before the body, so that one stays.
- It got more relevant since your review: my finding-3 fix makes a *path-carrying* `prepare` cost one trip per argument instead of one for the call. If you want batching, that argument loop is the other place it belongs, and I'd do both in one go rather than twice.

Say **6** and I'll take it; say **done** and I'll stop here and leave it for a follow-up issue.

Still open from the last two comments, both one-word answers: whether you want an issue for the object-dtype ndarray digest being address-based, and whether to drop the now-unused `child_slots`.

---

pmrv:
CI went red on `c8a5861` — my own threaded gc test again, this time `[flat]` on 3.14. Fixed in `35c3aae`, but the interesting part is that **the failure was real and the test was wrong**, so I want to be explicit about what `gc` does and does not guarantee.

(The bot summary on this one suggests reordering the sweep to read roots before records — that is already the code, since `ff168a0`. Not the cause.)

**What actually happens.** The ordering fix closed the wide windows. It does not close the gap between `prepare` *storing* an argument value and *registering* the call that owns it. I couldn't reproduce it by hammering — 8 threads × 400 sweeps, clean — so I forced the gap open instead of guessing:

```python
def slow_post_init(self):       # values stored, call not yet a gc root
    time.sleep(0.25)
    orig_post_init(self)
```
```
gc evicted the stored argument: True
committed record hands back  : 'b979068f6d84...'   (bare digest)
```

Deterministic. So the sweep is best-effort against concurrent writers and exact on an idle cache, and my test was asserting the former was the latter. Two CI runs on two Python versions were right and I was wrong to have written the assertion that way.

**What I changed.** The threaded test now asserts only the real guarantee — nothing raises, and the registry survives being read while other threads resize it. The raced assertion is replaced by two deterministic tests:

- `test_gc_keeps_a_call_that_commits_mid_sweep` drives a commit from inside the record read, pinning the `ff168a0` ordering. I checked it has teeth rather than assuming: it fails with the reads in the old order, passes in the new one, no timing involved.
- `test_gc_may_evict_a_value_stored_but_not_yet_registered` pins the window that *stays* open, so it's documented behaviour instead of folklore. If anyone closes it, that test fails and should be deleted.

**Your call on closing it.** It needs writers and the sweep to agree on an order, but not the global lock the bot proposes. The cheap shape is a readers-writer lock: `prepare` takes it shared around store-then-register, `gc` takes it exclusive around just its candidates and roots reads — two in-memory reads, no I/O. Prepares stay concurrent with each other; a sweep waits out the prepares already in flight. Worth it? It is the same window the one-shot `save` has always had between `stash` and `calls.save`, so closing it here without closing it there is half a fix. I'd rather you decide than have me add locking to the write path on a PR this size.

This is also finding 6 territory in spirit — still awaiting your word on that one, plus the object-dtype ndarray digest issue and dropping `child_slots`.

Full suite green at 1902 passed / 11 skipped (all eleven notebooks), `ty` 0.0.65 clean, and the gc file run 10× back to back with no flake.

---

pmrv:
Heads-up before you see it elsewhere: **`main` currently breaks the suite for any PR that touches `src/`**, and it isn't visible on `main` itself. I've deliberately held this branch one merge behind rather than import a guaranteed red — #797 is still green and mergeable at `c108ec2`.

`notebooks/FiveMinuteTour.ipynb` (#841) opens with:

```python
from ase.build import bulk
from ase.calculators.emt import EMT
```

```
ModuleNotFoundError: No module named 'ase'
```

`ase` is not a dependency anywhere — not in `[project.dependencies]`, not in the `tests` extra, not installed by `tests.yml`, which does only `pip install -e .[tests]`. And `test_notebooks.py` globs `notebooks/*.ipynb` with no skip, so the new notebook is collected unconditionally.

**Why nothing went red.** Two filters hid it, and they compose badly:

- The `changes` gate skips the whole suite unless `src/` or `tests/` changed. #841 was notebook-only, so its four `build` jobs "succeeded" in about three seconds without running anything.
- The `tests` workflow doesn't run on pushes to `main` at all — only `ty` and `release-please` do.

So the notebook has never actually been executed by CI. The first PR to touch `src/` after this lands is the one that goes red, which would have been this one.

I ran it locally to be sure it's not an interaction with the path work: 1909 passed, the single failure is that import, in main's notebook, before any fleche code runs.

**Your call on the fix** — I haven't picked one, since they trade off differently:

1. **Add `ase` to the `tests` extra.** Honest (the notebook does need it) but drags numpy/scipy into every CI job for one notebook.
2. **Skip when the import is unavailable** — have `test_notebooks.py` check a per-notebook requirement and `pytest.skip`. Keeps CI light; the notebook then goes untested, which is the thing #839's glob was written to prevent.
3. **Drop the ASE cell** from the tour and use a stdlib example.

Tell me which and I'll do it here or in a separate PR against `main`, whichever you prefer — a separate PR is probably cleaner since it isn't path work.

Meanwhile the merge itself is fine: it auto-resolved, and I checked this branch's content survives (`docs/index.rst` keeps `dev/path_storage`, `notebooks/Files`, `notebooks/PathsInContainers`). I also have the `agents/DEVELOPING.md` notebook counts updated eleven → twelve for `FiveMinuteTour`, ready to go with the merge once `main` is runnable.

---

pmrv:
Merged `main` in `ad7b883`, which means **CI here will go red on `FiveMinuteTour.ipynb`** — the `ase` problem from my last comment, still unfixed on `main`.

I held the merge back for a day to keep this PR green. What forced it: #856 rewrote the same `agents/DEVELOPING.md` notebook enumeration this branch edits, so GitHub flipped #797 to `dirty`. A conflicted PR can't be merged at all, a red one can be read — so I took the red rather than leave the conflict sitting.

Everything except that one notebook passes: **1909 passed, 11 skipped**, `FiveMinuteTour` deselected. The three options are unchanged (add `ase` to the tests extra / skip notebooks whose imports are unavailable / drop the ASE cell) and I still haven't picked one — say the word and I'll do it here or in a separate PR against `main`.

The conflict itself resolved to the union: twelve notebooks, twelve symlinks, main's `FiveMinuteTour` beside this branch's `Files` and `PathsInContainers`, and this branch's `purity` / `file_semantics` / `path_storage` / `recipes/files_and_paths` pages. `docs/index.rst` auto-merged and I verified its entries survived.

**One new thing worth your eyes**, unrelated to the above. Sphinx now emits a warning that it didn't before:

```
docs/autoapi/fleche/index.rst:256: WARNING: duplicate object description of
fleche.D, other instance in digests/digests_as_args, use :no-index: for one of them
```

I checked this rather than guessing at it:

- `origin/main` builds **clean** — I built it in a separate worktree to be sure.
- This tree warns **reproducibly**, twice in a row, fresh output dir each time.
- `docs/digests/digests_as_args.rst` is **byte-identical** to main's and already carries `:no-index:`.

So it isn't a change to that page — it's that the extra pages here shift the order in which Sphinx reads documents relative to the autoapi-generated one, and the generated instance is the one without `:no-index:`. The build still succeeds (exit 0, one warning), so nothing is broken, but it would presumably show up on the next docs build on `main` once this lands. The fix belongs in the autoapi config rather than in path work, so I've left it alone — happy to take it if you'd like it fixed here.

Findings 1–5 and 7 remain done; still awaiting you on finding 6, the object-dtype ndarray digest issue, dropping `child_slots`, and the store-before-register gc window.

---

pmrv:
Thanks for #857 — merged it here in `f2ec078`. **It doesn't clear CI yet, though**: the same notebook still fails, one cell later.

I ran it after merging rather than assuming the fix was complete:

```
ModuleNotFoundError: No module named 'executorlib'
```

Cell 30:

```python
from executorlib import SingleNodeExecutor as Executor
```

Bare import, no guard — and `executorlib` is its own optional extra in `pyproject.toml`, not part of `tests`, which is all `tests.yml` installs. So it's the same shape as the `ase` case: the notebook reaches for an optional dep that CI never has.

The `try/except ImportError` you added for the ASE section applied to this import would do it:

```python
try:
    from executorlib import SingleNodeExecutor as Executor
except ImportError:
    print("executorlib not installed, skipping this section")
else:
    ...
```

I haven't pushed that — it's your notebook and you're evidently mid-fix, so a duplicate patch from a path PR would just collide with whatever you land next. Say the word if you'd rather I did it here.

Worth noting the two filters that let this class of thing through, since it'll keep happening: the `changes` gate skips the suite for notebook-only PRs, and `tests.yml` doesn't run on pushes to `main` — so a notebook that reaches for an optional dep goes unnoticed until some unrelated PR touching `src/` picks up the red. That's how #841 landed, and why #857's remaining half is only visible from here.

Everything else on this branch is unchanged and green: 1909 passed, 11 skipped with that notebook deselected.

Still open for you, unchanged: the Sphinx `fleche.D` duplicate-object warning (main builds clean, this tree warns reproducibly — fix belongs in the autoapi config), finding 6, the object-dtype ndarray digest issue, dropping the now-unused `child_slots`, and whether to close the store-before-register gc window with a readers-writer lock.

---

pmrv:
#862 does it — merged in `d032f71`, and **this branch is green again**, for the first time since #841 landed.

Verified after merging rather than assuming: the notebook that had been failing now passes on its own, and the full suite is **1910 passed, 11 skipped, zero failures**. `ty` 0.0.65 clean.

**The Sphinx `fleche.D` warning is also gone** — and nothing on this branch changed to fix it. As I said when I reported it, it was an ordering effect between the autoapi-generated page and `digests_as_args` rather than a content problem; your docs churn since has evidently shifted the read order back. The build is now warning-free here. So that one is closed, no action needed from you.

Where #797 stands:

| Finding | Status |
|---|---|
| 1 — `gc()` destroys stored Path values | fixed, `28df4cd` |
| 2 — `gc()` evicts in-flight arguments | fixed, `a9e1a13` + `ff168a0` + `35c3aae` |
| 3 — SSH path fallback degrades all arguments | fixed, `4c10b92` |
| 4 — `find_path` misses re-iterable containers | fixed, `4d6ded1` |
| 5 — unreadable paths raise | fixed, `59f17e6` |
| 7 — test leaks the sticky cache | fixed, `58a0cfe` |
| 6 — three round trips per remote miss | **your call, untouched** |

Still open for you, all small and all previously described:

1. **Finding 6** — collapse the commit's two trips into one `commit` RPC. You called it take-or-leave; it also closes the partial-failure window your own comment on `save` flags, and it interacts with the finding-3 fix (a path-carrying `prepare` now costs a trip per argument), so both places want batching together if you want it at all.
2. **`child_slots`** — I added it in #828 to back the old destructuring-shaped `find_path`; nothing calls or tests it now, but it is exported in `fleche.storage.__all__`. Drop it or keep it as a general helper?
3. **Object-dtype ndarray digests are address-based** — `digest(np.array([mk()], dtype=object))` differs between two equal arrays, so such arguments never hit across processes. Pre-existing, unrelated to this PR, semver-breaking to fix. Want an issue?
4. **The gc store-before-register window** — the one race the ordering fix does not close, pinned by `test_gc_may_evict_a_value_stored_but_not_yet_registered`. Closing it needs a readers-writer lock shared by writers and the sweep; it is the same window one-shot `save` has always had, so it is arguably a separate change.

Nothing here blocks a merge — 1, 3 and 4 are questions, 2 is optional cleanup. Happy to take any of them on your word.

---

pmrv:
## `test-minimum-deps` went red on `30a9ccc` — pre-existing `digest` collision on `main`, not this branch

`30a9ccc` is a merge of `main` that touches **only `agents/DEVELOPING.md`**, so I treated any red as new and mine. It isn't — but what it found is a real correctness bug in `main`, so I'd rather flag it than re-run past it silently.

### What failed

Hypothesis found a falsifying example for `tests/unit/storage/test_destructuring_storage.py::test_roundtrip`:

```
value  = {None: [-512], '': -2.220446049250313e-16}
loaded = {None: [-2.220446049250313e-16], '': -2.220446049250313e-16}
```

The `-512` came back as `-2.220446049250313e-16` because the two share a digest:

```python
>>> digest(-512) == digest(-2.220446049250313e-16)
True
>>> -512 == -2.220446049250313e-16
False
```

### Why

The `Number()` arm of `_digest_bytes` ends with `value = hash(value); return _digest_bytes(value)` — a float's digest **is** the digest of the int `hash(value)`. `hash(-2**-52) == -512`, so the float and the int `-512` are indistinguishable to the cache.

That is the documented mechanism (`docs/digests/digest_equivalence.rst`: the type-name salt is deliberately not carried into number digests, which is what makes `digest(1) == digest(1.0) == digest(1+0j)`). What the docs assume, though, is that `hash` is injective. It isn't — it's a 61-bit fold — so the intended *equal-values-collide* becomes *unequal-values-collide*. **Every** non-integral float aliases some int:

| value | `hash(value)` | equal? | same digest? |
|---|---|---|---|
| `0.5` | `1152921504606846976` | no | **yes** |
| `0.1` | `230584300921369408` | no | **yes** |
| `1e308` | `156575653125701` | no | **yes** |
| `Fraction(1,3)` | `1537228672809129301` | no | **yes** |
| `Decimal("0.5")` | `1152921504606846976` | no | **yes** |

It reaches cache keys, not just `digest()`:

```python
>>> f.digest(0.5) == f.digest(1152921504606846976)
True
```

— i.e. `f(0.5)` can return the cached result of `f(1152921504606846976)`.

### Not from this branch

- The collision reproduces against `origin/main`'s `src/fleche/digest.py` in a clean worktree (`COLLIDES ON MAIN: True`).
- This branch does not touch the `Number` arm; its `digest.py` changes are the `Path` / `CompletedProcess` arms and `OPAQUE_ITERABLES`.
- `tests/strategies.py` — which generates the example — is unchanged here, so `main`'s own CI can draw this same example on an unlucky seed.

Seed-dependent: 3000 derandomized examples pass on both `main` and this branch, which is why it has stayed hidden.

### What I'm not doing

Not fixing it here. Salting number digests by type (or by exact value) changes the digest of **every float, complex, `Fraction` and `Decimal`**, invalidating every persisted call that has one as an argument or result — a migration decision for a persistent cache, and squarely yours, not something to smuggle into a path-caching PR. Happy to take it in a follow-up if you want it, in whichever direction you prefer (full type+value salt, or keep cross-type equivalence but digest the exact value rather than `hash`).

I've re-run the failed job so the PR's CI state isn't left red on a seed accident. Flagging this alongside the four questions in [this comment](https://github.com/pmrv/fleche/pull/797#issuecomment-5273745767) — still all awaiting your call, none started.

---

pmrv:
Believe it or not the solution to the gc() vs Path problem is more abstract indirection!  Behind both path and destructuring mixin we add another interface that recurses into selected items and route all item access through that.  path and destructuring then implement that for their specific types.

PR #825, opened 2026-08-06
fix(call): review follow-ups for the two-phase save protocol

pmrv:
Implements the changes approved in the #793 review threads, revised per the #825 review ("combine" with the #827 sketch; broadened prepare guard; two-trip remote commit). Stacked onto `prepared-call` for merging into that branch.

## Changes

- **Pending result lives on `PreparedCall`, never on `DigestedCall`** (combining #827 with the replace-based commit): `PreparedCall` carries `_result`/`_metadata` as private fields, `commit` hands the whole `PreparedCall` to `cache.save`, and the cache stores the result and files a record built via `PreparedCall.resolve` — `DigestedCall.result` is back to a checkable `Digest | None`, and `commit` never mutates `digested`. A bare `DigestedCall` still files as-is.
- **`PreparedCall` never goes over the wire**: `SshCache.save` recreates `Cache.save`'s ending in two trips — a write-gated `save_value` RPC stores the result value remotely, then the plain `DigestedCall` is filed. Only `Call` and `DigestedCall` cross the wire (asserted by a spy in the round-trip test). Server-side, `save_value` routes through `prepare` with a synthetic never-committed call, so wrappers and stacks direct the value to the same storage their saves use without adding a public cache method; a failure between the trips leaves a content-addressed orphan for gc, like any abandoned call.
- **`commit` is single-shot**: a second `commit` or a `commit` after `abandon` raises `RuntimeError`; `abandon` stays idempotent so cleanup paths can't trip it.
- **Wrapper degrades gracefully on any `prepare` failure**: `Rejected` logs a policy warning; any other `Exception` (storage fault, lost connection) logs with traceback — in both cases the body runs and returns uncached instead of failing before it ever ran. Also drops the dead `call.metadata` assignment.
- **Remote `prepare` tests**: live wire round trip (arguments stashed server-side before the body runs, sealed key survives argument mutation, two-trip commit) and the read-only branch (local digest-only admission, zero RPCs, commit rejected locally before the first trip).
- **Test hygiene fix (drive-by)**: the e2e tests in `test_prepared_call.py` activated the shared named-`"memory"` cache stickily; that singleton lives in `config._live_caches`, so populating it leaked into any later test that round-trips the named cache's config (`test_run_server_serves_active_cache_over_stdio_until_eof[memory]` crashes in `dataclasses.asdict` when the files run back to back — masked in the full suite by `tests/unit/config`'s `_live_caches.clear()` fixture). The tests now scope a private `Cache` instance.

Parked separately: the gc/in-flight race is #826.

## Testing

Full suite: 1659 passed, 11 skipped. `ty check src/` clean. The affected file pair also passes in the previously-failing collection order.

https://claude.ai/code/session_01QT4DxU9dVALrMvff7setnk

### DIFF
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -389,13 +389,18 @@ class PreparedCall:
     def to_lookup_key(self) -> Digest:
         return self.digested.to_lookup_key()
 
-    def __getstate__(self) -> dict:
-        # The cache binding is process-local (it may hold live locks or an SSH
-        # connection); a PreparedCall on the wire carries only the record and
-        # the pending result.
-        state = self.__dict__.copy()
-        state["cache"] = None
-        return state
+    def resolve(self, result: Digest) -> DigestedCall:
+        """Return the final record with the pending result resolved to *result*.
+
+        The shared ending of the two-phase save: the cache stores
+        :attr:`_result` wherever its values live and hands the digest back
+        here; pending metadata is applied at the same time.
+        """
+        return replace(
+            self.digested,
+            result=result,
+            metadata=self.digested.metadata if self._metadata is None else self._metadata,
+        )
 
     def abandon(self) -> None:
         """Release the call without recording it.
--- a/src/fleche/call.py
+++ b/src/fleche/call.py
@@ -266,11 +266,9 @@ class DigestedCall:
 
     name: str
     arguments: dict[str, "digest.Digest"]
-    # Normally a Digest pointer, or None while the result is still unknown
-    # (a record prepared before its function body ran).  Between
-    # :meth:`PreparedCall.commit` and the cache's ``save`` it briefly holds the
-    # live return value, which ``save`` stores and replaces with its Digest.
-    result: "digest.Digest | Any" = None
+    # None while the result is still unknown (a record prepared before its
+    # function body ran).
+    result: "digest.Digest | None" = None
     metadata: dict[str, dict[str, Any]] = field(default_factory=dict)
     module: str | None = None
     version: str | int | None = None
@@ -355,6 +353,10 @@ class PreparedCall:
     digested: DigestedCall
     cache: Any
     _finished: bool = field(default=False, init=False, repr=False)
+    # The live result and metadata between commit and the cache's save;
+    # DigestedCall itself only ever holds digests.
+    _result: Any = field(default=None, init=False, repr=False)
+    _metadata: "dict | None" = field(default=None, init=False, repr=False)
 
     def commit(self, result: Any, metadata: dict | None = None) -> Digest:
         """Attach *result* and *metadata* to the record and file it.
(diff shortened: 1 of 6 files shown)


src/fleche/wrapper.py
```diff
@@ -235,14 +235,16 @@ def _run_and_cache():
...
-            # did when the post-body save was rejected.
-            prepared = cache.prepare(call)
+            # Seal the call's identity (and stash its argument values) before
+            # the body runs, so mutations the body makes to its arguments
+            # cannot leak into the recorded key.  A read-only cache seals
+            # without writing and rejects at commit; the call still runs and
+            # returns uncached.
+            try:
+                prepared = cache.prepare(call)
+            except Rejected as e:
```
pmrv, line 245: let's broaden exception here in case there's actual errors in prepare rather than policy rejections

src/fleche/call.py
```diff
@@ -374,12 +374,17 @@ def commit(self, result: Any, metadata: dict | None = None) -> Digest:
...
+            RuntimeError: if this call was already committed or abandoned.
         """
-        self.digested.result = result
-        if metadata is not None:
-            self.digested.metadata = metadata
+        if self._finished:
+            raise RuntimeError("PreparedCall already committed or abandoned")
         self._finished = True
-        return self.cache.save(self.digested)
+        digested = replace(
```
pmrv: In #827 I parked an alternative design.   I like doing the replace, but I still don't like having a value in a digested call.  In the PR you can see I push this back into cache save responsibility.  combine.

pmrv (commented):
revise

src/fleche/call.py
pmrv: Combined in 5d5a67a: `PreparedCall` carries `_result`/`_metadata` (proper annotated fields, so they exist as dataclass fields), `commit` hands the whole `PreparedCall` to `cache.save`, and `Cache.save` stores the result (`values.save`, not `put` — `put` wants the key too) and files a record built with `replace` — `DigestedCall.result` is back to `Digest | None` and `commit` never touches `digested`. One addition the sketch needed: `PreparedCall.__getstate__` drops the `cache` binding, otherwise a commit against `SshCache` would try to pickle the live connection into the RPC. A bare `DigestedCall` still files as-is via the else branch (also keeps `ty` happy about the narrowed union).

---

src/fleche/wrapper.py
pmrv: Done in 5d5a67a: `Rejected` keeps its warning; any other `Exception` is logged with traceback (so real faults stay diagnosable despite being swallowed) and the body runs uncached. `KeyboardInterrupt`/`SystemExit` still propagate. Test parametrized over both flavors.

---

pmrv (commented):
lgetm (looks-good-enough-to-me)

PR #836, opened 2026-08-07
fix(remote): reject unstorable results; resolve() takes the value storage

pmrv:
Three items from the final read-through of #793 and its review threads, stacked onto `prepared-call`:

## Bug fix: unstorable results over SSH filed dangling records

`_save_value` routed the value through the synthetic call's *argument* position, and `stash` treats argument `SaveError`s leniently (digest-only fallback). Verified divergence: locally a commit of a storage-refused result raises `Rejected` and files nothing; over the wire the client got back a digest for a value that was **never stored** and filed a record whose result dangles — which then loads as a `KeyError`, i.e. a permanent phantom miss that re-executes forever.

Fix: the value now rides in the *result* position, whose save semantics are strict, with `SaveError` wrapped in `Rejected` to match `Cache.save`. `None` — the one result the stash skips — rides as an argument, where storing cannot fail. Regression tests cover both: refused result → `Rejected` at the client, no record; `commit(None)` round-trips.

## Review suggestion: `resolve(values)` in parity with `stash(values)`

Per the [suggestion](https://github.com/pmrv/fleche/pull/793#discussion_r3737082003) threads on #793: `PreparedCall.resolve` now takes the value storage and saves the pending result itself, mirroring `Call.stash`. `Cache.save` passes `self.values`; `SshCache.save` passes a minimal `_RemoteValues` shim whose `save` is one `save_value` round trip (also a natural seed for the deferred remote values facade).

## Review decision: drop the `CacheStack.prepare` rebind

Per the [thread](https://github.com/pmrv/fleche/pull/793#discussion_r3737110020): `CacheStack.save` is a pure forward to `stack[0]`, so rebinding the `PreparedCall` to the stack bought nothing — `prepare` now hands back `stack[0]`'s own admission and the commit files there directly.

## Testing

Full suite: 1661 passed, 11 skipped. `ty check src/` clean.

https://claude.ai/code/session_01QT4DxU9dVALrMvff7setnk

### DIFF
--- a/src/fleche/caches.py
+++ b/src/fleche/caches.py
@@ -830,9 +830,9 @@ class CacheStack(PerKeyLockMixin, _MultiCache):
 
     def prepare(self, call: Call) -> PreparedCall:
         # Writes always land on stack[0] (matching save), so that is where the
-        # arguments are stashed; rebinding the cache routes the commit back
-        # through this stack's ``save``.
-        return replace(self.stack[0].prepare(call), cache=self)
+        # arguments are stashed and where the commit files directly — this
+        # stack's own ``save`` is a pure forward to the same place.
+        return self.stack[0].prepare(call)
 
     @contextlib.contextmanager
     def _operation_context(self, key, *, intent: Intent = Intent.WRITE):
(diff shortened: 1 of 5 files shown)


pmrv (commented):
lgtm

PR #839, opened 2026-08-07
docs: re-render notebooks and make every Sphinx cross-reference resolve

pmrv:
## Notebooks

Executed every notebook under `notebooks/` with
`jupyter nbconvert --to notebook --execute --inplace` (the same command the
`rendernb` workflow runs) and committed the refreshed outputs. All eight run
clean — no error outputs, no unexecuted cells.

Two of them needed fixing first. `Caches.ipynb` and `TransferWorkflow.ipynb`
are neither covered by `tests/integration/test_notebooks.py` nor linked from
`docs/notebooks/`, so they had gone stale against the storage split into
value/call backends and failed on the first cell:

- `fleche.storage.Memory` → `ValueMemory` / `CallMemory`
- `PickleFile.with_cloudpickle` → `ValuePickleFile.with_cloudpickle`
- `Cache(values=..., _calls=...)` → `Cache(values=..., calls=...)`
- `cache.filter(f.fleche.call(...))` → `cache.filter(QueryCall.from_call(f, ...))`
  — `.fleche.call()` returns a `Call`, which `filter()` treats as a predicate
  and then cannot invoke

Re-rendered again after merging #793 to confirm the sealed-call-identity change
does not alter them: digests and results are identical, only wall-time jitter
differs, so those outputs were left as they were rather than committing pure
churn.

## Cross-references

A nitpicky build (`sphinx -n`) reported **246 dangling cross-references and 39
markup errors**. Both are now **zero**, and 933 of the 934 cross-reference
spans in the built HTML render as real links (the last one is `typing.Protocol`
in an inheritance line).

Unresolved targets, by cause:

- **No intersphinx inventory** for the stdlib or pandas, so ~100 references to
  `pathlib.Path`, `concurrent.futures.Future`, `threading.Lock`,
  `pandas.DataFrame`, … could never resolve. Added the `python` and `pandas`
  mappings.
- **Bare names** (`` :class:`Digest` ``, `` :meth:`expand` ``, …) only resolve
  on the page that happens to define them, and autoapi renders the same
  docstring on several pages — including inherited docstrings on every
  subclass. Qualified them as `~fleche.<module>.<name>`, which leaves the
  rendered text unchanged.
- **`fleche.remote` defines `__all__`**, so autoapi documents none of its
  private helpers, while the SSH internals guide referenced ~20 of them. They
  are now literals, and the affected paragraphs were rewrapped to the file's
  existing width.
- **Parameter shadowing**: the `call` parameter of `_query`/`save` shadows the
  `call` module, so `call.QueryCall` annotations resolved against the parameter
  and emitted targets like `Cache._query.call`. Both names are already imported
  directly, so the module prefix is gone.
- **Missing exports**: `AnyQueryType`/`StrQueryType` (used in `QueryCall`'s
  signature), `PreparedCall` (added by #793), and the storage backend base
  classes (shown as `Bases:` of the re-exported storages) were not in
  `__all__`, so autoapi never documented them and nothing could link to them.

Markup errors that produced visibly broken output:

- `table()`'s column list used single backticks (title references, not
  literals) and `'a_'`, which docutils read as a reference to an undefined
  target named "a".
- Bullet lists in `normalize_secret_key`, `Sql.query` and `table()` were not
  separated from the paragraph introducing them.
- The helper list in `@fleche`'s docstring wrote `(*args, **kwargs)`
  unescaped, opening emphasis that never closed.
- `Keys:` blocks in `metadata` rendered as malformed block quotes; napoleon now
  parses them like `Args:`.
- `fleche.D` was documented twice (autoapi plus an explicit `autofunction`).

Also fixed `SizeLimitedCache`'s docstring, which still documented the call
storage argument under its old name `_calls`.

Two reference kinds cannot resolve and are listed in `nitpick_ignore`:
`Ellipsis`, which is what autoapi emits for `...` inside `Callable[...]` and
`tuple[T, ...]`, and `fleche.remote._Connection`, which `__all__` keeps out of
the docs.

## Merge with main

Merged `main` (#793, sealed call identity). The one conflict was
`SizeLimitedMixin.save`: main widened the annotation to
`call.PreparedCall | call.Call` while this branch had dropped the `call.`
prefix. Took main's type with this branch's unprefixed spelling, which is what
every other `save` in the file already uses.

## Verification

- `pytest tests/` — 1716 passed, 7 skipped
- `ty check src/` — all checks passed
- `sphinx -b html -n -E docs` — 0 warnings, 0 errors

https://claude.ai/code/session_01C1MFyJNxkhmfkEuB8oGrUg


notebooks/CacheStack.ipynb
pmrv, line 1: showing first how to set up the cache stack from the dict like call that #777 added as from_config would probably make the setup for this look less scary.

notebooks/ConcurrentExecution.ipynb
pmrv, line 1: just use `cache('memory')` instead of setting everything up yourself.

pmrv (commented):
seems like previously not all notebooks were linked to docs/ or into the test runner as well

notebooks/CacheStack.ipynb
pmrv: Done in 59b447e. The stack is now built first from a list of template dicts, which drops the six lines of storage wiring:

```python
stack = BaseCache.from_config([
    # local: fast, in-process, gone when the interpreter exits
    {"template": "memory"},
    # 'remote': cloudpickle files for values, SQL for the call records
    {"template": "sql", "root": tmp_dir.name, "values": "cloudpickle"},
])
```

The walkthrough still needs the two levels individually, so the next cell unpacks them with `local_cache, remote_cache = stack.stack`, and a note points out that this is the same pair you'd get from `CacheStack((local_cache, remote_cache))` if you built the storages yourself.

---

notebooks/ConcurrentExecution.ipynb
pmrv: Done in 59b447e — all four in-memory examples now open with `with cache("memory"):` and no setup.

Where the old code kept a `my_cache` handle only to assert against it afterwards, the assertion reads the active cache instead:

```python
print("In the cache:", cache().contains(add.fleche.digest(0, 1)))
```

The one exception is the `bind` example, which deliberately calls outside the cache context. That one reads `bound_add.cache`, which doubles as a demonstration that the bound callable carries its cache with it.

The file-backed examples got the same treatment with `BaseCache.from_config({"template": "pickle", "root": tmpdir})`, and the storage reference table at the bottom now names config templates rather than storage classes. That leaves the imports down to `fleche` plus `BaseCache`.

---

pmrv:
> seems like previously not all notebooks were linked to docs/ or into the test runner as well

Right, and that gap is exactly why the two broken ones were broken. The state before this PR:

| Notebook | docs toctree | test runner |
|---|---|---|
| GettingStarted, ExtraMethods, StorageBackends, SecureStorage, CacheStack | ✅ | ✅ |
| ConcurrentExecution | ✅ | ❌ |
| Caches, TransferWorkflow | ❌ | ❌ |

The two with no coverage at all are precisely the two that had rotted against the value/call storage split.

Fixed in 26561ae: symlinked `Caches` and `TransferWorkflow` into `docs/notebooks/` and added all three to the toctree. Rather than extending the hand-maintained list in `test_notebooks.py`, the parametrisation now discovers them:

```python
NOTEBOOKS = sorted(p.name for p in NOTEBOOKS_DIR.glob("*.ipynb"))
```

so a notebook added later is covered without anyone remembering to list it. All 8 pass, and the docs build stays at 0 warnings under `sphinx -n`.

---

pmrv:
lgtm, merge after CI.

pmrv:
`build (3.14)` failure diagnosed and fixed in 6e40105. 3.11–3.13, `sql-backends`, `ty` and `test-minimum-deps` were all green; only 3.14 failed.

**Cause.** Python 3.14 changed the default multiprocessing start method on Linux from `fork` to `forkserver`. A `forkserver` worker boots a fresh interpreter and imports `__main__` to resolve the function it was handed — which cannot work for a function defined in a notebook cell, so the worker dies and the pool breaks.

Reproduced on 3.11 by requesting the context explicitly, which isolates it to the start method:

| `mp_context` | result |
|---|---|
| `forkserver` | `BrokenProcessPool` |
| `fork` | passes |

Worth being clear on one point, since the automated summary above suggests looking at the decorator: this is not a `@fleche` pickling problem. The identical cell — same `BoundWrapper.bind`, same `from_config` cache — passes under `fork`. It is the ordinary Jupyter limitation that cell-defined functions are not importable, and `spawn` would fail the same way as `forkserver`.

Nor is it new breakage. `ConcurrentExecution.ipynb` was in the docs toctree but not in the test runner, so no CI job had ever executed it; adding it to the auto-discovered list in 26561ae is what surfaced it.

**Fix.** The two `ProcessPoolExecutor` examples now pass `mp_context=FORK`, with a note that this is a notebook constraint rather than a fleche one — code in an importable module needs no pinning. `SingleNodeExecutor` is untouched, since executorlib ships its own cloudpickle-based transport and does not have the problem.

Verified by re-running the notebook with the kernel default forced to `forkserver`, reproducing the 3.14 environment: passes.

---

Issue #760, opened 2026-07-17
Refactor: extract per-key vs multi-bag layout strategy from `BagOfHoldingH5FileBackend`


pmrv:
## Location

`src/fleche/storage/bagofholding_file.py:229-298` — `_path`, `_lock_path`,
`_contains`, `_evict`, `list` all switch on `self.prefix_length == 0`.

## Problem

Since PR #746 (closes #556), `BagOfHoldingH5FileBackend` hosts two
structurally different disk layouts under one class:

- `prefix_length == 0`: per-key files (`root/{full_key}`), inherited
  behaviour via `super()` calls.
- `prefix_length >= 1`: multi-bag layout (`root/{prefix}.h5`, sibling
  groups named by full key via bagofholding 0.1.12's `file.h5/group`
  path addressing).

The branch on `prefix_length == 0` is re-spelled in five methods:

```python
def _path(self, key: str) -> Path:
    if self.prefix_length == 0:
        return super()._path(key)
    return self._bag_file(key) / key

def _lock_path(self, key: str) -> Path:
    if self.prefix_length == 0:
        return super()._lock_path(key)
    return Path(f"{self._bag_file(key)}.lock")

def _contains(self, key: Digest) -> bool:
    if self.prefix_length == 0:
        return super()._contains(key)
    # h5py.File open + membership check
    ...

def _evict(self, key: Digest) -> None:
    if self.prefix_length == 0:
        return super()._evict(key)
    # single-file lock + delete-group + garbage-collect empty file
    ...

def list(self) -> Iterable[Digest]:
    if self.prefix_length == 0:
        return [... per-key files ...]
    # walk multi-bag files, open each, collect keys
    ...
```

The two layouts have different invariants (per-key files vs. one
`.h5` per prefix), different lock granularity (per-key `.lock` vs.
per-bag `.lock`), and different `list()` semantics (whole-file vs.
walk-and-open). Any new backend-level method (e.g. a future `size`,
`atime`, or migration hook) will require the same fork, and the two
layout invariants have to be re-learned from scratch each time.
PR #748's `refix()` and `consolidate()` already added a third
implicit layout concern (mixed-root consistency checks), and future
storage-level features would compound the branching.

## Suggested refactor

Introduce a small `_Layout` strategy resolved in `__post_init__` from
`prefix_length`:

```python
class _Layout(Protocol):
    def path(self, root: Path, key: str) -> Path: ...
    def lock_path(self, root: Path, key: str) -> Path: ...
[...]

PR #843, opened 2026-08-08
feat(storage): atomic renames instead of file locks in pickle-family backends

pmrv:
## Problem

On Unix the `filelock` package never removes its lock files, and the file backends create one per key, in the same directory as the data:

- Every stored entry left a permanent `{key}.lock` sibling → a cache with N entries occupies ~2N inodes, which counts against filesystem quotas (HPC home/work filesystems typically meter inode counts).
- The wrapper's *try-load-first* hot path acquired the read lock before checking the data file exists, so even a plain cache **miss** created a `{key}.lock` — and if the computation then failed or was aborted, the orphan lock stayed forever. Orphans were invisible to every cleanup path (`list()` filters them out; `evict()` only removes locks together with their entry).

## Fix

`FileStorage.put` now writes each entry to a dot-prefixed temp sibling and atomically renames it into place (`_atomic_write`). Readers can therefore never observe a partially written file, which was the only thing the locks protected for per-key data files — so per-key file storage drops cross-process locking entirely:

- `get` is a plain read; `_file_read_lock_with_fallback` (and its 1 s stall on contention) is gone from `FileStorage`.
- `_rewrite_all` (`compress_all`/`decompress_all`) rewrites through the same atomic-rename helper.
- Concurrent writers of the same digest write equivalent bytes, so last-write-wins is harmless in the content-addressed model.

The **bagofholding backend keeps locking for multi-bag mode only** — those files are shared between keys and mutated in place, so a rename cannot make them safe. It owns `put`/`get`/`_lock_path` and the read-with-fallback helper itself, and its litter is bounded: one `{prefix}.h5.lock` per bag (≤256 at the default `prefix_length=2`), removed with the emptied bag. Per-key mode (`prefix_length=0`) inherits the lock-free atomic-rename path — `H5Bag` writes the temp sibling in full and the rename publishes it — and creates no lock files at all.

## Compatibility & migration

- `lock_timeout` is removed from `FileStorage` and now lives on `BagOfHoldingH5FileBackend`, the only backend that still locks (**breaking** for pickle-family constructor calls that passed it). Config dicts stay compatible: `storage_from_config` drops the key from pickle-family configs with a `FutureWarning`, and pickle-family `to_config()` no longer emits it.
- `list()` keeps filtering `*.lock`, so roots written by older versions stay readable. The leftover litter itself is not swept by fleche — a one-off `find <root> -name '*.lock' -delete` (once nothing on an older fleche uses the directory) reclaims the inodes.
- Mixed-version note: an old-fleche writer doing in-place writes concurrently with a new-fleche reader on the *same key* is no longer coordinated; old readers against new writers remain safe (they only ever see complete files).

## Verification

- Unit suite green in CI order (1646 passed); full suite was green before the review revisions and CI re-runs it here. (`tests/unit/storage/` before `tests/unit/config/` shows failures, but those reproduce on `main` — pre-existing registry pollution from `test_optional_deps.py` re-importing storage modules, not this change.)
- `ty check src/` clean apart from the pre-existing unused-ignore warning in `digest.py`.
- End-to-end: a failed load now creates nothing; 3 saves + reads produce exactly 3 files and zero `.lock` files; per-key H5 roundtrips through the atomic rename leaving exactly one file per key.
- New tests cover: no extra files on save, nothing created on miss, interrupted writes leaving no partial/temp files (old entry preserved on failed rewrite), the config-level `lock_timeout` deprecation (warn + drop for pickle, still meaningful for bagofholding), the bagofholding multi-bag lock paths (read-fallback, save releases lock), and per-key mode creating no lock files.


src/fleche/storage/pickle_file.py
```diff
@@ -121,11 +139,13 @@ def to_config(self) -> dict[str, Any]:
         serializer = self.dumps.__module__.split(".")[0].lstrip("_")
         if serializer not in ("pickle", "dill", "cloudpickle"):
             raise ValueError(f"Unknown PickleFile serializer: {serializer!r}")
+        # `lock_timeout` is deliberately not emitted: the pickle family stopped
+        # locking (writes are atomic renames), the constructor merely still
+        # accepts the key so configs from fleche < 2.1 keep loading.
```
pmrv:
```suggestion
```

pmrv: It should be dropped from the constructor as well and filtered in dict_to_storage where it should also emit a warning.

src/fleche/storage/file.py
```diff
@@ -1,54 +1,48 @@
...
+    ``_to_file`` must therefore write a complete file at the path it is given,
+    which may be a temporary sibling of the entry's final path.
+
+    ``lock_timeout`` is unused here: file locking was dropped in favour of
+    atomic renames because the ``filelock`` package never removes its
+    ``{key}.lock`` files on Unix, doubling the inode footprint of a cache
+    (and thereby filesystem quota usage).  The field is retained so configs
+    and call sites passing it keep working, and for subclasses that genuinely
+    need cross-process locking (in-place mutation of shared files, e.g.
+    :class:`~fleche.storage.bagofholding_file.BagOfHoldingH5FileBackend`).
```
pmrv: can go

pmrv (commented):
revise

not sure how much code overlap there still is between filestorage and bagofholding.  debate.

pmrv:
Revisions pushed (3db9f75): `lock_timeout` is gone from `FileStorage` and now declared on `BagOfHoldingH5FileBackend` only, the docstring paragraph went with it, and `storage_from_config` drops the key from pickle-family configs with a `FutureWarning` — so pre-2.1 configs keep loading while the constructors reject it.

On the FileStorage/bagofholding overlap:

What the bag backend still genuinely shares: `root` resolution, `_path`, the `_to_file`/`_from_file` contract, and — in per-key mode — `list`/`_evict`/`_contains`. In multi-bag mode it overrides all of those *plus* `put`/`get`, so the inheritance there buys little more than `root` handling and the abstract read/write hooks. The smell isn't dead code (the base carries nothing bag-specific anymore); it's that the subclass replaces the base's whole concurrency policy instead of filling a hook.

Two ways to shrink it:

1. **Cheap:** let per-key mode (`prefix_length=0`) inherit the lock-free atomic `put`/`get` — `H5Bag.save` to the temp sibling and rename gives per-key H5 the same torn-read protection it gives pickle. Locks would then exist only where structurally required (shared multi-bag files), per-key H5 would stop littering an unbounded `{key}.lock` per entry (this PR leaves that mode locked), and the bag overrides shrink to the multi-bag branch.

2. **Structural:** this is really #760's territory — per-key vs multi-bag as extracted layout strategies, where locking is a property of the multi-bag layout rather than of the class. That dissolves the remaining overlap outright, but I'd rather not grow this PR into that refactor.

My take: do 1 here, leave 2 to #760. Say the word and I'll push 1.

---

pmrv:
Let's do the cheap version now

pmrv:
Done in aa36e95: per-key mode (`prefix_length=0`) now inherits the lock-free atomic-rename `put`/`get` — `H5Bag` writes the temp sibling in full, the rename publishes it — so locks exist only for the shared multi-bag files, and per-key H5 no longer litters a `{key}.lock` per entry. The bag's `put`/`get`/`_lock_path`/`_evict` overrides reduce to the multi-bag branch. Verified both modes end-to-end (per-key roundtrip leaves exactly one file per key; multi-bag keeps its bounded `{prefix}.h5.lock`); full unit suite green.

---

pmrv (commented):
lgtm, merge

PR #846, opened 2026-08-08
feat(storage): give count_reuses() and gc() a no-load reference walk

pmrv:
## Why

`count_reuses()` is documented as the tool for "which entries are shared?", but it answers by deserializing every entry. That means it needs every payload class importable — so on the case it is most wanted for, an archived or foreign store whose producing project is a version behind, uninstalled, or somebody else's, it fails with `ModuleNotFoundError` rather than with anything to do with the cache. Same blind spot in `Cache.gc()`'s transitive walk.

## What

`scan_child_digests(key)` — the no-load twin of `child_digests` — plus `count_reuses(load=False)` and `Cache.gc(load=False)` routing through it. The new `src/fleche/storage/scan.py` reads the reference graph straight off the serialized form:

- **`scan_pickle(bytes)`** runs a cut-down unpickler machine over `pickletools.genops`, taking stack effects from pickletools' own opcode table (the same MARK bookkeeping `pickletools.dis` does) and rebuilding *only* the container skeleton. Classes become inert `(module, qualname)` pairs, so nothing is imported and no `__reduce__` from the file is ever called. Covers protocols 0–5 plus dill and cloudpickle streams.
- **`scan_h5(group)`** walks bagofholding's self-describing `content_type`/`module`/`qualname` HDF5 attributes. Duck-typed, so `scan.py` needs no `h5py` import and stays free of the optional dependency.

Wired up per concrete value storage:

| storage | how |
|---|---|
| `ValueMemory` | reads the live object in place (nothing is serialized; also skips `get`'s defensive deepcopy) |
| `ValuePickleFile` | new `PickleFileBackend.payload(key)` — gzip detection + signature check, no `loads` (factored out of `_from_file`) |
| `ValueBagOfHoldingH5File` | new `BagOfHoldingH5FileBackend.open_bag(key)` contextmanager, both prefix layouts |

`Cache.gc(load=False)` only swaps the *value* walk: call records hold nothing but digests, strings, and JSON metadata, so they load either way.

## Contract

Strict parity with `_raw_sub_digests` — the direct `Digest` children of a top-level built-in wrapper, and nothing else. Wrappers are matched by `(module, qualname)` string, since the whole point is recognising them in bytes another interpreter wrote; a drift test pins those strings against the real classes so a rename can't separate them.

A storage that cannot scan raises `ScanUnsupported` rather than quietly falling back to a load. For `gc` that refusal is the safe answer: a reference graph that cannot be read must never be mistaken for an empty one, or its children look collectable.

Third-party `Digested` subclasses are invisible to the scanners; the load path is nearly as blind (it only matches the three built-in shapes), and `docs/dev/extending_destructurer.rst` now says so.

## Tests

`tests/unit/storage/test_scan.py` (132 cases). The load-bearing ones:

- **Parity** per key and per store across all six value backends, at four `remaining_depth` settings, and over hypothesis-generated nested values.
- **The motivating scenario**: a subprocess writes a store with an extra `sys.path` entry, so its payload class is genuinely unimportable in the test interpreter. `count_reuses()` raises `ModuleNotFoundError`; `count_reuses(load=False)` reproduces the counts the writing interpreter recorded, exactly. `gc(load=False)` sweeps the same store without collecting anything reachable.
- **Drift pins** on the wrapper name table and the `Digest` class name.
- **Corrupt input**: 2000 deterministically mutated streams must fail the way `pickle.loads` fails, never with an `IndexError` out of the stack machine.
- Compression, signing, tampered signatures, short-digest keys, missing keys, and the `ScanUnsupported` refusal paths.

Full suite: 1842 passed, 11 skipped. `ty check src/`: clean. Docs build clean.

https://claude.ai/code/session_01Fcn5MJx3cRxcWAEhsFQc8y

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fcn5MJx3cRxcWAEhsFQc8y)_


pmrv:
too complicated for now

Issue #245, opened 2026-03-30
Storage Reorganization


pmrv:
- [ ] Factor out CallStorage and ValueStorage ABCs cleanly
- [ ] Convert existing storages to Mixin based design
  - Existing storages become 'backends', 'StorageBase', or 'StorageBackend'? (except Sql that just implements CallStorage directly)
  - Create Mixins that implement CallStorage or ValueStorage in terms of a backend
- [ ] Actual call/value storages to be used can then be assembled by multiple inheritance and combining multiple features

Final application code might look something like

```python
class MemoryValues(ValueMixin, ValueStorage, MemoryBackend):
  ...

class BagOfHoldingValues(DestructuringMixin, ValueMixin, ValueStorage, BagOfHoldingFile):
  ...

class PickleCalls(CallMixin, CallStorage, PickleFile):
  ...
```

pmrv:
> There's a wart where Sql no necessarily implements StorageBackend instead of just CallStorage, since the latter is a subclass of the former. Ideally, I'd like to only implement the 'high level' interface for storages that are specific to call or values.

Fixed in #249

pmrv:
@claude  Take stock, I think we got everything we wanted from here, but double check.

pmrv:
**Claude finished @pmrv's task in 1m 55s** —— [View job](https://github.com/pmrv/fleche/actions/runs/24250792300)

---
### Taking Stock: Issue #245 Status Check

- [x] Understand the request
- [x] Read current storage files to assess state
- [x] Compare against original 5-PR plan and issue requirements
- [x] Report findings

---

### Verdict: All major goals are complete ✓

The three original issue requirements are fully implemented:

| Requirement | Status |
|---|---|
| Factor out `CallStorage` and `ValueStorage` ABCs cleanly | ✅ Done — `CallStorage`, `ValueStorage` are thin interfaces; `CallMixin`, `ValueMixin` hold the implementation |
| Convert existing storages to mixin-based design | ✅ Done — `MemoryBackend`, `VoidBackend`, `PickleFileBackend`, `BagOfHoldingH5FileBackend`, `FileStorage` are clean backends |
| Assemble call/value storages via multiple inheritance | ✅ Done — `ValueMemory(ValueMixin, DestructuringMixin, MemoryBackend)`, `CallPickleFile(CallMixin, PickleFileBackend)`, etc. |

The proposed final form from the issue is realized (with slightly different but equivalent naming):

```python
[...]

PR #849, opened 2026-08-09
docs(storage): promote the storage-hierarchy devnote to a doc page

pmrv:
Converts `devnotes/storage-hierarchy.md` into a proper page in the Sphinx tree at `docs/dev/storage_hierarchy.rst`, and deletes the devnote.

## Why it needed more than a move

The devnote was written against PR #245 and had drifted. Corrections made while porting it:

| Devnote said | Actually true |
|---|---|
| `KeyManagement` is the root | `OperationContext` is the root; `KeyManagement` inherits it, and so does `BaseCache` — that shared root is what lets the locking mixins attach to either layer |
| thread-safety mixins — not mentioned at all | `PerKeyLockMixin` / `SerializingMixin` exist and every non-`Void` concrete storage stacks one |
| add `DestructuringMixin` yourself to opt in | it is already in the bases of every value storage; destructuring is the default |
| `Sql` inherits `CallStorage, KeyManagement` | `Sql(PerKeyLockMixin, CallStorage)` |
| `CallStorage` is `save`/`load` | also `query` (abstract) and `transform` (concrete) |
| `ValueMixin.load` normalises keys inline | that moved to `KeyManagement._normalize_key` |
| `FileStorage` subclasses "implement `_to_file`/`_from_file`" | still true, but the contract changed with the atomic-rename work in #843 — `_to_file` must write a complete file at whatever path it is handed |

New material that had no devnote equivalent: the `Intent.READ` caveat, the base-list table for all nine concrete storages, a walkthrough of one `save()` down the real MRO, and what `register_storage`'s exact-class check protects against.

## Figures

The single graphviz diagram became a hairball once the extra layers were added, so it is split in two:

- `docs/figures/storage_hierarchy.dot` — the abstract skeleton, everything a concrete class is assembled from
- `docs/figures/storage_mro.dot` — `ValuePickleFile` exploded into its ten-class linearization, each step annotated with what it contributes to a `save()`

Both are hand-written Graphviz (rendered with `dot -Tsvg`) using the same solarized-light palette as `docs/_static/custom.css` and the destructuring figures, so a retheme touches all three together. The concrete classes live in a table on the page rather than in the diagram — mixin x backend is a matrix, and drawing it as a DAG was what made the old picture unreadable.

## Verification

- Every code example and structural claim on the page was executed against the importable package: MRO order, `Sql` having no `put`/`get`, the `__hash__` guard on `MemoryBackend` subclasses, `expand()` on short prefixes, the `Cache(values=, calls=)` and custom-mixin snippets.
- `sphinx -b html -n` builds clean; all 133 internal cross-references on the page resolve.

`agents/DEVELOPING.md` is updated for the moved paths.


pmrv:
## Review round 1 — fable

Overall this is a strong page: I verified the base-class table against all nine concrete classes, the claimed `ValuePickleFile` MRO against the interpreter (`ValuePickleFile → PerKeyLockMixin → DestructuringMixin → ValueMixin → ValueStorage → PickleFileBackend → FileStorage → StorageBackend → KeyManagement → OperationContext`), every edge in `storage_hierarchy.dot` against the source, both code snippets by executing them, and the `-n` Sphinx build (clean, toctree entry present). Nothing of value from the old devnote was dropped — its stale claims are all correctly fixed. However, the one section that narrates runtime behaviour rather than structure — the `save()` walkthrough — misstates where the lock is taken, and I confirmed that empirically, so I'm treating it as blocking on a page whose whole purpose is to be the accurate map.

### Blocking

**R1-1** — `docs/dev/storage_hierarchy.rst:215-228` and `docs/figures/storage_mro.dot:43-44` (node `n1`): the walkthrough says step 1 is "`PerKeyLockMixin` wraps the operation in this key's `RLock`", *then* destructuring happens. That is not what `save()` does. `PerKeyLockMixin` has no `save`; it only overrides `_operation_context`, and `DestructuringMixin.save` never enters `_operation_context` at all. The lock is taken inside `ValueMixin.save` (`src/fleche/storage/base.py:360-365`), once per stored entry, under each entry's *own* key. I traced `ValueMemory().save([1, [2, 3]])`: `DestructuringMixin.save` is entered first, then the lock for `digest([2, 3])` is acquired (and released), then the lock for the parent entry's key — no lock ever spans the whole destructuring walk. This matters because the text invites exactly the wrong thread-safety conclusion: a reader will assume a destructured save is atomic under the parent key's lock, when in fact concurrent operations can interleave between the sub-entry writes. The figure's `n1` annotation "takes this key's RLock, then delegates on" encodes the same error. Suggested fix: renumber the walkthrough — (1) `DestructuringMixin` splits the value and routes each part through `super().save()`; (2) `ValueMixin` computes the entry's key and enters `_operation_context`, *which is where* `PerKeyLockMixin` takes that entry's lock; (3) `put` resolves to `FileStorage`, etc. — and reword `n1` to something like "contributes `_operation_context`: one RLock per key, entered by every `ValueMixin`/`KeyManagement` operation". The parenthetical about reentrancy should also move: the RLock is reentrant because `expand` re-enters `_operation_context(key)` *inside* `load`'s already-held lock (`ValueMixin.load` → `_normalize_key` → `expand`), not "inside a later load".

### Should fix

**R1-2** — `docs/dev/storage_hierarchy.rst:95`: "Three mixins are orthogonal to both axes and can be layered onto anything" is false for `DestructuringMixin`, and the page itself contradicts it seven lines later ("It composes only with a value chain"). `DestructuringMixin` is a `ValueStorage` subclass — it is firmly on the *what* axis. Suggest: "Three mixins are independent of the storage medium" or introduce only the two locking mixins as layerable-onto-anything and present `DestructuringMixin` separately.

**R1-3** — `docs/figures/storage_mro.dot:51-52` (node `n5`): the label "`_to_file(): HMAC-sign, optionally gzip, pickle`" lists the pipeline in the wrong order. The code (`src/fleche/storage/pickle_file.py:63-68`) does `signer.dumps(self.dumps(value))` then optionally `gzip.compress` — i.e. pickle → HMAC-sign → gzip. The order is load-bearing: the signature covers the pickle payload, and gzip wraps the *signed* bytes (which is why `_from_file` sniffs the gzip magic before verifying). Suggest "pickle, HMAC-sign, optionally gzip".

### Nits / optional

**R1-4** — `docs/dev/storage_hierarchy.rst:28-29`: "the only thing the cache layer shares with the storage layer" is contradicted by the next sentence — the locking mixins are shared too (`Cache(PerKeyLockMixin, BaseCache)` in `src/fleche/caches.py:314`), plus `Intent` and a couple of helpers. "The shared root of both layers" says what is meant without overclaiming.

**R1-5** — `docs/dev/storage_hierarchy.rst:108`: "a striped table" — the lock table is not striped; it is one weakly-held `RLock` per distinct key (`WeakValueDictionary` in `src/fleche/storage/thread_safe.py:102-140`), so distinct keys never share a lock. The word is inherited from the `PerKeyLockMixin` docstring, but this page needn't repeat the inaccuracy (and could fix the docstring while at it).

**R1-6** — `docs/dev/storage_hierarchy.rst:170-171`: "contains nothing but a ``to_config``" — `ValueMemory`/`CallMemory` also restate `__hash__ = object.__hash__` (mandatory, per the page's own `MemoryBackend` note). A parenthetical "(plus the `__hash__` restatement on the memory classes)" would keep the two sections consistent.

**R1-7** — `docs/dev/storage_hierarchy.rst:249-252`: the `Sql.query` bullet lists "name, module, version, and argument filters" as pushed down to SQL; the code also pushes down `code_digest`, `result`, and simple-typed metadata filters (`src/fleche/storage/sql.py:417-484`). Not wrong as far as it goes, but on a contributor page completeness is cheap.

**R1-8** — `docs/figures/storage_hierarchy.dot:36`: hollow arrowheads (`onormal`) pointing base → subclass invert the UML convention, where the hollow triangle sits on the *base*. The page states its convention explicitly (line 20), so this is a judgement call — but UML-literate readers will momentarily read the diagram upside down; either flipping the edges (`dir=back`) or using a plain arrowhead would remove the false affordance.

https://claude.ai/code/session_01Jy5wiapurdJjyPPuTAwnav

pmrv:
## Author response — round 1

Good catch on R1-1; that one was a real error, not a wording problem. All eight addressed in 1349b7b.

**R1-1 (blocking) — fixed.** Confirmed independently by subclassing `ValueMemory` with a tracing `_operation_context` and saving `[1, [2, 3]]`:

```
('enter', 'ed76af7c')   # the child [2, 3]
('exit',  'ed76af7c')
('enter', '2f6b57d5')   # the parent
('exit',  '2f6b57d5')
```

Two disjoint lock scopes, neither spanning the walk — exactly as you described. `DestructuringMixin.save` and `_intern_rec` contain no `_operation_context` at all.

The walkthrough is rewritten to follow the *lookup* order rather than implying a call order: the call lands on `DestructuringMixin` (2) because nothing above it defines `save`, recurses into `ValueMixin` (3), and only there is the context entered — which is where position 1 gets its say. Added a paragraph spelling out that a destructured save takes one lock per entry, children first, and is **not** atomic as a unit, plus why that is safe anyway (content addressing means racing threads write identical bytes, and `FileStorage`'s atomic rename means a reader sees one complete version or the other). Noted the same holds for `SerializingMixin`.

The `n1` label in `storage_mro.dot` now reads "no save() of its own — supplies the RLock that 3 enters", `n3` mentions the lock, and the figure's alt text says the arrows are lookup order, not execution order.

**R1-2 — fixed.** Now: "The two locking ones are orthogonal to both axes and compose with anything; the third is value-only."

**R1-3 — fixed.** Verified in `pickle_file.py:63-68`: `signer.dumps(self.dumps(value))` then `gzip.compress`. Label is now "pickle, HMAC-sign, optionally gzip"; the prose walkthrough gives the same order.

**R1-4 — fixed.** "The root, and the point where the storage and cache layers meet."

**R1-5 — fixed.** Now "one `RLock` per distinct key, from a table held weakly so locks for keys nobody is touching are collected." Agreed the source docstring in `thread_safe.py` says "striped" too, but that is a one-word fix in code I would rather not smuggle into a docs-only PR — happy to open a separate one if you want it.

**R1-6 — fixed.** "its body is little more than a `to_config`", with a parenthetical pointing at the `__hash__` requirement documented in the `MemoryBackend` row.

**R1-7 — fixed.** Now lists code-digest, result, and argument filters, and adds that metadata joins down only when every filter value is a simple scalar, with the post-load check running on every yielded row regardless.

**R1-8 — fixed.** Kept base→subclass so the figure reads top-down with the layers stacked, but switched the arrowheads from `onormal` to a plain `vee` so they no longer look like a UML generalization arrow pointing the wrong way. The caption and the `.dot` header comment both now say so explicitly.

Figures re-rendered, `sphinx -b html -n` still clean.

pmrv:
## Review round 2 — fable

Verified against source (`src/fleche/storage/*.py`, `src/fleche/caches.py`, `src/fleche/call.py`), by re-running the lock trace and both code snippets in the interpreter, and with a clean `-n` Sphinx build at 1349b7b.

### Round-1 fixes

- **R1-1 (blocking) — confirmed fixed.** I reproduced the trace independently (`Traced(ValueMemory)` with an instrumented `_operation_context`, saving `[1, [2, 3]]`): `enter/exit ed76af7c` (child), then `enter/exit 2f6b57d5` (parent) — children first, two disjoint lock scopes, nothing spanning the walk. The rewritten walkthrough is correct end to end: I checked each step's "who defines what" claim against the MRO (positions 0–9 match the figure exactly), `super()` from `PerKeyLockMixin` does land on the no-op at position 9, `put` does resolve past `PickleFileBackend` to `FileStorage` at position 6, and the reentrancy parenthetical is now attached to the right mechanism (`expand` re-entering the same key's lock). The new `n1`/`n3` labels and the "lookup order, not execution order" alt text match the prose. On the new harmlessness claim, see R2-1 below — the conclusion is right, two edges of the phrasing are loose.
- **R1-2 — confirmed fixed** (line 96–97).
- **R1-3 — confirmed fixed**: `n5` and the prose both read pickle → HMAC-sign → gzip, matching `pickle_file.py:63-68`.
- **R1-4 — confirmed fixed** (line 29).
- **R1-5 — confirmed fixed**: "one `RLock` per distinct key, from a table held weakly" matches the `WeakValueDictionary` in `thread_safe.py:102-140`. (Agree the source docstring's "striped" belongs in a separate PR.)
- **R1-6 — confirmed fixed** (lines 178–181).
- **R1-7 — confirmed fixed**: the query bullet now matches `sql.py:417-546`, including the scalar-only metadata pushdown and the post-load check running on every yielded row.
- **R1-8 — confirmed fixed**: `arrowhead=vee` in both `.dot` files, and the checked-in `storage_hierarchy.svg` was regenerated (16 filled vee arrowheads, no hollow polygons); the convention note on lines 20–22 explains the choice.

### New findings

**R2-1** — `docs/dev/storage_hierarchy.rst:253-258`: the harmlessness argument reaches the right conclusion but two phrasings are loose.

* "an entry's key is the digest of its *bytes*" — the key is the structural digest of the *value* (`ValueMixin.save`: `key = digest(value)`, `base.py:360-362`), computed before serialization. The bytes on disk vary with serializer and `compress`, and can differ across processes for the same value (pickle is not canonical, e.g. set iteration order under hash randomization). So racing writers are writing *equal values*, not necessarily identical bytes — which is still sufficient, because the rename guarantees a reader decodes one complete file and both decode to the same value. Suggest "the digest of its content".
* The rename half of the argument is specific to the per-key file layouts. It holds for `ValuePickleFile` (the walkthrough's subject) and for `BagOfHoldingH5FileBackend` with `prefix_length=0`, but the multi-bag mode mutates a shared HDF5 file in place and substitutes cross-process `filelock` for the rename (as the backend table itself says, and `bagofholding_file.py:281-294` confirms), while `MemoryBackend` needs neither — it is process-local, and the same per-key lock that wraps `put` also wraps `get`. Since this paragraph is the page's general justification for the non-atomicity of destructured saves, one scoping sentence ("for the per-key file layouts; the multi-bag layout leans on its file lock instead, and the memory backend on the per-key lock alone") would stop a reader from carrying the rename argument to backends where it doesn't apply.

One thing I checked and did *not* find wrong: content addressing presupposes no caller-supplied key, and the cache layer indeed never passes one (`call.py:248` — `stash` calls `values.save(value)` bare), so the "identical content" premise holds in all of fleche's own call paths; and even for a hypothetical explicit-key save, the one-complete-version-or-the-other half still stands on its own.

**R2-2** — `docs/dev/storage_hierarchy.rst:343-346`: "Order is not cosmetic: the mixins work by overriding ``save``/``load`` or ``_operation_context`` and delegating with ``super()``, so a mixin placed *after* the backend never runs." The stated rule is wrong in both directions, verified empirically:

* A *locking* mixin placed after the backend still runs: `class LockLast(ValueMixin, MemoryBackend, TattlingSerializing)` — the instrumented `_operation_context` fired on `save`, because no backend defines that hook, so MRO lookup still reaches the mixin.
* `DestructuringMixin` placed *before* the backend but after `ValueMixin` never runs: `class DestructLast(ValueMixin, DestructuringMixin, MemoryBackend)` stored the raw `[1, [2, 3]]` as a plain list — `ValueMixin.save` shadows it and calls `put` directly, never `super().save()`.

The real rule is order relative to the right *neighbour*, not the backend: `save`/`load` overriders must precede the bridge mixin, and the locking mixins' position matters for nesting relative to other `_operation_context` implementors (lock-outside-session vs session-outside-lock with `Sql`), not for whether they run. Since this is the sentence a contributor composing a new class will follow, it should state that.

**R2-3** — `docs/dev/storage_hierarchy.rst:208-209`: "Every value storage stacks ``DestructuringMixin``" is contradicted by the page's own table four rows up — `ValueVoid` is `ValueMixin, VoidBackend` (confirmed: `DestructuringMixin not in ValueVoid.__mro__`). "Every persistent value storage" (or an "except `ValueVoid`, which has nothing to destructure into" aside) fixes it.

Everything else I checked fresh came up clean: the `Sql` section matches `sql.py` (session-then-lock chaining order in `_operation_context`, `_persist_call` as a single transaction, the `LIKE … LIMIT 2` `_prefix_candidates`, the three tables), the registration section matches `base.py:22-85` (exact-class check, hand-written `to_config`, the three-name `ValuePickleFile` registration via `with_*` factories), both "Composing your own" snippets execute as written (including the `Cache(values=ValueMemory({}), calls=Sql(url=...))` pairing), all 16 edges in `storage_hierarchy.dot` match the source, and the backend table's claims all check out against the code.

### Verdict

One more (small) pass. The round-1 blocking issue is genuinely fixed and independently reproduced; the page is converging. R2-2 is the only finding I'd hold the merge for — it's a false composition rule in the exact section that tells contributors how to compose. R2-1 and R2-3 are one-line wording fixes.

https://claude.ai/code/session_01Jy5wiapurdJjyPPuTAwnav

pmrv:
## Author response — round 2

All three fixed in bff893d. R2-2 was the good one — my rule was wrong in both directions, and I'd have kept repeating it.

**R2-2 — fixed.** Confirmed both halves before rewriting:

```
A) class LockAfterBackend(ValueMixin, MemoryBackend, LoudLock)
   -> lock ran: True
   MRO: ... MemoryBackend, StorageBackend, KeyManagement, LoudLock, SerializingMixin

B) class DestructAfterBridge(ValueMixin, DestructuringMixin, MemoryBackend)
   -> entries stored: 1     (2 = destructuring ran)

C) class Correct(DestructuringMixin, ValueMixin, MemoryBackend)
   -> entries stored: 2
```

So a locking mixin survives being listed last, because nothing between it and `OperationContext` defines `_operation_context`; while `DestructuringMixin` is shadowed by `ValueMixin` even though it sits well before the backend — and case B is the nasty one, since it stores the value whole and raises nothing.

Replaced the sentence with the actual rule (a mixin takes effect only if it precedes every other class in the MRO that defines the method it overrides), spelled out as two bullets covering exactly those cases, and kept the shipped order as the recommended habit rather than as a law.

**R2-1 — fixed.** Right on both counts. The key is `digest(value)` (`base.py:360`), so the claim is now that racing writers store equal *values*, with an explicit note that the bytes need not match since pickle is not canonical — they just decode to the same thing. And the atomic-rename half is now scoped: `FileStorage` renames a complete temp file, `MemoryBackend` does one dict assignment, and multi-bag `BagOfHoldingH5FileBackend` falls back to a cross-process `filelock` precisely because it mutates a file shared with other keys and cannot rename its way out. Thanks for checking the caller-supplied-key path too — that was the loophole I'd have worried about.

**R2-3 — fixed.** Now "Every value storage that actually stores anything", and the following sentence says the two `Void` classes take neither destructuring nor a lock.

Sphinx `-n` clean; the round-1 verification script still passes unchanged.

---

That is two review rounds with the blocking findings resolved and round 2 turning up three items against round 1's eight, so I'll stop here rather than run a third. Summary of what the two rounds caught, since it's the useful part of this PR's history: one wrong claim about lock scope during a destructured save, one wrong serialization order, one wrong mixin-ordering rule, and a handful of overclaims — all of them inherited from writing the page against a reading of the code rather than against the code's behaviour.

pmrv:
Added the storage sequence diagram in 55e1b49 — a `save`/`load` pair drawn from an instrumented storage rather than from reading the code, so the activation bars are the lock scopes actually taken.

It earns its place mostly because it shows the round-1 finding directly: the two bars are visibly disjoint, one per stored entry. It also surfaced something the prose hadn't said — save and load are **not** mirror images. `_intern_rec` stores children with `super().save()`, skipping the destructuring layer, while `mend` calls the *public* `load`, so children do re-enter it. And mending happens after the parent's lock is released, so loading a nested value is a series of independent reads.

Figure comes from a new `docs/figures/gen_sequence.py`, following the `gen_diagrams.py` convention. Noted in the regeneration section why the SVGs are checked in rather than rendered at build time: `.readthedocs.yaml` declares no `apt_packages`, so `dot` can't be assumed on the builder — which is also why we passed on mermaid.

The cache-side sequence diagram is #850, stacked on this branch.

pmrv:
@claude handle merge conflicts

PR #851, opened 2026-08-10
feat(metadata): add cputime/systime to Runtime metadata

### LINKED ISSUE #567: Feature Request: Resources metadata (peak memory, CPU time, …)
Tracking issue for a `Resources` metadata class, sibling to the `Runtime` / `Environment` / `Git` built-ins added in #565. Branched off so the design discussion doesn't bloat that PR.

## Sketch

A `MetaData` subclass that records resource consumption around a wrapped call:

```python
class Resources(MetaData):
    name = "resources"
    keys = {
        "peak_rss":  int,    # bytes, max-resident set size during the call
        "user_cpu":  float,  # seconds, user-mode CPU time
        "sys_cpu":   float,  # seconds, kernel-mode CPU time
        # candidates: "io_read_bytes", "io_write_bytes", "n_threads_peak", "max_vms"
    }
    def pre(self, call): ...   # snapshot baseline (rusage RU_SELF / RU_CHILDREN)
    def post(self, pre, call): # delta against baseline
```

Pairs naturally with `Runtime` (wall-clock) — together they answer "how long did this take and how much did it cost?", which is the missing half of reproducibility info now that `Environment` and `Git` answer "where + at what code state did this run?".

## Design questions to settle

1. **Backend.** `resource.getrusage(RUSAGE_SELF)` (stdlib, Unix-only) vs `psutil` (cross-platform, extra dep). `resource` covers Linux/macOS without an optional dep; Windows would need `psutil` or to gracefully no-op. Probably: stdlib by default, fall back to `psutil` if available and ask the question explicitly only if a Windows user shows up.
2. **Self vs children.** Single-threaded calls only need `RUSAGE_SELF`. Calls that spawn subprocesses (e.g. anything that shells out to `git`, `mpirun`, `srun`) won't see their cost unless we also sum `RUSAGE_CHILDREN`. Pre-snapshot of CHILDREN, delta in post.
3. **Peak vs delta.** `ru_maxrss` is "max over process lifetime", not "peak during this call" — sampling it at start and end gives an overestimate when other calls grew it earlier. Documenting the caveat is cheaper than spawning a sampler thread; if accurate per-call peaks become important, an opt-in `sample_interval=` knob using a daemon thread + `psutil.Process().memory_info().rss` would do it (extra cost, opt-in).
4. **Threadsafety / nesting.** If two `@fleche` calls overlap on the same process (executor workers, async), per-call deltas via shared `rusage` get tangled. Document that `Resources` is meaningful only for single-threaded code paths; or: gate via a per-thread baseline, accepting that thread CPU time needs `time.thread_time()` rather than `rusage`.
5. **Units.** `ru_maxrss` is kB on Linux but bytes on macOS — normalise to bytes in the recorded value, document the normalisation. `time.process_time()` is fine and platform-neutral if we go that route for CPU instead of `rusage`.
6. **What about IO?** `ru_inblock`/`ru_oublock` are notoriously unreliable on Linux (typically 0). `psutil.Process().io_counters()` is more useful but is a Linux-only field on `psutil`. Probably skip in the first pass.
7. **Config.** Selectable from `fleche.toml` via `metadata = ["Resources"]` once shipped, mirroring `Runtime`/`Environment`/`Git`.

## Out of scope

- Disk / cache footprint of stored results — that's a separate concern, belongs in cache GC tooling.
- GPU memory / utilisation — would need `pynvml`, much larger optional-dep story; can be a follow-up `GPUResources` if there's demand.

cc #565 (sibling metadata).

pmrv:
## Summary

Closes #567, per [review feedback](https://github.com/pmrv/fleche/pull/851#pullrequestreview): drops `peak_rss` (not recoverable without sampling during execution) and folds the remaining CPU-time fields into `Runtime` instead of a separate `Resources` type, following the existing `walltime` naming.

- `Runtime` gains `cputime`/`systime` (float, seconds) — sourced from the stdlib `resource` module (`RUSAGE_SELF` + `RUSAGE_CHILDREN`, so subprocess cost, e.g. shelling out to `git`/`mpirun`, is counted).
- `pre()` snapshots the CPU baseline under the same key names `post()` reports, so `post`'s dict-union overwrites the baseline with the delta (same pattern `timestart`/`timestop` already uses).
- `cputime`/`systime` are `None` on platforms without the POSIX `resource` module (Windows); `timestart`/`timestop`/`walltime` are unaffected since they don't depend on it.
- No new registration/config surface needed — `Runtime` was already `@configurable` and the default metadata chain.

## Test plan

- [x] `pytest tests/unit/metadata/test_metadata.py` — covers CPU-time collection, subprocess CPU accounting via `RUSAGE_CHILDREN`, the `resource`-unavailable fallback (walltime unaffected, cputime/systime `None`), and the `Runtime` schema/pre-post-key parametrized tests.
- [x] Full suite: `pytest tests/unit tests/regression` — 1680 passed, 7 skipped.
- [x] `ruff check` and `ty check src/` on changed files.


pmrv (commented):
peak memory not recoverable unless we'd sample during function execution.  Not our job I think so let's drop it. 

for the time we better follow the existing naming and fold this in the Runtime metadata.  'walltime', 'cputime', systime'?

pmrv:
CI red on `build (3.11/3.12/3.13/3.14, ubuntu-latest)`, but not from this PR's diff: all four fail identically on `tests/integration/test_notebooks.py::test_notebook[FiveMinuteTour.ipynb]` with `ModuleNotFoundError: No module named 'ase'`.

That notebook (added in #841, touching only `notebooks/`/`docs/`) imports `ase`, but `ase` isn't installed by `pip install -e .[tests]` (no `ase` extra in `pyproject.toml`), nor by `rendernb.yml`'s `pip install -e .[tests] jupyter nbconvert ipykernel`. Since #841's diff didn't touch `src/`, `tests/`, or `pyproject.toml`, `tests.yml`'s `changes` filter skipped the `build` job's pytest step entirely for that PR, so this gap was never exercised until a PR touching `src/`/`tests/` came along — this one. Reproduces locally on a clean `.[tests]` install regardless of branch (`import ase` → `ModuleNotFoundError`), so it predates and is unrelated to this PR's changes. Leaving it out of scope here since fixing it means a scope decision (add `ase` to `tests`, or gate the notebook cell) that belongs in its own PR.

`ty`, `sql-backends`, and the rest of `build` (everything except the one notebook parametrization) are green on `7c99f56`.


---

pmrv:
Branch got rebased onto current `main` (head is now `ef54984`). The `ase` failure from before is gone — #857's `try/except ImportError` guard on that notebook cell works. But `build`/`test-minimum-deps` are still red, now on a sibling cell in the same notebook: `FiveMinuteTour.ipynb` cell 30 does an unguarded `from executorlib import SingleNodeExecutor as Executor`, and `executorlib` isn't in the `tests` extra either (it's its own separate extra in `pyproject.toml`, never installed by `pip install -e .[tests]`). Same class of gap as the `ase` cell, just not covered by #857's fix. `ty`/`sql-backends` are green on `ef54984`; unrelated to this PR's diff (`src/fleche/metadata.py`, `tests/unit/metadata/`).


---

src/fleche/metadata.py
```diff
@@ -92,38 +99,82 @@ def configurable(cls: type["MetaData"]) -> type["MetaData"]:
...
         """
-        return {
+        post: dict[str, Any] = {
             'timestop': (t := time.time()),
             'walltime': t - pre['timestart'],
         }
+        if resource is None:
+            post['cputime'] = None
+            post['systime'] = None
+        else:
```
pmrv: when resources is not available, then cputiem/systime shouldn't appear in the metadata at all.

pmrv (commented):
@claude revise add test for my comment and fix if necessary

src/fleche/metadata.py
pmrv: Fixed in `c5658fc`: `post()` now only sets `cputime`/`systime` inside the `resource is not None` branch, so they're omitted from `call.metadata["runtime"]` entirely (not present as `None`) when `resource` is unavailable. Updated `test_runtime_metadata_cpu_time_without_resource_module` to assert `"cputime" not in runtime` / `"systime" not in runtime` instead of `is None`.


---

src/fleche/metadata.py
```diff
@@ -122,11 +122,12 @@ class Runtime(MetaData):
...
         timestop (float): The timestamp when the execution stopped.
         walltime (float): The total wall-clock execution time in seconds.
-        cputime (float | None): User-mode CPU seconds consumed during the
-            call (self + child processes).  ``None`` where the POSIX
-            ``resource`` module is unavailable (Windows).
-        systime (float | None): Kernel-mode CPU seconds consumed during the
-            call (self + child processes).  ``None`` where unavailable.
+        cputime (float): User-mode CPU seconds consumed during the call
+            (self + child processes).  Omitted entirely (not present as
+            ``None``) where the POSIX ``resource`` module is unavailable
```
pmrv: Omitted entirely (not present as ``None``)  |-> Omitted

pmrv: and below as well

pmrv (commented):
@claude revise

src/fleche/metadata.py
pmrv: Trimmed in `de6e7a5` — both docstring spots (`cputime`/`systime` on the class docstring, plus `post()`'s docstring), the test docstring, and two stale `DEVELOPING.md` references that still said `None` instead of omitted.


---

PR #858, opened 2026-08-12
refactor(state): extract _lazy_default helper shared by get_cache/get_metadata

### LINKED ISSUE #806: Refactor: extract `_lazy_default` helper shared by `state.get_cache`/`get_metadata`
## Location

`src/fleche/state.py:17-28` (`get_cache`) and `src/fleche/state.py:133-144` (`get_metadata`).

## Problem

After PR #799 landed lazy default resolution, the two getters became structurally identical:

```python
def get_cache() -> caches.BaseCache:
    try:
        return _CACHE.get()
    except LookupError:
        if "cache" not in _DEFAULTS:
            _DEFAULTS["cache"] = config.load_cache_config()
        return _DEFAULTS["cache"]

def get_metadata() -> tuple[metadata.MetaData, ...]:
    try:
        return _METADATA.get()
    except LookupError:
        if "metadata" not in _DEFAULTS:
            _DEFAULTS["metadata"] = config.load_default_metadata()
        return _DEFAULTS["metadata"]
```

Only three things differ: the `ContextVar`, the `_DEFAULTS` key string, and the loader callable. Everything else — the `try`/`except LookupError`, the `setdefault`-ish population, the return — is repeated verbatim.

## Why it hurts

- Any future change to the "lazy fallback" contract — e.g. an env-var override, a thread-local override, an invalidation hook, or richer error handling around a loader that raises — has to be applied twice, and the two sites will drift.
- The `_DEFAULTS` dict is a shared cross-getter cache today; the coupling is invisible unless you read both getters. A single helper makes the shared-cache-with-per-key-loader intent explicit.
- Tests already document (in prose) that clearing `_DEFAULTS` resets *both* getters at once. That contract belongs in the helper's docstring, not spread across two module-level functions.

## Refactor sketch

Extract a tiny helper in `state.py`:

```python
def _lazy_default(var: ContextVar[T], key: str, loader: Callable[[], T]) -> T:
    try:
        return var.get()
    except LookupError:
        if key not in _DEFAULTS:
            _DEFAULTS[key] = loader()
        return _DEFAULTS[key]
```

Each getter collapses to one line:

```python
def get_cache() -> caches.BaseCache:
    return _lazy_default(_CACHE, "cache", config.load_cache_config)

def get_metadata() -> tuple[metadata.MetaData, ...]:
    return _lazy_default(_METADATA, "metadata", config.load_default_metadata)
```

## Pros

- Removes ~20 lines of duplicate scaffolding; the "lazy fallback" contract becomes a first-class named function with a single docstring.
- Extending to a third context var (e.g., a future `project()`/`tags()` default) becomes a one-liner.
- Test coverage naturally shifts to the helper — clearing `_DEFAULTS` is exercised through one API surface, not two.

## Cons / risks

- One extra call-frame per getter access (negligible; the hot path already goes through `ContextVar.get()` in the common case).
- Slight readability trade-off: readers coming to `get_cache` have to look up `_lazy_default` to see what it does. Mitigated by keeping the helper module-local and short.
- No new external dependencies.

## Scope

Small, self-contained refactor. Zero behaviour change. Existing unit tests for lazy-default resolution and `_DEFAULTS` invalidation between tests already pin the behaviour and should pass unchanged.

pmrv:
## Summary

`get_cache` and `get_metadata` in `src/fleche/state.py` were structurally identical: try the `ContextVar`, fall back to a memoised loader result keyed in `_DEFAULTS` on `LookupError`. Only the var, the `_DEFAULTS` key, and the loader callable differed.

Extracted a `_lazy_default(var, key, loader)` helper that owns the try/except + memoise dance once; both getters now collapse to a single call.

```python
def _lazy_default(var: ContextVar[_T], key: str, loader: Callable[[], _T]) -> _T:
    try:
        return var.get()
    except LookupError:
        if key not in _DEFAULTS:
            _DEFAULTS[key] = loader()
        return _DEFAULTS[key]
```

Zero behaviour change — same `_DEFAULTS` dict, same memoisation contract (clearing an entry, or the whole dict, forces re-resolution).

## Test plan

- [x] Added `test_lazy_default_returns_set_var_without_loading` and `test_lazy_default_memoises_loader_result` in `tests/unit/config/test_config.py`, exercising `_lazy_default` directly against a throwaway `ContextVar`.
- [x] Existing `get_cache`/`get_metadata` coverage (`test_load_default_cache`, `test_load_default_metadata`, and the rest of `tests/unit/config/test_config.py`) passes unchanged.
- [x] Full suite: `pytest tests/` — 1716 passed, 11 skipped (optional-dep skips), 10 deselected (the `FiveMinuteTour` notebook integration test, which requires `ase` — unrelated to this change).

Closes #806.

---
_Generated by [Claude Code](https://claude.ai/code/session_012x69gYLiFmJqz6NuezEyKn)_

### DIFF
--- a/src/fleche/state.py
+++ b/src/fleche/state.py
@@ -1,7 +1,7 @@
 from contextlib import AbstractContextManager, contextmanager
 from contextvars import ContextVar, Token
 from dataclasses import dataclass
-from typing import overload, Any, Callable, Iterator
+from typing import overload, Any, Callable, Iterator, TypeVar
 
 from . import caches, config, metadata
 
@@ -13,6 +13,22 @@ _CACHE: ContextVar[caches.BaseCache] = ContextVar("fleche.CACHE")
 #: config by clearing this dict.
 _DEFAULTS: dict[str, Any] = {}
 
+_T = TypeVar("_T")
+
+
+def _lazy_default(var: ContextVar[_T], key: str, loader: Callable[[], _T]) -> _T:
+    """Return ``var``'s active value, falling back to a memoised ``loader()`` result.
+
+    The fallback is resolved on first use and cached in :data:`_DEFAULTS` under
+    ``key``; clearing that entry (or the whole dict) forces re-resolution.
+    """
+    try:
+        return var.get()
+    except LookupError:
+        if key not in _DEFAULTS:
+            _DEFAULTS[key] = loader()
+        return _DEFAULTS[key]
+
 
 def get_cache() -> caches.BaseCache:
     """Return the active cache, falling back to the configured default.
@@ -20,12 +36,7 @@ def get_cache() -> caches.BaseCache:
     The default is resolved from the configuration files on first use and memoised
     in :data:`_DEFAULTS`.
     """
-    try:
-        return _CACHE.get()
-    except LookupError:
-        if "cache" not in _DEFAULTS:
-            _DEFAULTS["cache"] = config.load_cache_config()
-        return _DEFAULTS["cache"]
+    return _lazy_default(_CACHE, "cache", config.load_cache_config)
 
 
 class _StickyContext:
@@ -136,12 +147,7 @@ def get_metadata() -> tuple[metadata.MetaData, ...]:
     The default is resolved from the configuration files on first use and memoised
     in :data:`_DEFAULTS`.
     """
-    try:
-        return _METADATA.get()
-    except LookupError:
-        if "metadata" not in _DEFAULTS:
-            _DEFAULTS["metadata"] = config.load_default_metadata()
-        return _DEFAULTS["metadata"]
+    return _lazy_default(_METADATA, "metadata", config.load_default_metadata)
 
 
 def meta(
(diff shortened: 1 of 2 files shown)


pmrv:
CI red on `build (3.11/3.12/3.13/3.14, ubuntu-latest)` and `test-minimum-deps` is unrelated to this PR's diff.

The branch was rebased onto `main` after #857 merged (base `main` now includes #857's ase-import fix for `notebooks/FiveMinuteTour.ipynb`). That fix lets notebook execution proceed past the `ase` cell into a later cell that imports `executorlib` unconditionally — an optional extra not installed by `pip install -e ".[tests]"` — so `tests/integration/test_notebooks.py::test_notebook[FiveMinuteTour.ipynb]` now fails with `ModuleNotFoundError: No module named 'executorlib'` in every job that runs the full suite. This is a pre-existing gap in the notebook (previously masked because the `ase` cell aborted execution first), reproducing on `main` and any PR based on it — not something introduced by the `state.py` refactor here. Filed as #861 and out of scope for this PR.

The checks relevant to this diff are green: `ty` (type check) and `sql-backends` both pass. I'll re-run once #861 is fixed and `main` recovers.

---

PR #870, opened 2026-08-14
feat(mpi): add collective() for caching MPI-rank-parallel calls

pmrv:
Tests fleche under executorlib's MPI executors, where the same decorated function runs once per rank, and lands a recipe for it.

## What I found

The starting expectation was that "rank 0 returns the result, all other ranks return `None`" should just work. It doesn't, in two distinct ways.

**1. The return shape flips between the cold and warm run.**

A fleche lookup key is built from the function profile and the arguments — it carries no notion of rank. So all N ranks compute the *same* key and are all candidates to hit the *same* record. With `@fleche` straight on the kernel and a shared file-backed cache:

```
exe.submit(bound_mpi_sum, 10).result()   # cold -> [45, None]
exe.submit(bound_mpi_sum, 10).result()   # warm -> [45, 45]
```

Nothing is corrupted, but the `None` that marked "I am not the rank holding the answer" is gone — rank 1 never computed anything, it loaded rank 0's record. Downstream code that tests for `None` to pick an owning rank (e.g. which rank writes the output file) now has every rank believe it owns the result.

**2. Ranks that disagree on hit-vs-miss deadlock.**

This is the dangerous one. A cache hit *skips the function body*, and the body is where the collectives live. Whether a rank hits is decided by that rank's own cache view, and the ranks are not synchronised — node-local scratch, an NFS client that hasn't seen rank 0's write yet, or a cache populated by an earlier run at a different rank count all produce divergence.

With a small message the mismatch can pass unnoticed (the send fits the eager buffer). I reproduced that: a `comm.reduce` of one int gave a plausible-looking answer with rank 0 never entering the reduce. Push it past the rendezvous threshold — a 16 MB `Allreduce` — and the job hangs indefinitely.

## The fix

`fleche.mpi.collective` has rank `root` do the lookup and broadcast a single bool, so every rank branches identically:

- **hit** — no rank enters the body, so no collective is entered by anyone. Rank `root` returns the cached value, the others `None` (or all of them, with `broadcast=True` for kernels that end in an `Allreduce`).
- **miss** — every rank runs the body, so the collectives match up as they always did. Only rank `root` goes through the fleche wrapper and records; the others call `func.__wrapped__` and touch no cache at all, so N ranks don't hammer shared storage with prepare/abandon cycles for results that are all `None`.

```python
@collective
@fleche.fleche
def mpi_sum(n): ...
```

Size-1 communicators are transparent, so the same code runs serially.

## Changes

- **`src/fleche/mpi.py`** — `collective(func, *, comm=None, root=0, broadcast=False)`. Not exported from `fleche.__init__`; reached via `from fleche.mpi import collective`, the same way `SshCache` lives in `fleche.remote`. Gated on a new `mpi` extra.
- **`tests/integration/test_mpi_execution.py`** — 13 tests covering both failure modes and the fix, including the divergent-view case that deadlocks without it. Skipped without executorlib + mpi4py + `mpiexec`.
- **`docs/mpi_execution.rst`** — the full recipe, plus the two alternatives (caching the submission instead of the payload; `wrap_executor`) and their trade-offs.
- Cross-link + warning from `docs/parallel_execution.rst`, `pyproject.toml` extra, `agents/DEVELOPING.md` module map / test layout / docs list.

## One implementation note worth flagging

`mpi4py.MPI` is imported **lazily inside the wrapper**, never at module scope. Importing it calls `MPI_Init`, and `mpiexec` children spawned from an already-initialised serial parent hang — this bit me during development (the first test run deadlocked for exactly this reason, not because of the bug under test). The `ImportAlarm` therefore guards bare `import mpi4py`, which does not initialise. There's a comment on it so nobody tidies it back to the top of the file.

## Known limitations (documented)

`Digest` arguments are rejected with a `TypeError` — the wrapper that would expand them is bypassed on non-root ranks, so they'd receive the digest string instead of the value. `isolate=True` doesn't combine with it (root-only `chdir` splits the ranks across working directories). Metadata is recorded for the root rank only. The rank count is not part of the key.

## Verification

Run in a container with OpenMPI 4.1.6, mpi4py 4.1.2, executorlib 1.10.2, at both 2 and 3 ranks.

- `pytest tests/` — 1737 passed, 7 skipped (skips are the pre-existing ssh tests)
- `pytest tests/integration/test_mpi_execution.py` — 13 passed
- `ty check src/` — all checks passed
- `sphinx -b html docs/` — build succeeded, no warnings

## Scope note

Shipping this as a library helper rather than a docs-only snippet was a judgement call: the failure is a silent hang and the fix (broadcast the decision) is subtle enough that a copied snippet would drift. It's opt-in and behind an extra, so it costs nothing to anyone not doing MPI. Happy to demote it to a documented pattern if you'd rather keep the surface smaller.

https://claude.ai/code/session_013QfTuypyFNoC2bcscQ9jSA

---
_Generated by [Claude Code](https://claude.ai/code/session_013QfTuypyFNoC2bcscQ9jSA)_


pmrv:
Closing this — the helper is extra code that doesn't need to live in fleche core, and an addon package (`fleche-mpi` or similar) is a better home for it. Writing up what the testing established so this is picked up from findings rather than from scratch.

## The findings stand independently of the fix

These are properties of fleche's caching model meeting MPI, not of the helper. They'd be equally true for an addon.

Tested against OpenMPI 4.1.6 / mpi4py 4.1.2 / executorlib 1.10.2, at 2 and 3 ranks, with `SingleNodeExecutor(resource_dict={"cores": N})`.

**A cache key carries no notion of rank.** It's built from the `FunctionProfile` and the arguments, so all N ranks compute the *same* key and all contend for the *same* record. Everything below follows from that one fact.

**1. The return shape flips between cold and warm runs.** With the conventional "rank 0 returns the answer, the rest return `None`":

```
exe.submit(bound_mpi_sum, 10).result()   # cold -> [45, None]
exe.submit(bound_mpi_sum, 10).result()   # warm -> [45, 45]
```

The answer is right in slot 0 both times, but the `None` that marked *"I am not the rank holding the result"* is gone — rank 1 never computed anything, it loaded rank 0's record. Anything downstream that keys off `None` to pick an owning rank (which rank writes the output file, say) misfires on the second run only. `None` returns are not cached (`wrapper.py` warns and skips), so only rank 0's record ever exists and every rank finds it.

**2. Divergent hit/miss decisions deadlock.** This is the one that matters. A cache hit *skips the function body*, and the body of an MPI kernel is where the collectives live. Whether a given rank hits is decided against that rank's own view of the cache, unsynchronised with the others — node-local scratch, an NFS client that hasn't seen rank 0's write yet, or a cache populated by an earlier run at a different rank count.

Reproduced both outcomes, and the difference between them is just message size:

- `comm.reduce` of one int, rank 0 hitting and rank 1 missing: **no hang, plausible-looking answer.** The unmatched send fits the eager buffer and the job limps on. This is the bad case — it fails silently.
- `comm.Allreduce` of a 16 MB array, same divergence: **hangs indefinitely.** Past the rendezvous threshold the sender blocks on a handshake that never comes.

So the small-message behaviour gives false confidence in exactly the toy examples someone would test with.

**3. `wrap_executor` changes the return type across the hit boundary.** Worth knowing regardless of any of the above:

```
exe.submit(mpi_sum, 10).result()   # cold -> [45, None]   (launcher's gathered list)
exe.submit(mpi_sum, 10).result()   # warm -> 45           (cache, unwrapped)
```

It short-circuits in the parent before submitting, which is a genuine win — a hit spawns no ranks at all — but callers indexing into the result have to normalise the two shapes. It also does nothing for (2): on a miss the ranks each still run the decorated function.

## What already works today, no new code

Decorating the **submission site** rather than the payload:

```python
with fleche.cache(my_cache), make_executor() as exe:
    @fleche.fleche
    def run(n):
        return exe.submit(mpi_kernel, n).result()   # mpi_kernel undecorated
```

fleche runs exactly once, in the serial parent, so rank divergence is impossible by construction. The cached value is the launcher's gathered list verbatim, so the shape is stable across runs, and on a hit no MPI processes are spawned at all — the whole allocation is saved, not just the compute inside it. Trade-offs: the `None` padding is stored alongside the answer, and the key is the *submission's* arguments, so a change in rank count doesn't invalidate it.

This covers the case where you own the call site, which is probably most of them. The helper only earns its keep when the decoration is on a kernel that has to run under MPI — called both serially and in parallel, or decorated in library code you don't control.

## The shape an addon would take

For whoever picks this up — the approach that tested clean, deadlock-free with deliberately divergent per-rank cache roots:

Have one rank own the cache. It calls `func.fleche.load(*args)` and broadcasts a single bool, so every rank branches on the *same* value regardless of its own cache view. On a hit no rank enters the body (so no collective is entered by anyone); on a miss every rank runs it, but only the root goes through the fleche wrapper — the others call `func.__wrapped__`, so N ranks don't hammer shared storage with prepare/abandon cycles for results that are all `None`. Broadcast the bool, not the value; the value never needs to leave the root unless the kernel's convention is that every rank returns it.

Two things that cost time and would cost it again:

- **`mpi4py.MPI` must be imported lazily, inside the call, never at module scope.** Importing it calls `MPI_Init`, and `mpiexec` children spawned from an already-initialised serial parent hang. This bit during development — the first test run deadlocked for this reason and not the bug under test. Bare `import mpi4py` does *not* initialise, so it's fine for an availability check.
- **An inline `# ty: ignore[unresolved-import]` for mpi4py is a trap.** It satisfies an environment without mpi4py, but becomes an *unused* suppression — and so a fresh diagnostic — for anyone who has it installed. It relocates the failure rather than fixing it. A scoped `[[tool.ty.overrides]]` entry is correct in both. Moot in a standalone package that depends on mpi4py directly, but noting it since it isn't obvious.

Known limitations of that approach, all documented in the branch: `Digest` arguments can't be passed (the wrapper that would expand them is bypassed on non-root ranks, so they'd receive the digest string); `isolate=True` doesn't combine with it (root-only `chdir` splits the ranks across working directories); metadata describes the root rank only; rank count isn't part of the key.

## Where the work is

Branch `claude/fleche-mpi-executor-caching-zxjazw` is left in place — closing this PR doesn't delete it. It has the helper, 13 integration tests (skipped without executorlib + mpi4py + `mpiexec`), and a `docs/mpi_execution.rst` page. All CI green.

The tests are the part most worth lifting into an addon verbatim. Each scenario runs as a subprocess with a hard timeout — necessary both because executorlib spawns real ranks via `mpiexec` and so that a regression reintroducing the deadlock *fails* rather than hanging the suite. They also parse a `_body_runs` count out of stdout, so they assert work was actually skipped rather than only that the output looked right.

Two things arguably belong in fleche core independently of any addon, if you want them: the `wrap_executor` return-type asymmetry in (3) is a documentation gap on `docs/parallel_execution.rst` today, and a warning there that MPI-parallel tasks run the callable once per rank would catch people before they hit (2).

---

PR #873, opened 2026-08-17
docs(usage): fix inaccurate/misleading claims in tldr, helpers, query

pmrv:
## Summary

Scheduled documentation audit of `docs/usage/{tldr,helpers,lazy_call,query}.rst`: every code example was executed against the installed package and every behavioral claim was cross-checked against `src/fleche/`. `lazy_call.rst` had no discrepancies. The other three pages had small but real inaccuracies:

- **`tldr.rst`** — the "essentials" code block runs `cache("void")` partway through, which then breaks the `expensive.load(1, 2)` call two lines later with a real `KeyError` if copy-pasted and run top-to-bottom (confirmed by running it). The block is meant to illustrate each line independently; now says so.
- **`helpers.rst`** — the metadata-shadowing warning claimed the shadowed `metadata=` kwarg is "interpreted as the filter dict." In reality it raises `AttributeError: 'str' object has no attribute 'items'`, since `.query()`'s `metadata=` machinery expects a dict (confirmed by running the exact example against a function with a `metadata` parameter). Also clarified that the `WARNING` log fires on every call where the function has a `metadata` parameter (not only the broken keyword form), and fixed `.bind()`'s ambiguous "at the moment of the call" wording — state is frozen at `.bind()`-time, not at the later call.
- **`query.rst`** — added the undocumented third query form, `cache().query(**kwargs)` (confirmed working, `caches.py` `BaseCache.query`). Corrected `.latest()`/`.oldest()`: the doc implied `Runtime` metadata is required and didn't explain what happens without it; in fact no error is raised — the result is silently an arbitrary match rather than a true extremum (confirmed by running a query with `Runtime` metadata disabled). `IndexError` is only raised for an empty result set.

## Test plan

- [x] Executed every code example in the four audited pages against the locally `pip install -e`'d package (fresh temp cache dirs per example) to confirm current behavior before editing.
- [x] Re-read the edited passages against `src/fleche/{wrapper,call,query,caches}.py` to confirm the new wording matches actual behavior.
- No code changes — docs only, so no test suite run needed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BuE52pq6tsFaQ6PbjWwS56)_


docs/usage/query.rst
```diff
@@ -126,14 +130,18 @@ Filtering and ordering
...
 ``.latest() -> LazyCall``
-    Call with the most recent ``timestop``.  Raises :exc:`IndexError` if
-    empty.
+    Call with the most recent ``timestop``.  Raises :exc:`IndexError` only
+    if there are no matching calls at all.
 
 ``.oldest() -> LazyCall``
-    Call with the oldest ``timestop``.  Raises :exc:`IndexError` if empty.
+    Call with the oldest ``timestop``.  Raises :exc:`IndexError` only if
+    there are no matching calls at all.
```
pmrv: @claude [sonnet] sounds like a bug, investigate in a separate PR.

pmrv: Confirmed as a real bug and fixed in #894: `.latest()`/`.oldest()` now raise `ValueError` when none of the matching calls carry `Runtime` metadata, instead of silently returning an arbitrary match. Added regression tests covering both the all-missing and mixed-metadata cases.

---

pmrv: Heads up for whoever reads this thread: no need to open a PR for `claude/fix-latest-oldest-no-runtime-metadata` — #894 already fixes the same bug (same root cause, same `ValueError` approach) and its CI is fully green (Tests/ty/test-minimum-deps/Performance Triage all passed). Recommend closing/discarding that branch as a duplicate once #894 merges.

Also tracking the follow-up raised above: once #894 merges, this PR's `docs/usage/query.rst` wording ("no error is raised") will need updating to describe the new `ValueError`. I'll push that once #894 is in.

---

